PR #17, opened 2016-07-18
Add ChatMsg decoder and stubbed handler

tfarley:
Adds a decoder for the ChatMsg packet. Also stubs the ChatMsg handler to replay the message back to the sender, so that the message shows up in the UI.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/PSPacket.scala
+++ b/common/src/main/scala/net/psforever/packet/PSPacket.scala
@@ -159,4 +159,8 @@ object PacketHelpers {
   def encodedWideString : Codec[String] = variableSizeBytes(encodedStringSize.xmap(
     insize => insize*2,
     outSize => outSize/2), utf16)
+
+  def encodedWideStringAligned(adjustment : Int) : Codec[String] = variableSizeBytes(encodedStringSizeWithPad(adjustment).xmap(
+    insize => insize*2,
+    outSize => outSize/2), utf16)
 }
--- a/common/src/main/scala/net/psforever/packet/game/ChatMsg.scala
+++ b/common/src/main/scala/net/psforever/packet/game/ChatMsg.scala
@@ -2,85 +2,9 @@
 package net.psforever.packet.game
 
 import net.psforever.packet.{GamePacketOpcode, Marshallable, PacketHelpers, PlanetSideGamePacket}
+import net.psforever.types.ChatMessageType
 import scodec.Codec
 import scodec.codecs._
-
-// ChatMessageTypes were reversed by:
-  // Checking the type in the packet upon receipt by server while using slash commands
-  // Replaying the ChatMsg packet back to the sender after modifying the messagetype value
-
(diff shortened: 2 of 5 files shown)


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -146,6 +146,13 @@ class WorldSessionActor extends Actor with MDCContextAware {
       sendResponse(PacketCoding.CreateGamePacket(0, KeepAliveMessage(0)))
 
     case PlayerStateMessageUpstream(_) =>
+    
+    case msg @ ChatMsg(messagetype, unk1, recipient, contents) =>
+      log.info("Handling " + msg)
```
pschord: Small nit - can you change Handling to "Chat: " + msg ?

common/src/main/scala/net/psforever/packet/game/ChatMsg.scala
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016 PSForever.net to present
+package net.psforever.packet.game
+
+import net.psforever.packet.{GamePacketOpcode, Marshallable, PacketHelpers, PlanetSideGamePacket}
+import scodec.Codec
+import scodec.codecs._
+
+// ChatMessageTypes were reversed by:
```
pschord: Can you change these // comments to
```
/** Some brief comment
  *
  * Some longer explanation.
  */
```
That way they get placed in generated documentation

common/src/main/scala/net/psforever/packet/game/ChatMsg.scala
```diff
@@ -0,0 +1,102 @@
...
+  // Checking the type in the packet upon receipt by server while using slash commands
+  // Replaying the ChatMsg packet back to the sender after modifying the messagetype value
+
+// Some other unchecked commands, if they actually exist as message types:
+// Reply (/r), BattleGroup (/bg), ReplyCSR (/gmreply)
+
+object ChatMessageType extends Enumeration {
```
pschord: Would it possible to this ChatMessageType to separate file to keep packet files somewhat clean?
Could you create a new package `net.psforever.types` and place it as `ChatMessageType`? I want to do this eventually as having all game types separate from packets will help readability down the road

common/src/main/scala/net/psforever/packet/PSPacket.scala
```diff
@@ -159,4 +159,8 @@ object PacketHelpers {
   def encodedWideString : Codec[String] = variableSizeBytes(encodedStringSize.xmap(
     insize => insize*2,
     outSize => outSize/2), utf16)
+
+  def encodedWideStringAligned(adjustment : Int) : Codec[String] = variableSizeBytes(encodedStringSizeWithPad(adjustment).xmap(
```
pschord: Figured this would have to be added eventually :wink:

pschord:
Looks good. If you could write a decode and encode test that tries a few packets, that'd be nice to make sure things dont break in the future. That could be another PR

PR #19, opened 2016-07-23
Add ChangeFireModeMessage packet and test

tfarley:


common/src/main/scala/net/psforever/packet/game/ChangeFireModeMessage.scala
```diff
@@ -0,0 +1,21 @@
...
+                                       fire_mode : Int)
+  extends PlanetSideGamePacket {
+  type Packet = ChangeFireModeMessage
+  def opcode = GamePacketOpcode.ChangeFireModeMessage
+  def encode = ChangeFireModeMessage.encode(this)
+}
+
+object ChangeFireModeMessage extends Marshallable[ChangeFireModeMessage] {
+  implicit val codec : Codec[ChangeFireModeMessage] = (
+      ("item_guid" | uint16L) ::
```
pschord: Can you change the GUID type to this?
https://github.com/psforever/PSF-LoginServer/blob/master/common/src/main/scala/net/psforever/packet/game/CharacterInfoMessage.scala#L15-L19
I know this type isn't visible. We need to move this type to the `net.psforever.types` package in another PR

PR #20, opened 2016-07-24
Add ChangeFireStateMessage packets and tests

tfarley:


pschord:
Can you figure and go ahead and rename which is for stopping and starting?

PR #32, opened 2016-07-25
Add Vector3 class

tfarley:


common/src/main/scala/net/psforever/types/Vector3.scala
```diff
@@ -0,0 +1,26 @@
...
+package net.psforever.types
+
+import net.psforever.newcodecs._
+import scodec.Codec
+import scodec.codecs._
+
+final case class Vector3(x : Float,
+                         y : Float,
+                         z : Float)
+{
```
pschord: Doesnt need a body

common/src/main/scala/net/psforever/newcodecs/package.scala
```diff
@@ -1,10 +1,15 @@
...
 
+import scodec.Attempt
+import scodec.Attempt.{Failure, Successful}
 import scodec.Codec
 
 package object newcodecs {
 
   def q_double(min: Double, max: Double, bits: Int): Codec[Double] = new QuantizedDoubleCodec(min, max, bits)
 
+  def attemptToFloat(v: Double): Attempt[Float] = Attempt.successful(v.toFloat)
```
pschord: can you consolidate this function?

PR #33, opened 2016-07-25
Fixes encoding 0.0 as q_double

tfarley:


pschord:
Was there tests added originally to handle this case?

pschord:
I'm just curious what the bug was. If there was I'll accept

tfarley:
The special handling of 0.0 was actually a piece of the IntCodec code that I mistakenly transferred over from scodec when using it as a reference. All of the q_double encoding should indeed go through the quantization function and not have a special 0.0 case.

PR #40, opened 2016-07-27
Add WeaponFireMessage packet

tfarley:


pschord:
This seems like it's on the right track, but I dont think this is correct.

The `Option[Option[T]]` thing just doesnt make sense to have.

tfarley:
It's sort of a double whammy. First, the unk6 field has to be 3 (probably some sort of type/enum) to even consider adding what appears to be a velocity, and then a boolean is written to denote if there is a velocity or not following it (i.e. that bool is zero and the velocity is not written if the magnitude of the velocity is small). That's what causes the double-Optioned Vector3, but I'm not sure if there is a more idiomatic way that flattens it down to just one Option.

pschord:
I'll accept this but we need to change that to a better API. The details about == 3 and that boolean should be hidden away from the packet user (ideally)

PR #41, opened 2016-07-27
added AvatarJumpMessage packet

Fate-JH:
I'm not certain I could have picked an easier packet as a first attempt.
The value is eight bits.


pschord:
So the value is 8 bits but you are only extracting 1 bit?

pschord:
Can you write a test for this as well (I know it's trivial, but we want to have some good tests)

Fate-JH:
I'll reset it to grab as an Int (uint8L).

I see the files where the tests are to be written.  How do I run them?

Fate-JH:
As a uint8L, it is consistently reporting as 128 (hex 80).

common/src/main/scala/net/psforever/packet/game/AvatarJumpMessage.scala
```diff
@@ -0,0 +1,20 @@
...
+  type Packet = AvatarJumpMessage
+  def opcode = GamePacketOpcode.AvatarJumpMessage
+  def encode = AvatarJumpMessage.encode(this)
+}
+
+object AvatarJumpMessage extends Marshallable[AvatarJumpMessage] {
+  implicit val codec : Codec[AvatarJumpMessage] = (
+    ("state" | bool)
+    ).as[AvatarJumpMessage]
+}
```
tfarley: Can you put a newline at end of file here? Diff can get funky otherwise.

pschord:
Oh nvm, I was confused by your first comment. The packet decoder has it as 1 bit. You can change it back to bool.

As for running tests in IntelliJ right click on the `PSF-LoginServer\common\src\test\scala` folder
You should see "Run Specs2 in Scala".

You'll want to add a test to the GamePacketTest.scala file. Here is an example of a test: https://github.com/psforever/PSF-LoginServer/pull/40/commits/3411edb94c7fd6225bc5d0670ee149cbcae69286

tfarley:
Can you also remove the pslogin-debug.log file? We really need to add it to gitignore.

Fate-JH:
@pschord Actually, changing it to uint8L is probably correct.  When I was checking, I had tried to put the value into a PlanetSideGUID and it said it couldn't extract 16 bits from 8.

@tfarley Newline at the end, and getting rid of the debug file.

I should probably close this and submit a new pull request.

pschord:
No need to close it. You can keep adding commits.
Do 
`git rm pslogin-debug.log`
And then add a test.

I'm 100% you had it right before. It's a bool as I can see it in IDA.

Fate-JH:
Sounds good.

The menu option on the test/scala folder in the project tree isn't there for me.  Did I neglect to configure something?

PR #42, opened 2016-07-27
Adding tests for GamePacket AvatarJumpMessage

Fate-JH:
This test passing was actually unexpected.

Hopefully that newline is actually there in AvatarJumpMessage and pslogin-debug.log was actually deleted.


pschord:
This was already merged by #41 
No need for second PR

PR #43, opened 2016-07-27
Various fixes to ChatMsg

tfarley:
* Adds a codec that helps to properly choose between wide and single-byte coding for the contents string
* Updates tests and handler stub to match
* Removes some unnecessary whitespace


pschord:
Complicated, but you got it to work. I'm not a fan of the `has_wide_contents` field as it then requires the user of the ChatMsg class to know before hand what the encoding is. I wonder if there is a good way to do this with scodec or if writing a codec for this packet would be better.

I think this is a good effort for the time being.

PR #44, opened 2016-07-28
Packet ContinentalLockUpdateMessage

Fate-JH:
This packet is sent by the server, upon login (usually), to instruct clients to generate event messages about which faction owns which continent.


tfarley:
The empire field is only two bits in this packet, so you can directly use PlanetSideEmpire as the type instead of a uint8L. I'd suggest adding "Neutral" to PlanetSideEmpire enum as well to cover your 'open' case, as it is probably used elsewhere too.

PR #52, opened 2016-08-10
PlayerAvatar Doing Stuff on the Server

Fate-JH:
Ugh.  This commit history.  The others notwithstanding, I don't even know why 3e0f6b3 is still on the list since it's not contributing anything and it should have been removed the last time I re-forked the upstream.

Anyway, the bulk of the changes are the files in src/main/scala/net.psforever/objects that are the proof of concept for the player system.  The folder is something of a dumping ground that will hopefully get better file organization (and maybe even a better name).  As noted, in WorldSessionActor, things are forced into proper existence; however, good setup is key.  The server should be able to keep track of player positions, one of the "look" directions, and whether they have a weapon drawn or not (and which slot).  This is all obviously future work for synching up client data and letting each other see one another.

Hopefully, with further work, and a few more files, an actual folder structure can be realized.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -376,7 +376,7 @@ object GamePacketOpcode extends Enumeration {
       case UnknownMessage46 => noDecoder(opcode)
       case CharacterCreateRequestMessage => game.CharacterCreateRequestMessage.decode
       case CharacterRequestMessage => game.CharacterRequestMessage.decode
-      case LoadMapMessage => noDecoder(opcode)
+      case LoadMapMessage => game.LoadMapMessage.decode
 
       // OPCODE 50
       case SetCurrentAvatarMessage => game.SetCurrentAvatarMessage.decode
--- a/common/src/main/scala/net/psforever/packet/game/LoadMapMessage.scala
+++ b/common/src/main/scala/net/psforever/packet/game/LoadMapMessage.scala
@@ -1,0 +1,38 @@
+// Copyright (c) 2016 PSForever.net to present
+package net.psforever.packet.game
+
+import net.psforever.packet.{GamePacketOpcode, Marshallable, PacketHelpers, PlanetSideGamePacket}
+import scodec.Codec
+import scodec.codecs._
+
+
+/**
+ * map_name and nav_map_name should match (unless you want to be lost :))
+ * 
+ * ex:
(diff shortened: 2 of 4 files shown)


Fate-JH:
I actually did more work than what I have shown here but I stripped some of it out because I realized that we should introduce a test code arena for it before plowing ahead.  I had been testing the functionality by moving log messages around behind `if` statements and watching the command prompt, but obviously moved it back to where it was before the testing.

How do I set up a new test file in the project's usual folder, with the others?

Fate-JH:
So, after looking into it, all I have to do is drop a file into the appropriate folder and set it up in an appropriate way for it to be considered a "test" file.  That's easy enough.

Fate-JH:
I have to ask this pull request be __put on hold__.  I'm doing tests on the PlayerMasterList and I'm getting terribly inconsistent results.  What test fails one time passes the next, only to fail again when tried a third time.  I have one error with Options in which a .isDefined returns true, but then the subsequent .get throws a NoSuchElement error.  I don't know if this has to do with how the test environment works or whether the status of the core component as a Scala object with hashMaps is the problem.

Fate-JH:
Okay, I believe I've confirmed that the problem is testing code that utilizes the singleton pattern.  At the least, the problem is naive tests of code using the singleton pattern.

I believe that multiple tests are running concurrently and that causes pollution.  At one moment, a certain line of one test is running that checks whether a certain element of a map isDefined. Coincidentally, another test is running that gets the element of the same mapping.  Coincidentally, another test removes that element from the mapping.  Each test is expecting completely different validations in similar situations and even varying the input isn't going to completely defeat the pollution.   That causes a somewhat random combination of test failures and genuine thrown exceptions and, very occasionally, everything working properly.

I'll have to strip out an OOP version of the PlayerMasterList.  This exposes the data structures I have been trying to keep singular and obscured, however, so it won't be a solution that I am happy with.

Fate-JH:
:chart_with_upwards_trend:

Fate-JH:
I'm going to close this pull request later today even without any movement on it.  I need to do some significant sorting with my local repository as it keeps saying it is in disarray.

pschord:
I like the ideas of this pull request, but I wont merge it just like this. But as a request for comments you can leave it open!

Fate-JH:
I admit it's very messy, and much messier than I was prepared for.
I was planning to transfer the changes here to a branch of my fork and restoring the master to be just like upstream.  That's something I've been doing wrong from the start and it stems from my experience using Subversion source control. How will that affect the PR?

PR #54, opened 2016-08-12
Add LoadMapMessage

aphedox:


pschord:
I believe that last field is a checksum. Can you rename it and rebase against master?

aphedox:
Branch refuses to rebase properly. I'm going to just make a new PR rather than pollute the history with merge commits.

PR #55, opened 2016-08-14
Add SetEmpireMessage

aphedox:
A cascade of this message plus HackMessage are sent for every building_guid upon loading a map.


tfarley:
I suggest changing "building_guid" to "object_guid", since objects such as terminals also respond to this message. Otherwise looks good to me!

PR #65, opened 2016-08-21
Added: ZonePopulationUpdateMessage

Fate-JH:
This was my actually part of an attempt to determine how to display those little faction icons in the zone preview windows on the Interstellar View.  First, though, I had to figure how to get rid of the lock icon that was in the said windows, resulting in tracing down and sorting this packet.

I originally started working on ZoneLockInfoMessage (not included, or finished).

I also had to delete this branch twice and do a hard reset of my master just to get rid of phantom commits in this pull request. =/


aphedox:
Based on what I see in IDA, all the fields in this packet should be uint32L except the continent id.

Fate-JH:
I think the game is being overly ambitious when it thinks it can handle this much population per zone.

pschord:
Another beautiful packet. Thanks!

PR #66, opened 2016-08-22
Added: ArmorChangedPacket

Fate-JH:
All 128 neatly folded changes of standard exo-suit.  It's like a cartoon where the character opens their wardrobe and has hangers and hangers of a single outfit.

ArmorChangedMessage is a simple four byte - opcode plus three in the data section - packet sent from the server to the client.


aphedox:
Based on IDA, what you have as the armor field should actually be two uintL(3) fields. Don't think there should be a further two bits after that.

Fate-JH:
I reformatted some of the known armor byte values.
`00 - 000 000 00 (Agile)`
`20 - 001 000 00 (Reinforced)`
`40 - 010 000 00 (MAX)`
`44 - 010 001 00 (AI MAX)`
`48 - 010 010 00 (AV MAX)`
`4C - 010 011 00 (AA MAX)`
`60 - 011 000 00 (Infiltration)`
`80 - 100 000 00 (Standard)`
Well dang.

`

pschord:
Awesome.

PR #68, opened 2016-08-27
Initial support for server REPL

L-11:


pschord:
I'm getting `[ERROR] PsLogin - Unable to initialize pscrypto`, which means we need to add the runtime paths to the sbt config for `console`.

PR #77, opened 2016-09-25
Packet: DismountBuildingMsg

Fate-JH:
To trigger this packet in sanctuary, go to the training area entrance doors, run through them until the prompt is encountered, and then select "Exit Virtual Reality."  The building GUID called out in the packet is that same spawn building's GUID (tested using login packets to verify).

I really don't know how to explain "dismounting" a building, especially since you don't go anywhere in the known test case, even if everything were rigged correctly.


pschord:
Please rebase and move to new test layout

PR #79, opened 2016-09-27
Added: BindPlayerMessage Packet

Fate-JH:
This is not necessarily a normal packet submission.  I'm actually looking for additional information.

I have numerous personal packet captures of the process of tranferring between different zones.  This packet serves some part of that system.  In every capture, however, the data always takes the same form:
`16 02 80 04 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00`
... regardless of origin, destination, or character.  For me, all entries always contain the same eighteen byte sequence of data after the opcode.

Does this match other people's captures?

Update:
PR re-opened.  See comments below.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -344,7 +344,7 @@ object GamePacketOpcode extends Enumeration {
     case 0x13 => noDecoder(CharacterNoRecordMessage)
     case 0x14 => game.CharacterInfoMessage.decode
     case 0x15 => noDecoder(UnknownMessage21)
-    case 0x16 => noDecoder(BindPlayerMessage)
+    case 0x16 => game.BindPlayerMessage.decode
     case 0x17 => noDecoder(ObjectCreateMessage_Duplicate)
     // 0x18
     case 0x18 => game.ObjectCreateMessage.decode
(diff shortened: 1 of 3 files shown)


Fate-JH:
I've found some captures that work differently.  I thought this packet might have to do with matrix binding but I didn't know if I had any evidence of that at the time.  Now that I do, I get to try my hand at figuring it out.

Fate-JH:
This demonstrates a re-imagining of the packet.  The only thing I am guaranteeing is that this method of packet encoding/decoding reflects and re-creates all of the samples packet captures I have been recently discovered.  Having said that, other than the sense of accomplish that I have actually completed another packet, I have two minor misgivings about this one due to other evidence.

In the decompiled source code, there is both an encoding function and a decoding function for `BindPlayerMessage`, though the encoding function is never called anywhere.  All packets contain a mandatory string section (even if it is an empty string) and that which follows the string is always 128 bits.  The encoding function would surely encode all 128-bits; but, the decode function is different in a varety of points, and I can not determine how it would ever reliably retrieve all 128-bits from the message created/transmitted.  The decode function extracts _1u_, _1u_, _4u_, _20u_, _20u_, _14u_, _14u_, and _10u_, the last three values turned into float numbers between 0 and 1.  The encode function proceeds, _1u_, _1u_, _4u_ _32u_, _32u_, _20u_, _20u_, and _16u_.  Those last three values once again turn into float values but, in this case, those values and the bitsize of the fields align with our existing `Vector3.codec_pos` map coordinate parsing methods.  For our purposes, I followed the encoding algorithm and wrote the codec to work with the packet as it is dispatched by the server.  I can only hope the client actually knows what it is doing.

The other issue is that I am uncertain where to precisely place a gap in between codecs.  In between the two _1u_'s and the _4u_ seems to be bits that are not used.  I believe the first two are correctly positioned but I do not know if the _4u_ is placed properly.  It is even possible that this region is actually flush (no skipped bits), but I can not determine this on my own.

pschord:
Please rebase and move to new test layout

Fate-JH:
I actually have a better version of this packet and tests for this packet in `PSF-Server`.  I'm going to close this PR and bring over the improvements.

PR #83, opened 2016-10-13
Added: GenericActionMessage Packet

Fate-JH:
I want to say that this is the smallest packet that I have worked on but the truth is that I already worked on the smallest packet.  That's AvatarJumpMessage at 9 bits (8 opcode + 1 data).

~~The only thing this packet has going for it is its mysterious nature.  It handles about twenty-seven actions, some of which may be on-off states of a specific thing, but it's going to be hard to track them all down.  To help with that, I've left the message logged in the window when for it comes across the pipe.~~

~~I'll experiment and update the commentary if I discover anything besides LFS.~~


Fate-JH:
I discovered a bunch of interesting behavior for this packet.  It's possible I might have been in the wrong stance/state/condition to activate some of the remaining changes from this packet.
The list distinguishes between actions caused by sending apacket from server to client and actions that resulted in sending a packet from client to server.

Fate-JH:
This PR will be superseded.

PR #87, opened 2016-10-22
Added: ZipLineMessage Packet

Fate-JH:
Once I was a Fraggle but now filled with scorn
To resecure with only my Suppressor drawn
Left in this wide cavern to die and respawn
Annoyed by a fodderling gone AFK
Oh, this base I was posted was covered in CE
And I tried all I knew to buy more time
But I never could resist the assault on the CC
From the CN on the sprawling zipline!


Fate-JH:
To be superceded by a different PR.

Issue #88, opened 2016-11-01
Docker build


adamlc:
It may be nice to have a Dockerfile so people can easily build the server, will probably make it easier for people to test too! I can have a stab at this over the next few days if you think it would be useful? :)

Fate-JH:
This could be useful, especially when we move from it just being just server application on its own and other details need to be set up.

PR #89, opened 2016-11-06
First attempt at Docker file

adamlc:
Heres my first attempt at a Dockerfile (see #88), unfortunately I'm unable to test logging in at the moment but it appears to run without any errors etc. 

The sbt deps are downloaded in the build process, so everything should run instantly once the image is built. It may be worth doing an automated build on docker hub so everything stays up to date :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,9 @@ EXPOSE 51000
 EXPOSE 51001
 
 # Download Login Server and pscrypto
-RUN git clone https://github.com/psforever/PSF-LoginServer.git /PSF-LoginServer && \
-    cd /PSF-LoginServer && \
+RUN wget https://github.com/psforever/PSF-LoginServer/archive/master.zip && \
+    unzip master.zip && rm master.zip && \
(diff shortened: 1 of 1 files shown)


adamlc:
At some point I think ill borrow the dockerfile content from the upstream image so we can ensure we have the correct versions for things. At the moment the upstream uses the versions we need but it would be nice to specify them. Ill have a stab at that in the next few days! Going to dust off my windows pc this weekend to do some testing ;)

pschord:
@adamlc Looks good. Can you download the repository ZIP instead of cloning? `https://github.com/psforever/PSF-LoginServer/archive/master.zip`

adamlc:
@pschord sorry for the delay, all done :)

pschord:
@adamlc This build doesn't appear to be working since the ZIP file change.
The `WORKDIR /PSF-LoginServer` line is being set incorrectly.

adamlc:
Apologies @pschord, my fault for not testing! I did this time, its all good :)

pschord:
Thanks!

PR #90, opened 2016-11-08
Added: ReplicationStreamMessage Packet

Fate-JH:
Who named this thing?

You would think something called "ReplicationStreamMessage" performed an incredibly technical operation when it really only seems to manage the list of squads.  It still may do something more; but, future decompilation aside, that's the only thing this packet did on live, so that is the only evidence of what we have it doing.  To its credit, the mysteriousness of the name belies the complexity of the format.  I have done a lot of pattern matching of the binary that goes into the data section.  While I can explain how to assemble each known form, I still don't know for certain how to disassemble all of the fields, especially the ones that explain what each entry does to the list of squads.  They are currently separated into what I think are logical groupings that seem consistent across the different forms and I have left some things in a state that hedges my bets should they have to change.

The class structure of this file is `A ( B ( C ( D ) ) )` proceeding from `ReplicationStreamMessage`, `SquadListing`, `SquadHeader`, and to `SquadInfo`, in that order.  `SquadInfo` is the only part I think is perfectly represented.  The rest is just the most convenient way I have found to represent all of the information in the known packets.  Starting from the bottom, it switches between `Codec`s progressively as it dives deeper into the nest, sometimes switching directly with `scodec` branching, and, finally, with a blatant `if...else` declaration to determine how to pull information from the "non-formulaic" parts of the data.  I wanted to depict the execution pathways with an ASCII drawing but I am not good at those.

The fun part is that the next parts of squad representation seem just as complicated as the first.  I thought this might be a good distraction from ObjectCreateMessage but it's really just a can of worms unto itself and I really need to go back ...

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -591,7 +591,7 @@ object GamePacketOpcode extends Enumeration {
     case 0xe3 => noDecoder(ZoneForcedCavernConnectionsMessage)
     case 0xe4 => noDecoder(MissionActionMessage)
     case 0xe5 => noDecoder(MissionKillTriggerMessage)
-    case 0xe6 => noDecoder(ReplicationStreamMessage)
+    case 0xe6 => game.ReplicationStreamMessage.decode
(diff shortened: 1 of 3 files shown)


Fate-JH:
This PR will be superceded.

PR #91, opened 2017-02-27
Some little corrections

SouNourS:


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -335,7 +350,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
     case msg @ SquadDefinitionActionMessage(a, b, c, d, e, f, g, h, i) =>
       log.info("SquadDefinitionAction: " + msg)
 
-    case default => log.debug(s"Unhandled GamePacket ${pkt}")
+    case default => log.info(s"Unhandled GamePacket ${pkt}")
```
pschord: Can you make this an error?

SouNourS: Done

PR #96, opened 2017-03-02
Fix for test leaking a DH crypto object

pschord:
Error was: net.psforever.crypto.CryptoInterface.CryptoDHState: class not closed.
memory leaked


Fate-JH:
Looks good.

PR #105, opened 2017-03-05
Packet: GenericCollisionMsg

Fate-JH:
Ouch!

The first field probably indicates the form of collision but I don't have enough information to make certain of that.

Those last few fields are supposedly called "impactForce," as suggested by [this extracted information](https://docs.google.com/document/d/1QPODlBmVXjKZhwqITKb4LCCERgviqmfGD1RHpfrFbjA/edit).  They may be set differently when the collision is between non-environmental aspects but, for things like fall damage, only the last field ever gets set to a number.  It's also a field that doesn't scale well into a value that makes sense, which is why I left it as a `Long` for now.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/game/GenericCollisionMsg.scala
+++ b/common/src/main/scala/net/psforever/packet/game/GenericCollisionMsg.scala
@@ -6,6 +6,26 @@ import net.psforever.types.Vector3
 import scodec.Codec
 import scodec.codecs._
 
+/**
+  * Dispatched by the client when the player has encountered a physical interaction that would cause damage.
+  * Being run over by vehicles and falling from an exceptional height (without hot drop) are common causes.
+  * @param unk1 na
+  * @param player the player or player-controlled vehicle
+  * @param target the other party in the collision
+  * @param player_health the player's health
+  * @param target_health the target's health
+  * @param player_velocity_x the x-component of the player's velocity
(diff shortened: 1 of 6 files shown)


common/src/main/scala/net/psforever/packet/game/GenericCollisionMsg.scala
```diff
@@ -0,0 +1,74 @@
...
+}
+
+object GenericCollisionMsg extends Marshallable[GenericCollisionMsg] {
+  implicit val codec : Codec[GenericCollisionMsg] = (
+    ("unk1" | uint2) ::
+      ("p" | PlanetSideGUID.codec) ::
+      ("t" | PlanetSideGUID.codec) ::
+      ("p_health" | uint16L) ::
+      ("t_health" | uint16L) ::
+      ("p_vel_x" | floatL) ::
```
pschord: Make a new Vector3 codec please.

Fate-JH:
`[info] Passed: Total 322, Failed 0, Errors 0, Passed 322`
`[error] (pslogin/compile:compileIncremental) Compilation failed`
`[error] Total time: 107 s, completed Mar 5, 2017 6:58:23 PM`
`The command "sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport" exited with 1.`
I didn't do it?

pschord:
Weird that this is failing.

PR #107, opened 2017-03-05
Packet: HitHint

Fate-JH:
In plain terms, this packet tells players from where they're being killed.

Apparently, any valid game object with a uid can be a "source of damage."

### DIFF
--- a/common/src/test/scala/game/HitHintTest.scala
+++ b/common/src/test/scala/game/HitHintTest.scala
@@ -11,9 +11,9 @@ class HitHintTest extends Specification {
 
   "decode" in {
     PacketCoding.DecodePacket(string).require match {
-      case HitHint(unk1, unk2) =>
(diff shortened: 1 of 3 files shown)


common/src/test/scala/game/HitHintTest.scala
```diff
@@ -0,0 +1,28 @@
...
+import net.psforever.packet._
+import net.psforever.packet.game._
+import scodec.bits._
+
+class HitHintTest extends Specification {
+  val string = hex"0A 460B 0100"
+
+  "decode" in {
+    PacketCoding.DecodePacket(string).require match {
+      case HitHint(unk1, unk2) =>
```
pschord: These fields are not unknown

Fate-JH: Fields are known.

PR #108, opened 2017-03-05
Packet: AvatarImplantMessage

Fate-JH:
Something something implants write better comments.

Right now, I honestly don't remember how I determined the data in this packet, save that I knew that the test was accurate for what I learned from it.  I was able to figure out which implants were which number in the last field and I was reliably able to determine that second wind was not represented in that group.  Implants need to be activated properly before testing on this packet can proceed.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/game/AvatarImplantMessage.scala
+++ b/common/src/main/scala/net/psforever/packet/game/AvatarImplantMessage.scala
@@ -1,0 +1,23 @@
+// Copyright (c) 2016 PSForever.net to present
+package net.psforever.packet.game
+
+import net.psforever.packet.{GamePacketOpcode, Marshallable, PlanetSideGamePacket}
+import scodec.Codec
+import scodec.codecs._
+
+final case class AvatarImplantMessage(player_guid : PlanetSideGUID,
+                                      unk1 : Int,
+                                      unk2 : Int)
+  extends PlanetSideGamePacket {
+  type Packet = AvatarImplantMessage
+  def opcode = GamePacketOpcode.AvatarImplantMessage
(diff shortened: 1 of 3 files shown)


common/src/main/scala/net/psforever/packet/game/AvatarImplantMessage.scala
```diff
@@ -0,0 +1,69 @@
...
+import scodec.Codec
+import scodec.codecs._
+
+/**
+  * An `Enumeration` of the available implants.
+  */
+object ImplantType extends Enumeration {
+  type Type = Value
+  val advanced_regen,
+      targeting,
```
pschord: Please keep enum fields in the form of `Enum.MyValue`. Not `Enum.MY_VALUE` or `Enum.my_value`

common/src/main/scala/net/psforever/packet/game/AvatarImplantMessage.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  implicit val codec = PacketHelpers.createEnumerationCodec(this, uint4L)
+}
+
+/**
+  * Change the state of the implant.<br>
+  * Write better comments.
+  * <br>
+  * Implant:<br>
+  * `
```
pschord: Have you confirmed that this displays correctly in ScalaDocs?

Fate-JH: ScalaDocs knows how to accommodate the <br> tag correctly.  It also handles the backtick mark being on its own line.

I'm not certain that it likes the rest of the packet, though.  When i run the command, it raises complaint about everything, especially `scodec` stuff.

PR #112, opened 2017-03-07
Packet: WeatherMessage

Fate-JH:
"I'm the Whether Man, not the Weather Man, for after all it's more important to know whether there will be weather than what the weather will be." (i)

As noted in the comments, this packet had its fields named through debug information currently archived [here](https://docs.google.com/document/d/1QPODlBmVXjKZhwqITKb4LCCERgviqmfGD1RHpfrFbjA/edit).  If ever, during testing, I come across any packet that makes it rain, or snow, or sandstorm, we'll have a better idea about how this packet factors into it, hopefully, maybe.

(i) Juster, Norton.  _The Phantom Tollbooth_.  Random House, 1961.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -495,7 +495,7 @@ object GamePacketOpcode extends Enumeration {
     case 0x92 => noDecoder(PlanetsideStringAttributeMessage)
     case 0x93 => noDecoder(DataChallengeMessage)
     case 0x94 => noDecoder(DataChallengeMessageResp)
-    case 0x95 => noDecoder(WeatherMessage)
+    case 0x95 => game.WeatherMessage.decode
     case 0x96 => noDecoder(SimDataChallenge)
     case 0x97 => noDecoder(SimDataChallengeResp)
     // 0x98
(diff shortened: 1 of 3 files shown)


Fate-JH:
Found sandstorms.  Will write more later.

SouNourS:
Can you make sandstorms on snow cont ?!

Fate-JH:
@SouNourS I don't think so, at least the packet doesn't seem to have enough detail to cover that kind of manipulation.  As far as I can gather, each continent has one type of weather associated with it based on clime.  Tropical and temperate continents have rain, desert continents have sandstorms, and frigid continents have snow storms.

I identified which fields are the `radius` and the `intensity` of storms but I've hit a brick wall when it comes to positioning them.  The scale does seem perfectly squared for coordinates based on a 8192 x 8192 continent.

Sending the following data:
`950100000000d0d74ac00b8485bf8911803eb8ce753d0800000074e3b93de234413f080a03267b3ff003443f441d2774a83eb5fc7a3f1d1ef344453e552eb83e5e43fad9e73eae3d8b3e51cd1fb7703ff01be13e2c41100d1c3fd47ddb3e6219c88b3a3fffce803e7655`
... will create a sandstorm on Ishundar that engulfs the southern plateua around Marduk.  The center of the storm is just very slightly northeast of Marduk's southern tower; you can make the storm very small to check that.  It allegedly creates other storms around the continent too but that seems to be the only one that is active.  I think that may have to due to the positioning of the cloud, but I don't know how to change that properly either.

Please ship this packet as-is for now.  We'll round back to it another day.

PR #116, opened 2017-03-09
Packet: ObjectDetachMessage

Fate-JH:
Everything in this game is a measure of "attachment" - the Beamer to your hand, the magazine to the rifle, the REK in your backpack, the player to the vehicle - for which this packet is the antithesis.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -364,7 +364,7 @@ object GamePacketOpcode extends Enumeration {
     case 0x24 => game.SetEmpireMessage.decode
     case 0x25 => game.EmoteMsg.decode
     case 0x26 => noDecoder(UnuseItemMessage)
-    case 0x27 => noDecoder(ObjectDetachMessage)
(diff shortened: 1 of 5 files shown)


Fate-JH:
I should have written "This packet is _sometimes_ considered formal response to a `DismountVehicleMsg` packet."

pschord:
are the Yaw, Pitch, Roll values similar to other found euler angle codecs?

Fate-JH:
@pschord 
A one byte angular measurement where `0x1` is equal to 2.813 degrees seems to be a standard across a few packets for various kinds of yaw rotation.  I have to go back to other places, though, and make doubly-certain that the zero'd cardinal direction is the same.  I have it listed as North in some cases, possibly incorrectly, and here as East.  I have all these details noted in respective packets' documentation sections.

An example of an outside case, for example, is `viewPitch` in `CharacterAppearanceData` where the valid angles get translated from the numbers 0 to 32 and the angles available are only apparent forwards-facing to downwards-facing for some reason.  All other numbers are just downwards facing.  That's different from `pitch` here which is more like `yaw` here.

There are also the `facing*` values in `PlayerStateMessage` where there's a wide range of numbers that are a dead-zone in between angles that look left or right or angles that look up or down.

PR #118, opened 2017-03-10
Packet: RepairMessage

Fate-JH:
I don't believe this packet works unless you have the appropriate Engineering certification.  That may indicate that it works closely with the progress bar that is visible when you are repairing something personally.

At the very least, nothing happens if this packet is dispatched to the client on its own.


pschord:
Can't wait to see some tests!

PR #119, opened 2017-03-10
Packet: Planetsideattributemessage

SouNourS:
It's @Fate-JH decode + change unknowns + add test file + add some incomplete informations.


pschord:
Fricken awesome. Where did you figure out which numbers went to specific attributes?

SouNourS:
@pschord Testing ! :1st_place_medal: 
Value after value with answer to a /t in game like : 
/t 01 1000
send : Planetsideattributemessage(my guid, 01, 1000)

PR #120, opened 2017-03-11
Packet: ObjectDetectedMessage

Fate-JH:
I do not yet fully understand what this packet does but I've found that it tends to follow after some `ObjectCreateMessage_Duplicate` packets and it appears often during continent loading, mixed in with `PlanetsideAttributeMessage`.  It probably plays some role involved in loading elements into the game world but currently I can not get it responsive (in a visual way).  There is another capture where the same packet with the same data appears twice with a span of 1000 entries, only a handful of seconds, not flanked by `OCM_D` or any form of continent loading.


pschord:
I believe this is just for placing "dot"s on the radar screen. I could be wrong though!

PR #128, opened 2017-04-20
Object Create Message Update #2

Fate-JH:
Just to remind that I had stated I planned for three or four of these large updates.  The first handled basic character-as-avatar packet parsing.  This, as the numbering indicates, is the second update and deals primarily with things that are not your character as an avatar.

Major technical changes are first are foremost.
I worked on object data primarily discovered in `0x17` `ObjectCreateMessage_Duplicate` packets.  Following information from this [extracted debug text](https://docs.google.com/document/d/1QPODlBmVXjKZhwqITKb4LCCERgviqmfGD1RHpfrFbjA/edit), the following class re-branding was performed:

- `0x18`: `ObjectCreateMessage` (old) --> `ObjectCreateDetailedMessage`
- `0x17`: `ObjectCreateMessage_Duplicate` --> `ObjectCreateMessage` (new)

In addition, all previous classes that were created and utilized by `0x18` have been changed to be prefixed with a `Detailed` descriptor, e.g., `WeaponData` (old) is now `DetailedWeaponData`, to make way for a new `WeaponData` to be used in `0x17` packets.  Due to perfect alignment between character data for both `0x17` and `0x18` character classes, a lot more information for creating avatars now understood, so please pay attention to those changes.  Please note where there is the rare re-use of classes between `0x17` and `0x18`, e.g., `CharacterAppearanceData` and `InventoryData`, etc., and do note that decoding and encoding follow a strict `OCM` / `OCDM` split.

As it is still decently early in development, this is probably the best time to make this change.  It can be seamlessly introduced in the normal repository but our Live Test Server (that's you, @SouNourS) may require a bit more work.

The primary purpose of `ObjectCreateMessage` (new) is to provide a lightweight version of `ObjectCreateMessage` (old) that can be used to create game elements without having to reference an extensive list of data points that the client doesn't care about in the given context.  One client doesn't care about another player's list of "first time events," ever, or whether or not they have a medkit hidden in their knapsack.  In addition, `ObjectCreateMessage` (new) serves as the origin-point for all other in-game elements that are not the avatar, including all deployables and all vehicles.  Experimentation has proven that `0x18` and `0x17` object-structures can be swapped and still handled appropriately but, since the mechanism is unknown, their decoded classes have been segregated at the packet level here.  Four major things exist in this update, some which can already be utilized as-is, and others which need to be ironed-out by future packet work that extends beyond `OC*M`.

1. A second method of encoding "essential" character data is introduced that works best for depicting "other players."  **Quite a variety of appearances and alternate posing is now possible**.
2. **Items can now be spawned as "already dropped on the ground"** without having to be detached from an existing container.  Following from this ...
3. **All Engineering deployables are now available for spawning**.  Different items will have different levels of completeness to their analysis.  All will require additional packets to render them fully functional but some things should be serviceable as they are.  The Aegis Shield Generator should serve as a functional in-the-field concealed waypoint without needing further work, though creating a ASG with both radar, shielding, and resupply on spawn is going to take a lot more effort.
4. **Miscellaneous game objects have been decoded** including implant terminal components, spawn tubes, matrix panels, a lattice logic unit (originally to Qumu), and some projectiles that are synchronized between clients due to operation and control.  Most of this is essentially useless for the moment but are important building blocks for later.

The next major introduction, the third `ObjectCreateMessage` update, will involve every vehicle in the game, and continued refinement to existing classes.  I expect to preface the larger update with a smaller update granting access to an ATV and an early AMS for Live Test Server to incorporate.

### DIFF
--- a/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
+++ b/common/src/main/scala/net/psforever/packet/GamePacketOpcode.scala
@@ -345,9 +345,9 @@ object GamePacketOpcode extends Enumeration {
     case 0x14 => game.CharacterInfoMessage.decode
     case 0x15 => noDecoder(UnknownMessage21)
     case 0x16 => game.BindPlayerMessage.decode
-    case 0x17 => noDecoder(ObjectCreateMessage_Duplicate)
+    case 0x17 => game.ObjectCreateMessage.decode
     // 0x18
-    case 0x18 => game.ObjectCreateMessage.decode
+    case 0x18 => game.ObjectCreateDetailedMessage.decode
     case 0x19 => game.ObjectDeleteMessage.decode
     case 0x1a => game.PingMsg.decode
     case 0x1b => noDecoder(VehicleStateMessage)
--- a/common/src/main/scala/net/psforever/packet/game/ArmorChangedMessage.scala
+++ b/common/src/main/scala/net/psforever/packet/game/ArmorChangedMessage.scala
@@ -2,6 +2,7 @@
 package net.psforever.packet.game
 
 import net.psforever.packet.{GamePacketOpcode, Marshallable, PlanetSideGamePacket}
+import net.psforever.types.ExoSuitType
 import scodec.Codec
 import scodec.codecs._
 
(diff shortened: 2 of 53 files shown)


aphedox:
👍 impressive progress

pschord:
Glad you have taken point on this complicated packet. Tons of good improvements and documentation!
This truly is the god packet.

I'm unable to give this a good review as it is so extensive, but I trust your code and we're adding functionality, so that's great.

PR #129, opened 2017-04-20
Packet: ChildObjectStateMessage

Fate-JH:
Children are usually prohibited from being depicted in violent video games, but we're edgy.

Used by properly outfitted turrets and stuff for reasons similar to `PlayerStateMessageUpstream`.  (I tested using a self-spawned Orion.)


Fate-JH:
Temporarily rescinding this submission for further analysis.

Fate-JH:
Still suspicious but will have to wait for later.  Restoring.

PR #134, opened 2017-04-25
Packet: DeployableObjectsInfoMessage

Fate-JH:
I wish this packet were more-aptly named to match its purpose.  One can't determine that it creates or removes an icon of the deployable on the continental map by just looking at it in a list.


Fate-JH:
Speculative development rant follows:
What really stuns me is how the game tries to keep track of ownership using player GUID.  That number is prone to changing whenever players respawn, and that seems to be the reason why leaving a continent for too long can cause CE to become expatriated from the layer upon return.  I'm flabbergasted no one ever encountered an issue where they gained access to CE that wasn't their own over a period of respawns.  How does the client even track ownership between respawns?  Also, would that not also cause additional GUID drain for the life and the decay period of a deployable, beyond just that deployable object's own GUID?

I can't imagine this system as anything but extremely tempermental.

PR #137, opened 2017-05-02
Packet: TargetingInfoMessage

Fate-JH:
Crams numbers down the target's throat.

This packet can be used to update `0x17` object health and armor/shielding values that are normally valued between 0 and 255, or a scaled 0f to 1f value.


Fate-JH:
@pschord The tests on this build wouldn't complete the first time.  After restarting the tests, it passed.  I doubt it's the same issue as before with crypto files, however.

PR #146, opened 2017-05-06
Add interactive console support for server admins and devs

pschord:
This is a prototype for console support for administrating the server.
It's an early version and serves as a proof of concept. Many things are missing such as:

* An easy way of adding commands hierarchcally (abuse scala syntax maybe, pickup python argparse)
* Usage statements for commands
* Error handling for more advanced commands
* The logging from the console will overwrite the readline call. I have a fix for this, but I need to hook in the logback ConsoleAppender
* We need a standard way of communicating with actors in order to avoid hardcoding actor paths
* We need a well-defined way of defining an interface individual service APIs (e.g. SessionRouterAPI)
* Thoughts should be made about how we could make these APIs scale to another interface such as REST+JSON (would be better than doing database SELECTS or writing out to a file)

Some available commands:
```
quit, exit, shutdown - stops the server
session list - shows all world and login sessions
session drop-world <id> - drops world session by id
session drop-login <id> - drops login session by id
```

Still this a good first attempt!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,6 +42,11 @@ lazy val root = (project in file(".")).
 
 lazy val pslogin = (project in file("pslogin")).
   settings(commonSettings: _*).
+  settings(libraryDependencies ++= Seq(
+      "org.jline" % "jline-terminal" % "3.2.0",
+      "org.jline" % "jline-reader" % "3.2.0"
+    )
+  ).
   settings(
     name := "pslogin"
   ).
--- a/pslogin/src/main/scala/PsConsole.scala
+++ b/pslogin/src/main/scala/PsConsole.scala
@@ -1,0 +1,76 @@
+// Copyright (c) 2017 PSForever
+import org.jline.reader._;
+import org.jline.terminal._;
+import akka.actor.ActorSystem;
+
+import org.slf4j.LoggerFactory;
+import ch.qos.logback.classic.spi.ILoggingEvent;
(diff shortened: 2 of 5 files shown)


Fate-JH:
This is a good foundation to build upon.  Some things during testing:

1-  According to the startup, Scala was unable to create a system terminal and created a "dumb terminal" instead.
`May 08, 2017 8:25:07 PM org.jline.utils.Log logr`
`WARNING: Unable to create a system terminal, creating a dumb terminal (enable de`
`bug logging for more information)`
Is this correct?

2- The custom prompt is mixing in with the normal log output and that produces interesting results.
psconsole-prompts
It also doesn't always show itself.

3- I passed [blankspace] as input and caused an error.
psconsole-spacebar

4- I normally stop the server with a `Ctrl+c` but that throws a `NullPointerException` in `PSConsole.scala:line#28`.  This is probably just a consequence of the `PSConsole` input also using the command prompt's stdin.

SouNourS:
First test. Great ! 
I'll test that a little more ! (maybe on live)

console

mysteriousstranger:
tbh my famalam this is breddy gay :^)

pschord:
I dont think this is the right approach for now. We should develop a remote TCP API and management client. I tried to make it so that the prompt wouldn't be hidden in another patch, but this just gets lost in all of the debugging output.

PR #163, opened 2017-09-24
Different Zones

Fate-JH:
But first we need to talk about parallel universes.

In PlanetSide, the spawnable universe is divided into 35+ physically distinct arenas:
- ten full continents
- six caverns
- four battle islands
- three faction sanctuaries
- three driving training zones
- three shooting training zones
- three unlisted training zones
- three unimplemented orbital stations
- one missing black ops sanctuary
- one missing team deathmatch arena
- an unlimited number of unloaded mapless regions best described as "nowhere."

Until recently, this universe was consider all part of the same region in terms of `WorldSessionActor`.  Going to one region was the same as going to any other region and anything that could inhabit a "space" inhabitted all "spaces."  More recently, a distinction was defined thus that different regions as tagged "zones" loaded different groups of players and had some different logic applied to entities in that zone (thanks, @SouNourS).  All the same, this was an abrupt conditional check in `WorldSessionActor`.  Now, the regions are physically separated into associated items that exist only in that specific `Zone`.  The amount of overhead in this update is nowhere near as significant as the amount of preparatory overhead incurred in the last update.  Essentially, a lot of the changes necessary to support these different `Zones` had already been made before they were ready for implementation.  Nifty, huh?

To demonstrate, the ability to drop or pick up `Equipment` has been shifted into these `Zones`.  Likewise, an example of server objects such as `Terminal`s exclusive to a specific `Zone` has also been provided.  Compare the three working order terminals in the Hart C spawn building on `home3` (immediate right upon spawning) to other, unresponsive terminals on `home3`.

Note that the method for transporting between different `Zones` has not yet been implemented.  Only `home3` has been prepared for the example and any attempt to switch zones without defining other places should end poorly; but, the method for adding the remainder of the continents should be easy to figure out.  We're at the threshold of the point where having all of the environmental GUIDs for the different continents would be helpful.  No tests yet as things are still extremely fluid.


Fate-JH:
world startup
Something like this.

PR #164, opened 2017-09-28
Unique Number System

Fate-JH:
Since it's always somewhere 0 through 65535, it isn't very unique when you think about it.

The system now supports multiple pools via the `UniqueNumberSystem` class.  By default, any `Zone` has only two pools defined ("environment" and "dynamic") and one is always defaulted ("generic").  "Environment" just assumes all objects normally a part of the `Zone` are below 2000.  "Dynamic" is 2001 to 10000 randomly.  "Generic" encompasses any number not a part of the established pools; however, it should be considered sparing as it is merely a catch-all.  `NumberPool`s can only be set before startup.

Hopefully the last iteration of the global unique identifier system, save for setting up `Routing` workers later and deciding on an ideal distribution of available GUIDs between different pools.  Perhaps a space-saving optimization in the number-to-pool matching function ...


Fate-JH:
Actually, scratch that.  @SouNourS, you will reconfigure `UniqueNumberSystem` using `Routing` similar to how `TaskResolver` is setup if you port any of this over to Test Live.

PR #166, opened 2017-10-02
Packet: AvatarImplantMessage changes

SouNourS:
https://streamable.com/t5pxh


SouNourS:
Hmm, I need to test a little more that **** !
It seems the parameter I defined as "status" is 0 or 1 with "action" 2 and 3 (init and activate), but ... if I look packets, with "action" 0 and 1 (add and remove)... that's look like the implant number... Like @Fate-JH defined that packet before I change it....

I'm not sure to be understandable, I have some whisky in blood :dagger:

Fate-JH:
Not impossible since the fields are both the same bit length.

You could easily split it if this proves to be the case. `None`s are free.

`final case class AvatarImplantMessage(PlantSideGUID, Int, Option[ImplantType.Value], Option[Int])`

SouNourS:
I updated the informations, but I have no idea how change 

final case class AvatarImplantMessage(player_guid : PlanetSideGUID,
                                      action : Int,
                                      implantSlot : Int,
                                      **status : Int**)

to

final case class AvatarImplantMessage(player_guid : PlanetSideGUID,
                                      action : Int,
                                      implantSlot : Int,
                                      **status : ImplantType.Value**)

only when **"action" is 0**, all other cases use the Int.

I looked other packets to see possibilities, but I dont see how :(

Fate-JH:
I'll merge this PR for now as I am not working on the implant terminal or the packet right now.  I'll let @SouNourS's judgment to this one.

PR #171, opened 2017-10-10
Login Fix

Fate-JH:
Copied from Discord:
"Normally, on the login screen, the server would send the much longer `0x18 OCDM` avatar packets to list characters.
"I've been sending the shorter `0x17 OCM` avatar packets, forgetting about the need for crucial BEP information to get the entry's rank.(edited)
"I guess whatever is pulling data from the packet in the background pulls something out of the packet regardless of which type [of packet] it is and thinks that should be BEP data."

A special "LoginConverter" could probably be created to allow only the relevant `OCDM` information to be sent during login, rather than bloat the packet with data that is not useful to displaying the login character.  That's why I thought I could get away with sending `OCM` instead of `OCDM` in the first place.


Fate-JH:
Superceded by #173

PR #174, opened 2017-10-26
Quick Fixes

Fate-JH:
Adds two doors to home3.

Resolve issue with `TaskResolver` involving complicated tasks either not completing or happening out of order, resulting in failure of the primary task.  This was usually encountered during registering the GUID for the player's avatar by hanging in between the task of getting the zone for the client's `WSA` and loading the zone's map on the client.  Caused a mostly blank screen and no problems in the log.

Minor correction involving the fallback case for dropping items on the ground (fortunately, should be rare to ever hit).


Fate-JH:
I do need to write unit tests for `TaskResolver` at some point and establish "best practices" for building `Task` objects to hand-off to `TaskResolver`.

PR #176, opened 2017-11-26
Vehicle Terminals

Fate-JH:
Allows for spawning of all vehicles save the Switchblade, the Flail, and the Vulture, the former two because of cavern connection requirements and the latter because reasons of other requirements needing to be fulfilled.  Of course, BFRs are still out of scope.  Anyway, just pull a Liberator and call it a "Vulture," m`kay?

The test player character has been re-positioned for the purposes of this update, still in the college of the VS sanctuary HART C buildings, but between two spawn pads.  Either of the vehicle terminals - the one to the player's immediate right for ground vehicles and the other to the player's immediate left for air vehicles - can be used for spawning.  The actions involved in vehicle spawning are synchronized between clients so play with multi-boxing.

Note that the vehicle spawning process is only half realized.  As the vehicle can not be attached to the spawn pad internally, players will see the animation of mounting the vehicle as normally depicted as opposed to how it is normally obscured while below the ground on the spawn platform.  Additionally, the auto-drive capabilities of a newly-spawned vehicle are not ready.  To preserve the functionality of the spawn queue, a timed deconstruction if a vehicle loiters on the spawn pad for too long and two chat warnings have been implemented.  The message player should receive about being in such and such a position in the queue is not included.  Analysis of the `ChatMsg` `@` strings will be required for that.

Minor modifications have been performed to `Tool` objects.

The only new packet included is `GenericObjectActionMessage`.


Fate-JH:
According to Travis, the size of the `init` for `GlobalDefinitions` exceeds the code size limits of the JVM; however, this just means that Travis has the issue, as the code can be compiled and executed normally without issue.  I would recommend opening a command prompt or terminal in the PSF-LoginServer folder of one's local git repo and run `sbt test` once or twice to confirm that tests pass.

Unfortunately, due to the Travis issues, code coverage can not be statused.  I'll work around this at a later date.

Fate-JH:
"52.57% of diff hit (target 53%)"
Oh, come on!  Just round up!

PR #183, opened 2018-01-15
Pew Pew

Fate-JH:
Upfront caution: the following update takes resources from one's personal inventory for use in weapons.  This update includes support for: weapon fire, reload, change ammo, change fire mode, weapon dry fire.

Weapon fire:
Each shot taken expends one round from the weapon's internal magazine.  Dropping the weapon and picking back up will not (no longer) reset the magazine contents and the current capacity will be carried between players.  Weapon fire is synchronized between clients for handheld weapons and for vehicle weapons.  There is a safety condition where the server-side weapon will attempt to force the local weapon to coincide if it gets the order to shoot but the magazine should be empty.

Reload:
When the weapon is less than completely full, a player will reach into their backpack to find a sufficient amount of ammunition of the same type to put into the magazine.  Weapon reload is synchronized between clients for handheld weapons and for vehicle weapons.

Change ammo:
Players will search through their backpack for other ammo in the order that is indexed for the given fire mode in `GlobalDefinitions`.  Ammo may be dropped if it does not fit into backpack during the swap but all efforts will be expended to attempt to distribute the displaced ammo among the existing boxes of that ammo type already in the backpack.  Dropping a weapon and picking back up should keep the same ammo type in the current fire mode.  Change ammo is synchronized between clients for handheld weapons and should be synchronized for vehicle weapons (but no vehicles actually require this functionality). As was the case on Gemini Live, ammo type is preserved when getting a weapon from a Favorite loadout.

Change fire mode:
Fire mode is now preserved across weapons.  Weapons having the different ammunition types in alternate fire modes (e.g., the Reaver rockets still showing up as nosegun bullets) are now correctly assigned the proper ammunition type in the given fire mode.  Dropping the weapon and picking it back up should keep the same fire mode which is necessary to keep the server object and the local object aligned.  Fire mode encoding was finally granted to `0x18` `ObjectCreateDetailed` packets as was necessary.  Change fire mode is synchronized between clients for handheld weapons and for vehicle weapons.  As was the case on Gemini Live, fire mode is not preserved when getting a weapon from a Favorite loadout.

Dry fire:
It emits characteristic noise if the server-side weapon's magazine is empty.  Weapon dry fire should be synchronized between clients for handheld weapons and for vehicle weapons; but, I can not confirm that myself because multiboxed PlanetSide clients pollute each other's audio settings.  This will need to be tested on DevTest Live.  It does not crash, at least.

Caveats:
- An other-clients crash, which had mass client crash potential, was occurring during testing due to some issue when dropping and picking up a weapon betwen ammunition swappings.  As the original reason for the crash is foggy, and the change that seemed to correct the issue does not reveal *why* it should have resolved the issue, stay guarded.
- To accommodate testing, all players have been temporarily issued a Punisher with an assortment of Punisher ammunition upon spawn.  The ammunition has been adjusted so the weapon can swap ammunition without being forced to drop any (happens for the default loadout too).
- I have not tested all weapons.  The Punisher and the Rocket Rifle are the only handheld weapons tested and the Lightning and the Reaver are the only vehicle weapons tested for a measure of consistency.
- Some weapons may see a minor issue where ammunition dropped or returned to the inventory is larger than the normal capacity of a single box.  In the case of the MCG, that'll be 100 9mm (AP) in one box that is normally 50 each.  This will need to be corrected at a later date but the effort is more trouble than it is worth right now.
- There is little to no feedback for successfully landing shots or for being shot yet.


Fate-JH:
A special pass was made for many Infantry weapons to determine functionality.  Buckshot-style weapons that shoot pellets, mostly the NC, will benefit from a fire mode that supports the number of pellets in the chamber per round fired.

The Decimator does something interesting during its last shot by not generating a `ChangeFireStateMessage_Start` packet, which means other players would never see the final rocket.  The last packet has to be sent manually based on our model.  On Gemini Live, there was probably a difference in how each weapon treats its fire mode and what kind of packet it generates.  Quite a few weapons generate `WeaponFireMessage` packets before their `ChangeFireStateMessage_Start` or `ChangeFireStateMessage_Stop` reach the server.

The Decimator and grenades now clean up after themselves after they are expended.  The code for grenades, however, just deletes the now-empty grenade item in the sidearm holster rather than pushing a new packet of grenades into the same slot.  The current functionality is for the sake of convenience only.

Cavern weapons are all sorts of screwy.  Just let them do whatever for now.

@SouNourS, the dev branch would benefit from the work you did with the `LashMessage` packet.

SouNourS:
Ok, will do that 👍

PR #185, opened 2018-01-25
Doors Open from Inside

Fate-JH:
Stop locking yourself inside.

Doors that can be hacked should define directionality in terms of an angle.  (Remember, the angle system we are pushing towards is a North-corrected one that mimicks compass directions - North being 0 / 360 and East being 90.  Natively, angles in the game are laid out like on a graph.)  This directionality is aimed "towards the outside" of the door.  Anything that tries to open the door from the inside may do so freely which is the point of this PR.  Anything outside obeys the normal ownership and locking rules.

Doors will also make a sincere effort to not close on the last person who opened it if that person is standing a certain distance from it, outside or inside.  Doors will not currently inherit another player's presence until that other player also attempts to open the door, meaning it needs to close again.

A REK has been returned to test characters in place of their sidearm.  The tower that is hacked by the NC also has all four doors properly set up to be opened and hacked.

Caveat:
- If a door closes on the player just right, it may lock with them within the door mesh, with view on both parts of the map, and one of the two parts - the inside of the building or the outside of the world - will be culled visually.  If this happens to you, merely step forward or backward to unwedge yourself for now.  Stepping into the culled level will restore it.
- Doors had some method of transferring "openness" between players thus that they wouldn't close so easily on Gemini Live.  This mechanism is not imagined and so, as mentioned above, a door must close and open due to another player before it may wait on the proximity of that next player.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
+++ b/common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
@@ -4,28 +4,58 @@ package net.psforever.objects.serverobject.doors
 import net.psforever.objects.serverobject.PlanetSideServerObject
 import net.psforever.objects.Player
 import net.psforever.packet.game.UseItemMessage
+import net.psforever.types.Vector3
 
 /**
   * A structure-owned server object that is a "door" that can open and can close.
   * @param ddef the `ObjectDefinition` that constructs this object and maintains some of its immutable fields
   */
 class Door(private val ddef : DoorDefinition) extends PlanetSideServerObject {
-  private var openState : Boolean = false
+  private var openState : Option[Player] = None
+  /** a vector in the direction of the "outside" of a room;
+    * typically, any locking utility is on that same "outside" */
+  private var outwards : Vector3 = Vector3.Zero
 
-  def Open : Boolean = openState
+  /**
+    * While setting the normal rotation angle for the door (?),
+    * use the angular data to determine an "inside" side and an "outside" side.<br>
+    * <br>
+    * Doors are always positioned with the frame perpendicular to the ground.
+    * The `i` and `j` components can be excused for this reason and only the `k` component (rotation around world-up) matters.
+    * Due to angle-corrected North, add 90 degrees before switching to radians and negate the cosine.
+    * @param orient the orientation of the door
+    * @return the clamped orientation of the door
+    */
+  override def Orientation_=(orient : Vector3) : Vector3 = {
+    val ret = super.Orientation_=(orient)
+    //transform angular data into unit circle components
+    val rang = math.toRadians(orient.z + 90)
+    outwards = Vector3(-math.cos(rang).toFloat, math.sin(rang).toFloat, 0)
+    ret
+  }
 
-  def Open_=(open : Boolean) : Boolean = {
+  def Outwards : Vector3 = outwards
+
+  def isOpen : Boolean = openState.isDefined
+
+  def Open : Option[Player] = openState
+
+  def Open_=(player : Player) : Option[Player] = {
+    Open_=(Some(player))
+  }
+
+  def Open_=(open : Option[Player]) : Option[Player] = {
     openState = open
     Open
   }
 
   def Use(player : Player, msg : UseItemMessage) : Door.Exchange = {
-    if(!openState) {
-      openState = true
+    if(openState.isEmpty) {
+      openState = Some(player)
       Door.OpenEvent()
     }
     else {
-      openState = false
+      openState = None
       Door.CloseEvent()
     }
   }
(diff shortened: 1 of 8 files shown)


common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
```diff
@@ -92,4 +122,25 @@ object Door {
...
+    * Instantiate and configure a `Door` object that has knowledge of both its position and outwards-facing direction.
+    * The assumption is that this door will be paired with an IFF Lock, thus, has conditions for opening.
+    * @param pos the position of the door
+    * @param outwards_direction a vector in the direction of the door's outside
+    * @param id the unique id that will be assigned to this entity
+    * @param context a context to allow the object to properly set up `ActorSystem` functionality
+    * @return the `Door` object
+    */
+  def Constructor(pos : Vector3, outwards_direction : Vector3)(id : Int, context : ActorContext) : Door = {
+    import akka.actor.Props
```
xTriad: I've been seeing imports like this all throughout the codebase. Why not import them at the beginning of the file?

common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
```diff
@@ -92,4 +122,25 @@ object Door {
...
+  /**
+    * Instantiate and configure a `Door` object that has knowledge of both its position and outwards-facing direction.
+    * The assumption is that this door will be paired with an IFF Lock, thus, has conditions for opening.
+    * @param pos the position of the door
+    * @param outwards_direction a vector in the direction of the door's outside
+    * @param id the unique id that will be assigned to this entity
+    * @param context a context to allow the object to properly set up `ActorSystem` functionality
+    * @return the `Door` object
+    */
+  def Constructor(pos : Vector3, outwards_direction : Vector3)(id : Int, context : ActorContext) : Door = {
```
xTriad: Why are the first two parameters separated from the last two?

Fate-JH: This is a form of currying.  A typical function can be defined as `(A, B) => C`, where `A` and `B` are the parameter types to a function and `C` is the returned type.  When you do `func(A, B)`, it executes and produces a `C`.  Separating the parameters like this defines an `A => B => C` function - an `(A) => ((B) => C)` function, if you will - that can accept an `A` parameter and outputs a `(B) => C` function.  Its execution has been deferred.

What I am essentially doing is providing the function with some of its parameters in a place where it can reliably expect those parameters and then putting the function somewhere until I need it later.  What I am waiting on is an appropriate `ActorContext` so the actor of this object is created under a particular actor hierarchy.  I have not yet built the basic object because I can make a new object each time I call this function which will be useful to avoid some config redundancy (zones using the same map).

common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
Fate-JH: Importing resources like this limits visibility to those resources to the scope of the containing code block.  The same way one creates local variables close to the relevant place where they are used and then lets them be discardedwhen they fall out of scope, resources can be locked to specific regions where they are applicable and while they are applicable, or, in certain cases, close to where they will be referenced to reduce ambiguity of the references.

xTriad: Interesting, and since these doors are created once when the server starts, putting the `import` within the scope of the Constructor function means it could be discarded after they are all created? I'm assuming the JVM is smart enough to "release" `akka.actor.Props` if it's no longer needed, although that would not happen here because `Props` is used extensively throughout the codebase. You put it there to make it more readable.

common/src/main/scala/net/psforever/objects/serverobject/doors/Door.scala
xTriad: Ah, I was wondering if that was a curry. At my job I have been working with React.js related libraries that use currying such as Redux and Recompose (if you are familiar with all the Javascript craziness), but I was not familiar with this Scala syntactic sugar.

So this lets you go ahead and create a `Door` before you have the appropriate `ActorContext`, but more importantly, it let's you easily create multiple instances of a `Door` with the given `pos` and `outwards_direction` with different `ActorContext`'s?

Fate-JH: Yes.  Consider the VR training in the game - the same map layout for different faction-separated zones.  In our case, we'd use the same map to set up VR driving zones `tzdrtr`, `tzdrnc`, and `tzdrvs`.

I could easily hardcode everything into the `Zone` objects and make the `ZoneMap` class and the currying superfluous.  There's only nine normal zones - three layouts - that would feature the duplication.  From a purist perspective, that would be the most definite of forms because it would be hard coded.  I'm holding out for the enthusiasm to write/include a proper no-compile-file configuration method, such as something with XML, however.  That's why the main point for me is the ability to separate use of more primitive configuration data from use of the Actor system wiring.

I feel I have to prepare for the ambitious people who want to expand on the game after we finally get to vanilla PlanetSide Classic.

PR #186, opened 2018-02-08
Buildings and Stuff

Fate-JH:
No changes here.  Nada.  None.
You should see no changes to how things work.

This is the stalled build I had been working on for a release related to object initialization functionality and it would help later with hacking and base capture and such.  No, it can't do that yet.  Mounting and Dismounting code has changed and so has the location of some the some faction checks.  In general, though, you shouldn't see any real change since this was just the setup elements for the eventual initialization process.  The main things that need to be known, on the techical side, are that the `Base` class has been replaced by the `Building` class and many `PlanetSideServerObject` objects go onto becoming `Amenity` objects before being installed into `Building` objects.

We now have 1012 tests, some of them useful.


common/src/main/scala/net/psforever/objects/serverobject/structures/Amenity.scala
```diff
@@ -0,0 +1,60 @@
...
+  def Owner : PlanetSideServerObject = owner
+
+  /**
+    * Set an object to have a direct association with (be superior to) this one.
+    * @see `Amenity.AmenityTarget`
+    * @param obj the object trying to become associated as this object's "owner"
+    * @tparam T a validation of the type of object that can be an owner
+    * @return the object associated as this object's "owner"
+    */
+  def Owner_=[T : Amenity.AmenityTarget](obj : T) : PlanetSideServerObject = {
```
xTriad: Why is this syntax used?

Why not use `def Owner_=(obj : Amenity.AmenityTarget) : PlanetSideServerObject = {`

Edit: I think I get it. You're using the generic type T, but you're saying that although T is generic, it still needs to be a subtype of Amenity.AmenityTarget.

common/src/test/scala/objects/BuildingTest.scala
```diff
@@ -0,0 +1,169 @@
...
+      val reply = receiveOne(Duration.create(100, "ms"))
+      assert(reply.isInstanceOf[FactionAffinity.AssertFactionAffinity])
+      assert(reply.asInstanceOf[FactionAffinity.AssertFactionAffinity].obj == bldg)
+      assert(reply.asInstanceOf[FactionAffinity.AssertFactionAffinity].faction == PlanetSideEmpire.VS)
+      assert(bldg.Faction == PlanetSideEmpire.VS)
+    }
+  }
+}
+
+class BuildingControl3Test extends ActorTest {
```
xTriad: You get a gold star for these tests. This is great, really helps me understand how amenities work!

PR #188, opened 2018-02-18
AMS Deployment

Fate-JH:
Parking fees will be levied.

Caveats:
- Although this update is specifically for AMS vehicles, other deployable vehicles can use the same logic and even have token representation in the code.  They have not been verified, however.  Deploy anything other than the AMS at your own discretion.


Fate-JH:
The pslogin tests are still not represented in coverage reports.  Perhaps I need to point codecov at the directory in the yml file?  I can't wrap them in a normal Specification test, can I?

PR #196, opened 2018-04-02
The Real Resurrection Update

Fate-JH:
Still just in time for Easter.

The short of it is: base-installed spawn tubes and spawn locations and corpses.  Spawn tubes can now be operational and just the existence of the established tubes themselves will allow the server to mark spawn points for the user.  I say "predict" because the client and the server never actually communicate with each other regarding where the player is going to spawn.  Instead, the client tells the server something like "closest tower" or "closest base."  At the moment I can only narrow down those two options with any certainty.  This makes the prediction algorithm relatively simple - shortest distance and then randomly select any spawn tube.  Much like locked doors, however, that means more work that usual is required for set up.  Spawn tubes must have their position and, optionally, a specific facing direction defined to work correctly.

Here's your testing regiment.
Spawn tubes:
For this update, the default spawn point has been moved to Ceryshen, the Anguta and Akna region.  Anguta, Anguta Watchtower, and Akna Air Tower have had their amenities sorted for the VS.  Since damage is not yet implemented, I recommend using `/suicide` to hop about.  Kill yourself once and choose between Akna tower and Anguta base.  If you go to Akna tower, subsequent spawns will always include that tower or Anguta.  If you go to Anguta, however, your future spawn options will always be either Anguta tower or Anguta base.  Moving back East towards Akna will make Akna tower available again.  (Yeah, I just described how proximity works and you read it.)

The option to spawn back on the VS sanctuary continent is also available; the only spawns available are the spawn tube building in the HART C campus.  You know the building.  If you keep killing yourself and respawning, you will discover that you will appear in all of the tubes.  You can also use the deconstruction option from a tube directly by walking into one and following the onscreen prompting, though remember that the option isn't available from all tubes.

The timers are set: 2 minutes to choose to release before going back to sanctuary automatically; 30 seconds respawn to the same continent; and, 0 seconds for switching continents.

Corpses:
Informally, I call a dead player that can be resurrected on the spot a cadaver and a dead player who is lootable a corpse.  In any case, in between dying and coming back, notice that you are now leaving copies of former lives around on the ground.  (Other players will be doing likewise.)  Walk up to the corpses and loot their stuff.  Equipment can, of course, also be transferred onto the corpse too.  Eventually, the corpse and everything still on it will despawn on its own ("decay").  If looted sufficiently, to the point of being emptied-out of all its available equipment, the corpse will manually be despawned a few seconds after closing the window.

The timers are set: 3 minutes for normal corpse decay, or 1 second after leaving no equipment on the corpse and closing the window.

Vehicles:
Since Anguta is online, more or less, you can pull vehicles at the terminals.  If you `/suicide` in a vehicle, you will be able to respawn following the previous guidelines.

"Go kill yourself," is what I'm saying.

Caveats:
- The spawn tubes in sanctuary are finicky and I believe that is due to more cramped collision on the physical models.  Standing in the wrong location can cause the tubes to forcibly eject the player or throw them from side to side or even completely spin them around to face the inside of the tube upon spawning.  I spent a good few hours comparing coordinates before settling on things that seemed to work well enough.
- Like vehicle trunks, corpses do not have prioritized access control yet.  Please don't fight over game items.
- If the corpse is being looted when it despawns, the game will insist on destroying any item in your hand if it came from the corpse.  You can avoid this by droppng it or putting it into your own backpack.  This is a willful choice of the game engine - it actually sends a `RequestDestroyMessage` packet just for that item - so I have left it do that.
- If you leave Ceryshen, there is no way back short of relogging.
- The only reason sanctuary travel is possible is because warping back to sanctuary is instituted as the fallback destination if your spawn-search parameters fail to find a spawn somewhere.  Failing to spawn in sanctuary (should) result in the game gracefully disconnecting you.
- Your cadaver will normally stick out of the side of the vehicle in a quiet sitting pose when dead, resulting in an awkward visual that was covered up by making the player model invisible.  When the cadaver is kicked from their seat, he will manifest under the vehicle, let out a delayed death cry, and will fall to the ground in a normal death pose.  This is hilarious.
- The situtation with respawning and vehicle ownership is weird.  Just keep in mind that, when you respawn, your new player character will no longer have control of their original vehicle.  Another player can claim it.  Just another symptom of a vehicle mechanic that needs due attention.
- There is currently no mechanism for checking that too much of everything is spawned.  Every dead player, cadaver and corpse alike, is like a whole other player until despawning.  Keep this in mind.
- I wrote the code myself and even I consider it a miracle any of this stuff works.  Things just trust that some other thing is doing what it should be doing correctly.  Depending on how the Test Live or Dev Test Live versions operate under duress, the system may have to be refined for more reassurance.

Unrelated:
- If you `/destroy` while looking at a vehicle, the server will attempt to deconstruct it.  The deconstruction will only work if: you own the vehicle, or if the vehicle is the same faction as yourself and either the owner is absent or the vehicle has zero health.


common/src/main/scala/net/psforever/objects/serverobject/tube/SpawnTube.scala
```diff
@@ -13,22 +13,24 @@ object SpawnTube {
     new SpawnTube(tubeDef)
   }
 
-//  import akka.actor.ActorContext
```
xTriad: Did you forget to remove this commented code, or are you keeping it in for a reason?

pslogin/src/main/scala/Zones.scala
```diff
@@ -19,6 +19,11 @@ object Zones {
 
       import net.psforever.types.PlanetSideEmpire
       Building(2).get.Faction = PlanetSideEmpire.VS
+      Building(2).get.ModelId = 20
+      Building(38).get.ModelId = 0
+      Building(42).get.ModelId = 0
```
xTriad: Two buildings with the same ModelId?

Fate-JH:
I can't see the comments on this PR when on GitHub so I'll just reply to them.

1) Those dash means "removed."  Actually, it was uncommented and that is the change.

2) Two buildings with the same invalid `ModelId`.  This way, the actual buildings with the model ids 38 and 42 won't risk being polluted when the buildings are configured during continent load, e.g., as with a `SetEmpireMessage`.  At the time, I did not have access to the numbers, so I choose a number that indicates neither something on the map nor anything with a model.  It should be fine since they are bunkers.

PR #197, opened 2018-04-20
Proximity Terminals/Medical

Fate-JH:
I spent more time on this than I intended.

While I should not have to explain how to test the medical terminals, I should remind you to try kill yourself on the terminal while it is working to make certain it deactivates.  Check that the terminal deactivates when your character moves during its operation too.

Due to being based on the `spawn-tube` pull request, your character is still on Ceryshen in the middle of nowhere.  Anguta's two medical terms are operational, however, and respawning always resets your health to 50 and armor to 25.  The four terminals in the `home3` HART C building also work if you respawn to that sanctuary continent.

Caveats:
- Although the code suggests that other proximity-based utilities are prepared, only the medical terminals should be expected to correctly operate.  The cavern crystals are peculiar.  If, somehow, normal 
repair silos respond, assume that this functionality is unintended behavior.
- Be cautious of constantly inching oneself forward on the terminal platform, waiting only long enough for the terminal to activate, before starting to inch again.  While most hiccups have been ironed out, this sort of treatment did cause the terminals to enter into an unstable state in an earlier tests.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
+++ b/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
@@ -515,6 +515,14 @@ object GlobalDefinitions {
 
   val respawn_tube_tower = new SpawnTubeDefinition(733)
 
+  val adv_med_terminal = new MedicalTerminalDefinition(38)
+
+  val crystals_health_a = new MedicalTerminalDefinition(225)
+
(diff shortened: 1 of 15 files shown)


Fate-JH:
I've already found a good reason to change how ProximityTerminal works.
Ah, well, it works good as it should for medical terminals.

PR #198, opened 2018-04-24
WIP: Account creation & login with Postgres database

xTriad:
The goal of this PR is to begin database integration so we can persist world state. Player accounts and player avatars are the first game aspects I intend to persist, but only accounts will be persisted in this PR.

- [ ] Write tests
- [ ] Handle invalid token from client in `WorldSessionActor`'s `ReceiveAccountData` case. Currently, the `AccountIntermediaryService` doesn't send a response if invalid token given from client, so `WorldSessionActor` becomes sequestered. Client would probably hang due to no response from `WSA`.
- [x] Populate the `logins` table fields when users login so we have basic analytical data. Remove the  `accounts.last_login` field since this can be determined from the `logins` table.
- [ ] Install Postgres database on public server?
- [ ] Setup database migration capabilities?
- [x] Rewrite asrynchronous database code to use actor patterns (`self ! CaseClass()`, `context.become(nextMethod)`) etc.
- [x] Bcrypt passwords
- [x] Generate custom token for each client
- [x] Add error handling for non-blocking “for expression” Futures that connect to the database and retrieve query results. There's not an obvious solution in Scala's documentation.
- [x] Handle `LoginSessionActor`'s failed `accountLogin()` `case Failure(e)`, send failed `LoginRespMessage` to client so it doesn't hang.

Here's a screenshot of an invalid password.

psforever-login-invalid-password

Updates:

Decided to switch to Postgres because it is better supported by the Scala library we are using. I was having issues with Mysql and don't want to waste anymore time on it. Postgres is a great open source relational database that is easy to use and worth using anyway.

Questions:

- Account creation happens when an account's username does not exist in the database. Is this how we intend new accounts to be created? **Answer: This if fine for now. See next question.**
- I saw in your code comments @Fate-JH  that you want to save the user's email. Can we add an additional field to the login screen? **Answer: Player registration will happen on the website.**
- Are we good with MySQL? I'm not worried about the Maria/MySQL debate and MySQL being owned by Oracle, MySQL works fine for me. If we really wanted to use a great database, I'd recommend PostgreSQL. **Answer: Using Postgres because it has better library support.**


pschord:
This has been superseded/updated by #317 
Thanks again for the initial work on this

xTriad:
Super happy to see this work be continued! Thanks for picking it up and moving it across the line in the other PR.

PR #218, opened 2018-06-05
Terminal hacking

Mazo:
Few things implemented in this PR:

- Locker hacking
- Terminal hacking (cert / equipment / medical / vehicle)
- Hacking speed based on certification level
- REK beam colour based on hacking level
- Hack sound centered on player instead of hacked object
- QoL change - show hacked object as belonging to faction that hacked it for the duration of the hack (only for that faction's players)

As an example for that last one in case of any confusion:

Imagine a vehicle terminal is owned by the VS. A NC infiltrator manages to sneak in and hack the terminal. Each empire will see the terminal as belonging to the following:

**VS:** Terminal owned by VS
**NC:** Terminal owned by NC
**TR:** Terminal owned by VS (not TR or NC!)

This allows for the infiltrator to stay undetected and not give away to the VS (or TR unless they use it) that the terminal has been hacked. All three factions will however still be able to use the terminal during it's hacked state.

When the hack expires, the terminal will return to being accessible to only VS, and will look like it's owned by the VS to all three factions.

This also applies to IFF locks, so no more running head first into locked doors that were in a hacked state seconds before (assuming you're paying attention to the lock colour)


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -3389,13 +3441,42 @@ class WorldSessionActor extends Actor with MDCContextAware {
...
+    * The process of hacking a locker
+    * Pass the message onto the locker and onto the local events system.
+    * @param target the `locker` being hacked
+    * @param unk na;
+    *            used by `HackingMessage` as `unk5`
+    * @see `HackMessage`
+    */
+  private def FinishHackingLocker(target : Locker, unk : Long)() : Unit = {
+    target.Actor ! CommonMessages.Hack(player)
+    localService ! LocalServiceMessage(continent.Id, LocalAction.TriggerSound(player.GUID, TriggeredSound.HackDoor, player.Position, 30, 0.49803925f))
```
Fate-JH: So hacking a locker (or a terminal) makes a door hacking sound? or is this just from being a copy of the door hacking function?

We should change the name of the `TriggeredSound` value if the former.  We could also probably combine these into a "simple hack" completion function by making the param `target : PlanetSideServerObject with Hackable`.

Mazo: Yeah currently it's the door sound. I should have put a note on that. The only reason I've left it like that is I honestly can't remember if that's the correct sound for everything being hacked or not.

If anyone can remember differently I can change it.

PR #222, opened 2018-06-23
Map updates

SouNourS:


Fate-JH:
Cripes, look at this history, and with only two files changed.  Can I squash this?

PR #223, opened 2018-06-25
Facility Turrets

Fate-JH:
+3 DEF when fortified.

Wall turrets are now mountable and armed.  The default phalax turret should fire properly when controlled.  Walk up to a turret along the wall, swap ammo to an upgrade canister in a nano-dispenser, and apply to select the upgrade.  If you grab another canister, you can change the upgrade when already upgraded.  When upgraded, you may mode switch to either the anti-ground vehicle weapon or the anti-air vehicle weapon, depending on upgrade was taken.  The previous fire mode will persist between disembarks and mounts, but it will be reset when the turret downgrades from AA/AV to just the standard phalanx.

Your test character should now be armed with a nano-dispenser and provided an upgrade canister in their inventory for testing.  The canisters are also appropriately sized in the inventory now and will only load one at a time in the nano-dispenser.  All of the wall turrets around Anguta and Akna tower will be wired for use (the closest one is up the stairs behind the vehicle terminal).  Anguta's tower is a watchtower-style field tower and has no turrets.  Turrets will always update in 20 seconds and will remain in their upgraded form for 30 minutes.  The change-over of turret weapons can be near-instantaneous on good days.

Caveats:
- Turret weapons, much like implant terminal interfaces, do not have their expected GUID's.  They will, however, maintain that same GUID throughout their lifetime.  The turrets will, however, register random GUID's for their internal ammunition magazines each time they upgrade.  See `Maps.scala` and the functions `Zone.TurretToWeapon` and `Zone.BuildSupportObjects`.
- When turrets change state, any seated players are kicked from them.  This behavior may not follow from what was observed on Gemini Live and, indeed, the animation is a little buggy.  It is being used as a defense mechanism against blow-ups that may occur during the upgrade/downgrade process if the weapon is shooting while changing.
- In regards to "blow ups," a very small possibility exists where turrets can cause a logging-in player to crash when coninciding right smack-dab during its upgrade.  The window for this failure is extremely small so it will naturally plague me constantly.
- Only turrets with no players seated in them may start the upgrade process.  On Gemini Live, some kind of guard existed to block the upgrade prompt; we have to block it at a later point that may leave the player discharging their gluegun at the turret without any affect.  There will be no upgrade and this state is easily cancellable; it will be fixed later, all the same.
- Once the upgrade process starts, another player may mount the turret without interrupting the upgrade already in action.  The player in the turret will be kicked out of it when it completes to allow the turret weapon to change safely.  Neither is correct behavior.
- The turrets do not auto-fire upon enemy MAX's and enemy vehicles at the moment.  No obvious packets are generated by the client so more work will be needed to figure this out.
- The turrets can not be hacked or infected at the moment.
- The graphic model of a held nano-dispenser, when in first person, may switch from looking like an upgrade canister to a normal repair canister if the view is changed or if the tool stops being rendered.  An upgrade canister will remain loaded.  This is a legacy visual bug with the client and not much can be done about it, not right now.

Additional:
- Due to the continuing failure of `receiveN` in my Travis CI runs of tests involving Actors, I have introduced an attempted workaround that performs the same effect as `receiveN` but uses the otherwise reliable `receiveOne`.  If it continues to fail, all of the tests utilizing that function will need to be reworked from the ground-up.
- Due to the `services` package being moved in a prior PR, the `service` tests have also been moved accordingly, involving lots of busywork import changes.


Mazo:
Do turrets revert their upgrades when a base changes faction (the wiki states they should)? I had a look through but can't spot where they do if so.

http://wiki.psforever.net/wiki/Phalanx

Mazo:
Since it doesn't appear MannedTurrets use the Vehicle class we can probably remove the comment in WSA's DismountVehicleOnLogOut() about base turrets (line 183)

Fate-JH:
To forcibly downgrade turrets upon base capture, use a `HurrySpecific` message during the code that changes the base's hands, with the target being all of the turrets for that base.  (`collect` all of the Amenities for that base that match as `MannedTurret` objects.)

common/src/test/scala/objects/ResourceSiloTest.scala
```diff
@@ -0,0 +1,332 @@
...
+    "construct" in {
+      val obj = ResourceSilo()
+      obj.Definition mustEqual GlobalDefinitions.resource_silo
+      obj.MaximumCharge mustEqual 1000
+      obj.ChargeLevel mustEqual 0
+      obj.LowNtuWarningOn mustEqual 0
+      obj.CapacitorDisplay mustEqual 0
+      //
+      obj.ChargeLevel = 50
+      obj.LowNtuWarningOn = 25
```
Mazo: I have no idea why I made this an Int (I'm going to claim temporary loss of sanity). It's supposed to be a Boolean for if the warning is on or off.

I can't push to this PR so after it's merged I'll fix it and submit another PR.

PR #225, opened 2018-07-16
Damages

Fate-JH:
Me: "Just 500 lines of code that handles dealing damage?  All of it in WorldSessionActor?  Should be easy to port over.  Let's see if we can tighten up the workflow."
_Two weeks later_.
Showing 47 changed files with 2,143 additions and 238 deletions.
Me: "I think something went wrong somewhere."

Shooting guns at opponents in the official repo now causes damage to the target, eventually leading to death.  The process, originally just about 500 lines of code in `WSA`, has been reduced to less than 100 in `WSA` and a slightly verbose but easily testable (you can write tests for it!) configurable damage system.  That's my story.  All damage and healing must be synchronized; it must occur in a specific order, and no target may be healed by multiple sources simultaneously.  For all landed shots, all the preparation work is calculated on the shooter's side of the server code.  For infantry targets, the remainder work is then dispatched over to the his side of the code for application and distribution of results.  For vehicle targets, the remainder work is divded into two halves - the part that the vehicle handles, and the part that is returned to the shooter for distibution of results.  Other non-Infantry targets will probably be utilizing the latter plan.

To make any object capable of taking damage, one must include the `Vitality` trait and then set up the damage model.  Actual operation of the damage model workflow breaks down into the system as defined in the `Vitality` section below.

`ResolvedProjectile`:
As much an active means of pushing around information about weapon discharge that hits a valid target, the goal is that all `ResolvedProjectiles` can be used to perfectly replicate the history of the vital statistics of a player.  This will not only eventually factor into the player's validation (it already is used to determine how to accredit kill-shot) but it will be used for the eventual Merit Commendation system.

`Vitality`, `ResistanceProfile` and `DamageProfile`:
A vital object is one that can be damaged or repaired, healed or hurt, but the class itself does not care how that is actually executed.  It maintains a reference to the damage model that is then used to reconstruct damage values from the projectile information and resistance values from itself (the target).  In general, the model will exist on the same portion of the object where the resistance values are stored, though this is not required by code anywhere.  An example includes infantry and their exo-suit definition.  The model also contains information how to reconcile the damage and resistance and a means of emitting a function into which the target can later be fed to apply that damage to the target object.  The values used for damage are stored on the projectile, with known modifiers on the weapon's fire mode.   Other use of the `trait` is to hold the history of damage, repair, healing, hurt actions) and probe it for specific entries.  The example of this currently in use is determining who gets the kill shot on a target so to construct a `DestroyDisplayMessage` packet.
No actual connection between the eventual target to which the values are applied and the target called-out in the `ResolvedProjectile` object need exist.  These effects can be repeated on multiple objects as long as the input object is acceptable to the emitted function.

`Vehicle destruction`:
Vehicles can be destroyed from weapons' fire, much like on server Play-Live (51200).  Vehicles also have a lightweight variant called "wreckage" or "`DestroyedVehicle`" that creates the impression of said vehicle much in a similar way that dead infantry are turned into corpses models.  Previously, the same converter was used for both functionla and wrecked vehicles and that was resulting in the vehicle's former weapons being deposited underneath its wreckage.

`Passenger death`:
Players are responsible for dealing with their own death.  To maintain a consistent order of events, a client's `WSA` is informed when its player should die directly.  See the next section "Named event channel."

`Named event channel`:
All players will now have a named event channel - currently just on `AvatarService` - that is used to send messages from one player to another specifically.  Currently, this is only possible by sending every player a message and having each player check whether it is their message to receive,  This method had already been utilized on the `VehicleService` when ti comes to vehicle occupants receiving messages about ammunition changes and other updates.  This will reduce the burden in that respect as the overhead is low.  At the moment, `Service` message throughput is stable enough to support this expansion of responsibilities.
If only a means of sending packets to everyone except for a certain player were obvious, a lot of `if` statements could be eliminated.

`DestroyDisplayMessage` and `HitHint`:
Responsibility over who is performing damage should work.

Vehicle shields:
Vehicle shields will charge if the base receives lattice benefits from an Amp Station.  (At the moment, no base gives benefits.)  Vehicle shields are factored into standard damage calculations, always recharge at a fixed rate no matter which vehicle is affected, and have two different delays hardcoded: don't recharge if less than 1s has passed wince the last recharge tick; and, don't recharge if less than 5s has passed since the last damage.

Caveats:
- Only infantry and vehicles can damage each other.  Turrets and terminals will not be affected for the moment.
- (Most) Vehicles generally do not have resistances, despite the code that is used to calculate their damage values accounting for the possibility.
- The damage calculations will not necessarily match those that were used on 51200.  I played with the code a little to try eliminate the potential to encounter of negative damage, or "healing bullets" as it were.  While successful in that endeavor, some of the calculations were adjusted in other ways and the risk of negative damage still exists in some cases.  (Check `ResolutionCalculations.scala` for the said code.)
- One should note that damage is actaully calculated twice.  The second time is a part of the code where the values are applied to the target object; this second time is performed to attempt to preserve contribution of the original values should the target object have changed from when the `ResolvedProjectile` was produced.
- Not all vehicles have a simplified destroyed representation, known as `{vehicle}_destroyed` in the list of game objects.  The Phantasm and the BFRs do not have their own wreckage models, for example, and their wreckage must be generated using a modification of the standard vehicle converter.
- Cause of death will always be recorded as the last person who dealt damage to the victim with a projectile, regardless of how long ago that damage was performed.  `/suicide` may, therefore, result in kill-credit unless the player never took damage.  Better pruning of the history in the future will eliminate this.
- There are really no details on how shields work in the adb list and we most certainly will not have enough information for all of the vehicles in the packet captures, if they differ in any sort of way.  Functionality information here is guesswork.

Addenda:
- What kind of targets are affected by `damage4` values?
- Without the accommodations provided, vehicle shields would charge at a rate equal to the number of occupants in the vehicle.  Each seated player sends a packet to the server about the vehicle not having full shields energy and, without pacing, they would all contribute to a single "tick."


Fate-JH:
No changes to tests at the moment.  I'll append commits with tests later.

Fate-JH:
When I say multiple sources, I'm talking about race cases and working with damages on the same "pipeline."  Let's say that two players deal damage to the same infantry target at approximately the same time (you know, standard firefight stuff).

First scenario: target heals at a crucial time of death.  Some damage and the healing can evaluate in any order, such as one that could cause the target to die and then revive again because of delayed events.  Imagine what it would be like fighting an enemy in proximity of a health crystal or on a medical terminal - the hitpoints never match up properly so the target wildly jumps between low values and high values as both sources contend to detemine how healthy he should be.

Second scenario: imagine if the first of the two aggressors kills the target while the second is actually in the middle of working on the target's damage calculations because the target was a valid live one when it started.  Because the damage calculations are occurring out of order, the latter player would be applying damage to an "already dead" target and being attributed for the kill by arriving later.

Third scenario: the damage is simultaneous enough that the server grabs the same health/armor values from the target and calculates the same damage numbers both times and applies them independently.  That reduces getting shot twice into taking damage from getting shot once.

Mazo:
That makes more sense and seems sensible. As you were 😄

SouNourS:
Results from a short test : 
When you kill someone with a vehicle, the kill spam have a "skull & bone" icon.


Seems good ;)

PR #230, opened 2018-09-01
Deployables

Fate-JH:
Deployables are feature heavy.

`ConstructionItem`:
The class base for the `ace` and `advanced_ace` has been given a significant content revamp.  Instead of a single-dimension list of construction options, as strings, it has obtained fire modes to organize deployable settings.  It does not count as a `Tool`, however, because that would require it having "ammunition" to test against.  Rather than shuffle the list with new options, each deployable setting contains the information of its certification requirements and tests upon switch, only settling on whatever is the next possible valid setting.

The advanced construction engine (ACE) and field deployment unit (FDU) are now available for being pulled from any equipment terminal.  For the sake of convenience, the test character is decked out in construction items and one Suppressor.  The following certifications are important for testing: `CombatEngineering`, `AssaultEngineering`, `FortificationEngineering`, `AdvancedEngineering`, and `AdvancedHacking`.  `GroundSupport` is also important; but, its effects can not be tested at this time.  All combat engineering items - deployables - can be constructed.  Certifications, as is appropriate, change the available construction options on the ACE and the FDU.  Under normal circumstances, all deployables are organized on the equipment in order of certification access.

`DeployableToolbox`:
The `Avatar`-specific data structure that keeps track of owned deployables.  It controls the quantities and limits both for categories and for types.  It also helps coordinate UI element update information.  Commonly, deployables are added and removed from the internal lists, denoting "ownership" when added and on the same continent.  (Deployables can still be considered as "owned" while the engineer is off-continent, though they are no longer tracked here.)

The UI elements can be made visible, if not, by pressing `alt+e` a few times.  Mousing over the icons will reveal what deployable belongs to which icon.  The first number is the current number of owned deployables.  The second number is the maximum number of owned deployables.  For small windows of opportunity, it is possible that the number of owned deployables will exceed the maximum.  Going up to any certification terminal, a player should be able to update the engineering and related certifications and the UI elements will also update.  As a reminder:
- Having no available deployables will show a red 0/0.
- Having no owned deployables will show a red 0/y.
- Having a non-zero non-maximum number of deployables will show a green x/y.
- Having more that maximum deployables will be a yellow x/y.
- Having a maximum number of available deployables will show a red y/y.

Each deployable is grouped by **type**.  Some deployables are also grouped into **categories** by type.  Deployables have a maximum number both by type and by category.  For example, spitfire turrets, shadow turrets, and cerebus turrets share a category.  Although fifteen spitfire turrets can be placed with an advanced certification level, only five cerebus turrets or five shadow turrets can be placed at once.  If five cerebus turrets and five shadow turrets have been built, only five normal spitfires can be placed, for the total of fifteen.  Without cerebus turret or shadow turret, all fifteen spitfire turrets can be utilized at once.  Information about category limits is not explicitly reported in the UI, but frequent messages in event chat and occasional on-screen splash text will be encountered discussing these limits.

`DeployableRemover`:
A `RemoverActor` for deployables, used for instances where the deployable no longer has an owner on the continent (or in the game).  Without an owner, all deployables can subsist for three minutes (see `Addenda`).  In certain instances, deployables can be rushed to be removed from the game world.

`Deployables`:
Deployables have various elements that make them "work."  Deployables have animations that accompany their appearance and their disappearance, delays that accompany those same events, icons on the continental map visible to allied players, and they define very specific owners.  Officially, "owners" are defined by their global identifier and are usually the same person that placed them (we have support for owners by name; see `Addenda`).  They can either be destroyed by weapon discharge or from the owner manipulating the continental map icon.  Some deployables should have special functionality but, by majority, most are just decoration for now.  A lot of server-side functionality exists that is not supported in this update.

`BoomerDeployable` and `BoomerTrigger`:
Though, for the purpose of reducing complexity, boomers and their triggers are permitted a degree of separation, this separation actually introduces other forms of complexity.  For example, deployables have their own `RemoverActor` but that only handles the boomer itself.  Even though the trigger - just a piece of `Equipment` - could just as easily be handled by the removal `Actor` for items on the ground, the trigger is also explicitly handled for a minimal of fuss.  Boomers and their triggers go hand-in-hand; destroying the boomer destroys the trigger; destroying the trigger destroys the boomer.  This relationship occurs not only during manual deconstruction, not only during manual destruction, but also during continental transfers and logouts where owned boomers are removed from the game world via their held triggers.  As the trigger is shuffled around, so too shifts the allegiance of the boomer, as is indicated by the green circle on the ground under the explosive.  Boomers can be made to explode by their trigger but will perform no damage and will be cleaned up afterwards.  Boomer damage needs to be calcaluted on the server, so these units are just decoration at the moment.

`ExplosiveDeployable`:
Including both `he_mine`s and `jammer_mine`s, aka "Disruptor Mines," these deployables appear on the ground.  They do not explode.  Mine detection and mine damage both need to be calculated on the server, so these units are just decoration at the moment.

`SensorDeployable`:
Including both `motionalarmsensor`s and `sensor_shield`s, aka "Sensor Disruptors," these deployables can be placed upon most surfaces in many angles.  Sensors have a graphical element that appears when they are used.  Sensor operation needs to be calculated on the server, so these units are just decoration at the moment.

`TurretDeployable`:
Turrets comprise two different types of deployables, constructed by the two different construction items.  The "spitfires" - normal as `spitfire_turret`, shadow as `spitfire_cloaked`, and cerebus as `spitfire_aa` - can be produced from the `ace` by players of all factions.  The `portable_field_turret`s - TR Avenger, NC Osprey, and VS Orion - can be produced from the `advanced_ace` by players of the appropriate faction.  A nondescript field turret does exist but is relatively pointless currently.  The spitfires need to have their functionality calculated on the server and the units are just decoration at the moment.  The field turrets can be properly mounted and used as weapons, however, similar to facility turrets.  When exposed to a friendly SOI, the field turrets should have charged shields, much like vehicles; but, this is not currently supported.

`TrapDeployable`:
`tank_traps`, aka TRAP's, can be placed into the world and will obstruct enemies and permit allies.  As collision damage is not yet respected on Dev-Test, no damage is caused by interaction with TRAP's.

`ShieldDeployable`:
The Aegis Shield Generator may feature a variety of hacks and updates but none of them are supported in this update.  The `deployable_shield_generator` will, however, still negate visibility for anyone that steps under it.  Semi-functional, maybe?

`Zone.Where` and `ItemLocation`:
Assistance for absolutely finding where something is located in a zone.

Caveats:
- Information about deployable categories can not be determined through testing under current conditions.  The wiki has only a few parts of entries that explain it.  In the disruptor mine (`jammer_mine`) entry, the following text is encountered: "The Disruptor Mines count for the maximum number of mines that you can deploy, along with the High Explosive Mine."  In the shadow turret (`spitfire_cloaked`) entry, the following text is encountered: "The maximum number of Shadow Turrets that can be deployed is 5, and they count toward the total number of Spitfires you can deploy."  In the cerebus turret (`spitfire_aa`) entry, the following text is encountered: "The maximum amount of Cerberus Turrets that can be placed at any given time is 5, and they count towards the maximum amount of deployed Turrets along with the Spitfire and Shadow Turrets."  The sensor disruptor (`sensort_shield`) entry does not contain text that categorically links it to the motion sensor (`motionalarmsensor`); but, the obvious pattern of all deployables belonging to the same firemode also belonging to the same category is quite convincing.
- Some deployables do not have messages regarding having placed the maximum number of a deployable, or regarding overwriting the most long-standing of its type or category.  Once again, this is a little discussed and untestable situation.  For the purposes of the effect, however, some quick and dirty rules were put in place (without adequate testing).  For example, any deployable for which only one unit can be placed will not generate a "maximum number reached" message.  Additionally, "oldest deployable deconstructed" determination could happen at either the **category** level or the **type** level, depending on what is necessary.
- At the moment, deployable map icons are only visible to the faction to whom the deployable owner belongs.  This is typically unremarkable save for the instance of the boomer, whose ownership and allegiance is controlled by the player that possesses the trigger.  I am uncertain if some mechanism is exercised on the clients of non-allied players in regards to deployables being displayed.
- Deployable ownership is slightly enhanced.  When a player transfers continents after having built a deployable, that player loses ownership of it in as far as the available "slots" for building deployables matter.  Upon returning to the continent, the player normally does not regain possession of the deployables that they previously placed.  On such an occasion where the player, returning to the continent, is assigned the exact same globally unique identifier, however, the system will be tricked and those deployables that still exist will be partially given back to their control.  The UI elements will count incorrectly but the map icons will be marked for ownership.  This happened to me once.  Expanding upon this happy coincidence, a player is now permitted to retain ownership of all deployables that they previously placed if they should come and go from a continent.
- There is cleanup code should deployable construction system break down.  This should never happen but must still be accounted.  The construction will be canceled and the construction item can be reclaimed, either by not leaving one's hand in the first place, or by being placed on the ground as if dropped (for the FDU, this is aesthetically pleasing).
- Neither hacking nor upgrades are supported at the moment.  I take no responsibility for any visible interaction with the REK.
- When a cadaver possesses boomer triggers then becomes a corpse, those boomer triggers are dropped onto the ground.  This is not standard behavior but will be tolerated for consistency for now.
- All deployables will exist for only three minutes after becoming un-owned.  Normally, un-owned deployables persist for about twenty minutes unless destroyed.  This will be resolved in a future update but will be kept at the current duration for testing efficiency.
- For the sake of keeping boomers and triggers from becoming separate, players can not stow triggers into vehicle trunks or their own lockers.  This is not standard behavior but will be tolerated for consistency for now.
- According to the packet captures, deployables have specific animations for appearances and disappearances, represented by the other field in `ObjectDeleteMessage`.  Visual confirmation from testing on Gemini-Live attest to this fact.  Also according to the packets, however, the deployables are also made invisible within a short window before being formally destroyed and that masks the animation.  I can not make heads or tails of this behavior but, for what it is worth, deployable construction animations, especially those of the FDU, also behave incorrectly.
- No information on the amount of "health" for explosive deployables - mines and boomers - can be found in the ADB file.  Nonetheless, they can be shot to death so they have temporarily been granted 100 health for the purposes of calculations.  All deployables are presumed to take vehicle-type damage for both direct hits and splash, and do not suffer from lashing effects.
- The nameplate over a destroyed boomer is temporarily replaced by "boomerd" before being cleaned up.  This is pre-existing behavior.
- The nameplate over a destroyed sensor is temporarily replaced by "destroyed *" before being cleaned up.  This is pre-existing behavior.
- The total number of `deployable_router_telepads` available is set to 1024 which, in my opinion, is reasonably unrealistic.  Players can not acquire `router_telepad`s right now so this point is moot.

Addenda:
- `MannedTurret` class has become `FacilityTurret` class, with abstraction `WeaponTurret`.
- Fixed a `CSRWarp` issue involving "irkall[a]."
- Modified sanity checks for zone startup.  Log spam remains.
- I really don't like the `RemoteUnit` class.  Consider it `@deprecated` on release.
- Sampling a resolved projectile returns the damage value(s).  No need to insert dummy objects to extract field changes.
* Capture consoles have been added to various facilities and towers


Fate-JH:
**`/home/travis/build/psforever/PSF-LoginServer/pslogin/src/main/scala/WorldSessionActor.scala:251: Could not write class WorldSessionActor$$anonfun$Started$1 because it exceeds JVM code size limits. Method applyOrElse's code too large!`**

WorldSessionActor.scala: 7172 lines total; `Started`: _1923 lines_

Well ...

Fate-JH:
`handleGamePkt` is getting up there at 1875 too ...

PR #232, opened 2018-09-27
Router and Telepad

Fate-JH:
"Ever since I saw that person during the last event who pulled a Router and tried to bring it around to an enemy base, I felt bad for him.  Let's make Routers work, even if just a little bit more."

`Utility.TeleportPadTerminal`:
An `Amenity`-type object that dispenses only router telepads.  It located at the nondescript back of the vehicle and is functional when the Router has deployed.  It has no interface, only producing mesages when close enough to it, and interacting with the unit is enough to place a Router into one's inventory in any available space.

`TelepadLike`:
A common `trait` of units that demonstrate Telepad-like properties.  Actual `Telepad` objects, `TelepadDeployable` objects, and the `InternalTelepad` use this class to bus around a reference to the associated Router vehicle and whether or not the object is ready to serve as the endpoint in a teleportation system.  (The latter feature is inconsequential for telepad objects.)

`Telepad`:
A special `ConstructionItem` object designed to produce a single deployable - the `router_telepad_deployable`.  The only server-imposed limitation on Telepads is that they can not be restored through loadouts.  If the Router associated with the telepad is destroyed or has moved to a different continent, the telepad will refuse to deploy (client-enforced) and a message will be displayed.

`TelepadDeployable`:
A floaty `Deployable` item that will eventually split and produce a blue spheroid used for teleportation when paired with a functional deployed Router vehicle after 60s of being placed.  The 60s activation time can not be skipped.  Walking into a telepad deployable when it is active will instantly move the player to the position of the router, accompanied by a unique teleportation sound.  If the telepad deploys but the Router has shifted to a state of not being ready in the meantime, the telepad will patiently wait to activate and a message will be displayed.  If another telepad is deployed with an association to the same Router, regardless of the condition of the Router, or that old telepad, the old telepad will be deconstructed.  When the Router is destroyed or undeploys, the currently-associated telepad deployable will also be deconstructed.

`Utility.InternalTelepad`:
An `Amenity`-type object that serves as the Router-side endpoint of the teleportation system.  Clients are instructed to normally construct this object every time the Router deploys (or, at least, the first time it deploys ...).  The "beam" underneath the vehicle itself is always a visible component of the deployed Router.  It does not become active, however, until paired with this unit and until the whole of the teleportation system is ready.

`RouterTelepadActivation`:
A `SupportActor` that counts the 60s until `TelepadDeployable` is ready to connect to an active teleportation system.

Caveats:
- The `InternalTelepad` object must exist as a permanent aspect of the vehicle's object create packet due to the automatic way the server translates objects into network data.  Despite that, the code for independent creation (and deletion) of the `InternalTelepad` object occurs each time the system attempts to change its activation state.
- Routers require an indeterminate amount of time from deployment to be ready to attempt to active a teleport system.  This period is typically how long it takes the vehicle to switch between deployment states, with some extra milliseconds.  It takes longer to completely disconnect the client from the server (see `SessionReaper`) so it is permissible to perform this action in the local user space.
- The server attempts to impose a 2s cooldown between teleportations to make certain that the user has time to evacuate the beam before it triggers and attempts to return the user to the original side.  In theory, the user should be able to stand in the beam just after appearing and not be returned back to the other endpoint.  In practice, either the client does not send evenly-spaced packets, it has an internal cooldown of some sort, or the `ActorSystem` incurs some time penalty before processing an otherwise timely packet.  The user is still fully capable of not ping-ponging back and forth between endpoints most of the time despite infrequent slips.
- The number of `TelepadDeployable` objects that may be owned at one time is permanently set to 1024, regardless of the certifications belonging to a character.  This is because, unlike other equipment and vehicles that can be pulled from a terminal, even a person without the `GroundSupport` certification can obtain a router telepad from its vehicle-borne terminal and successfully deploy it.

Addenda:
- To allow other players to obtain a telepad from the terminal, the owner of the Router must permit access to the trunk.  This absolutely normal behavior but I feel that it needs to spelled out clearly.
- To deploy more than 1 of the available 1024 telepads permissible requires using multiple Routers.  Regardless of the number of telepads in the field owned by the player, the chat message for the deployable will always act as if whatever unit has just been created, reported as "1 out of 1."
- Terminal `UseItemMessage` handling code in `WorldSessionActor` has been changed a little.  Please be a second (third, fourth, fifth, and so on) pair of eyes and check all applicable interactions, e.g., hacking.
- Zones now accept `Option[PlanetSideGUID]` as searchable information.  This important not just to eliminate frequent `getOrElse` clauses and `match` cases but it also is much more fluid to read.
- An issue with Maelstrom and Radiator grenade projectiles where hit information exists but an actual hit target isn't defined, thus causing the client to disconnect, is resolved.
- Despite being adjusted slightly, `RemoverActorTest` classes are still commented out because they won't work (in continuous integration testing).

### DIFF
--- a/common/src/main/scala/net/psforever/objects/Deployables.scala
+++ b/common/src/main/scala/net/psforever/objects/Deployables.scala
@@ -22,7 +22,8 @@ object Deployables {
       DeployedItem.portable_manned_turret_nc -> { ()=> new TurretDeployable(GlobalDefinitions.portable_manned_turret_nc) },
       DeployedItem.portable_manned_turret_tr -> { ()=> new TurretDeployable(GlobalDefinitions.portable_manned_turret_tr) },
       DeployedItem.portable_manned_turret_vs -> { ()=> new TurretDeployable(GlobalDefinitions.portable_manned_turret_vs) },
-      DeployedItem.deployable_shield_generator -> { ()=> new ShieldGeneratorDeployable(GlobalDefinitions.deployable_shield_generator) }
+      DeployedItem.deployable_shield_generator -> { ()=> new ShieldGeneratorDeployable(GlobalDefinitions.deployable_shield_generator) },
+      DeployedItem.router_telepad_deployable -> { () => new TelepadDeployable(GlobalDefinitions.router_telepad_deployable) }
     ).withDefaultValue( { ()=> new ExplosiveDeployable(GlobalDefinitions.boomer) } )
   }
 }
--- a/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
+++ b/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
@@ -861,6 +861,8 @@ object GlobalDefinitions {
   val portable_manned_turret_vs = TurretDeployableDefinition(DeployedItem.portable_manned_turret_vs)
 
   val deployable_shield_generator = new ShieldGeneratorDefinition
+
+  val router_telepad_deployable = DeployableDefinition(DeployedItem.router_telepad_deployable)
   init_deployables()
 
   /*
@@ -5527,5 +5529,11 @@ object GlobalDefinitions {
     deployable_shield_generator.MaxHealth = 1700
     deployable_shield_generator.DeployTime = Duration.create(6000, "ms")
     deployable_shield_generator.Model = StandardResolutions.ComplexDeployables
+
+    router_telepad_deployable.Name = "router_telepad_deployable"
+    router_telepad_deployable.MaxHealth = 100
+    router_telepad_deployable.DeployTime = Duration.create(1, "ms")
+    router_telepad_deployable.Packet = new TelepadDeployableConverter
+    router_telepad_deployable.Model = StandardResolutions.SimpleDeployables
   }
 }
--- a/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
+++ b/common/src/main/scala/net/psforever/objects/GlobalDefinitions.scala
@@ -863,6 +863,8 @@ object GlobalDefinitions {
   val deployable_shield_generator = new ShieldGeneratorDefinition
 
   val router_telepad_deployable = DeployableDefinition(DeployedItem.router_telepad_deployable)
+
+  val internal_router_telepad_deployable = DeployableDefinition(DeployedItem.router_telepad_deployable)
   init_deployables()
 
   /*
@@ -5135,6 +5137,7 @@ object GlobalDefinitions {
     router.Seats += 0 -> new SeatDefinition()
     router.MountPoints += 1 -> 0
     router.Utilities += 1 -> UtilityType.teleportpad_terminal
+    router.Utilities += 2 -> UtilityType.internal_router_telepad_deployable
     router.TrunkSize = InventoryTile.Tile1511
     router.TrunkOffset = 30
     router.Deployment = true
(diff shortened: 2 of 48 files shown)


Mazo:
Have done some QA testing:

Bugs:
----------------------
Players joining zone don't see correct deployment animation for router (no beam effect but disc still spins) or telepad (no deployment animation). This also prevents them from teleporting if there is an active telepad.

In a similar vein to above, a player joining the zone that then deploys a telepad from an existing deployed router still can't see the correct animation on the router side (telepad is fine). They can teleport from either end still.

Standing under a teleport destination will ping pong you back and forth but only if looking directly upwards (!?)

Deployed router disappeared after ~10-20 minutes even with driver still logged in and in zone. Driver & trunk permissions were set to empire - no other driver had entered the vehicle. I guess this is the hardcoded vehicle decon timer that also applies to AMS. It probably shouldn't do this if someone has used an AMS/Router recently

Deploy a router, take a telepad, place it in a locker then retrieve it and deploy it. Then deconstruct the router and clients will get disconnected (happens with both telepad deploying and deployed)

Deconstructing a router with a telepad stored in a locker causes client disconnects (It may be a good idea to just disallow storing telepads in lockers completely. I can see it causing problems later on when lockers are persistent)

Deconstructing an owned router while a telepad is deploying doesn't clean up the telepad until after the telepad deployment phase ends (this isn't too bad, but could benefit from being removed immediately)

Putting a telepad in the router's trunk and deconstructing it will disconnect all clients EXCEPT the router owner

Doing the above but putting the telepad in the trunk from a second character will disconnect both clients


Questions / Thoughts:
----------------------
Is router <-> telepad distance enforced server side? The client will not allow deployment too far away

Could we possibly utilise the first bug (deployed with no beam effect) as a QoL change to show the difference between deployed routers that have an active telepad and without?

It might be worth considering what happens to a deployed router & telepad when vehicle hacking is added. What happens in that case?

Vehicle trunks currently have no range limitation so can be kept open indefinitely. Whatever solution we do for this needs to ensure that you can't put a vehicle trunk next to a telepad and have the trunk open while teleporting to bypass the distance restriction

Fate-JH:
The animation state of the Router is curious as that is caused by transitioning into a fully deployed (`DriveState.Deployed`) condition from an actively deploying (`DriveState.Deploying`) condition, and not caused any specific attribute exchange.  Dispatching supplemental deployment packets during `BeginZoningMessage` does help.  We need to be fully deployed to access the terminal.  Given the bug, I believe this does mean that we can switch directly to being deployed and bypass the teleportation beam effect AND have access to the terminal; BUT, I need time to rewire the deployment system to allow for custom deployment state transitions.  Since this would be QoL and not vanilla, I would provide Sounours with a PR that is just for Play-Test integration.

The enforced distance is only clientside right now.

As for vehicle hacking, I presume little happens.  The teleportation system is not IFF so that is not affected.  The active telepad may have to lose its owner, maybe.

I fixed the internal deployment states so the deployed telepad animation and, moreover, the teleportation system itself, should work on continent load.

For the ping-ponging effect, see caveat #3.  The fact that the client continues to send packets that would cause teleporting back to the other side suggests the possibility is not discounted and that's why I only have a cooldown.  As for why the effect continues to work when standing or looking or moving some directions and not all the time, I have no clue.  Perhaps, messages sent between internals of the server interfering with the read time of packets sent by the client?

I can not reproduce the locker crash interactions that you reported.  Normally, the router does not know about the telepad trying to connect to it until after that telepad is ready to be deployed so deconstructing a router should not deconstruct undeployed telepads obtained from it.  Not even those stored in someone's locker.  The deployed/ing telepad only cares about the the Router when it is first constructed and again when it actually connects.  Telepads do not overwrite each other until the teleportation system connects to a new one, to ensure uptime, and I did not want the router to have to keep track of an undefined number of telepads waiting to connect.

I resolved the deconstruction crash issue with a more wholesome unloading message.  In fact, this even resolved the locker issues.

Owned vehicles no longer deconstruct if the driver is running about freely anymore.

I'll test these fixes a bit more later today, then push the commit.

Fate-JH:
I left two pairs of `VehicleServiceMessage.Decon` calls commented out.  This is where the vehicles - the AMS and the Router - should reset their deconstruction timers if they are unowned.  Checking whether or not they are actually unowned at the time would need to be added.

Mazo:
Can confirm deconstruction crashes seem to be gone and deployment state is synced properly to clients joining the zone.

Nice work 👍

PR #233, opened 2018-10-10
Re-time Door Close

Fate-JH:
When doors are opened, but will not reclose due to the precense of the player that opened the door, that door will have its entry retimed before being put back into the queue of doors waiting to be closed.


Mazo:
👍

nwuollett:
Test comment.

PR #235, opened 2018-11-08
Damage Fix, and Stuff

Fate-JH:
Not really a compelling title, is it?

The primary features of this update are the corrections to the damage calculation functions, with special attention given the splash radial damage and lash damage.  A new modifier has been added (`Subtract`) and previously mistaken resistance numbers are now factored into the solution.  Damage tick markers should be consistent now between instances of damage-taking and types of damage (as they are manually controlled now, as opposed to being at the whim of some client as with their previous lazy implementation) and all projectiles are now visually communicated between clients.  Rocket launchers should all be forced to their pure dumbfire modes since other modes are complicated.  In general, all standard Infantry arms - Standard Assault, Medium Assault, Heavy Assault, Special Assault, and AntiVehicular - that are not Cavern weaponry should possess acceptable functionality.  These changes also influence vehicle weapons, against which Cavern vehicles have similarly not guaranteed.

Inventory-based fixes include an exosuit swap issue that destroyed the inventory of a MAX player, an inventory swap issue that caused slippery fingers and broken equipment, changes to how armor is replenished when swapping exo-suits, and changes to how much ammunition is restored when using a Favorite loadout.

Miscellaneous fixes include: vehicle trunks now having a range and velocity limitation in regards to how long they can be held open, and this quality is extended to all types of containers; an issue with swapping drawn equipment leading to server-driven phantom limbs; and, the long-standing issue with `ResourceSilo` objects accessing a GUID that they do not yet possess.

`BindPlayerMessage` has been changed thus that the `Vector3` field indicates the correct locations on the map.  This is important for later.

Caveats:
- Of any notable and documented differences in damage calculations, the Lasher using AP fire mode destroys MAX's in two less shots than reported on the wiki.  This is considered "acceptable" for now.
- Due to the way rocket launchers are mode locked, they always display the wrong message in events chat.  They will not stray from the expected fire modes functionally, fortunately, meaning that this is merely a cosmetic issue.
- Range limitation on vehicle trunks force the Lodestar trunk to only be accessible from exactly right next to it.  This is not really a problem so much as it is a point of finicky behavior.  BFR's are untested, of course, but, as they are generally not much bigger than Lodestars, should pose no problem.
- Despite having fixed MAX inventories when it comes to switching between subtypes of the exo-suit, MAX's are still essentially broken.

### DIFF
--- a/pslogin/src/main/scala/WorldSessionActor.scala
+++ b/pslogin/src/main/scala/WorldSessionActor.scala
@@ -879,6 +879,9 @@ class WorldSessionActor extends Actor with MDCContextAware {
         AnnounceDestroyDeployable(target, None)
       }
 
+    case Vitality.DamageResolution(target : FacilityTurret) =>
+      HandleFacilityTurretDamageResolution(target)
+
     case Vitality.DamageResolution(target : PlanetSideGameObject) =>
       log.warn(s"Vital target ${target.Definition.Name} damage resolution not supported using this method")
 
@@ -1017,7 +1020,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
               if(obj.Health == 0) Some(obj)
               else None
             case Some(obj : FacilityTurret) =>
-              if(obj.Health == 0) Some(obj)
+              if(obj.Health == 1) Some(obj) //TODO proper turret death at 0 health
               else None
             case _ =>
               None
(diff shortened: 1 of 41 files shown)


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -2880,7 +2880,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
               accessedContainer = None
             }
           case Some(container) => //just in case
-            if(vel.isDefined || Vector3.DistanceSquared(player.Position, container.Position) > 25) {
```
Mazo: Are lockers the only thing affected by this? With PR #236 we can set the position of lockers (which we have now) that would render this point moot, and would be a good idea to enforce server side anyway.

Fate-JH: If that's the case, the distance check should be performed in the `UseItemMessage` block for Lockers and not the periodic block for `PSUM`.  The real reason the distance was to be enforced for vehicle is that vehicles can move.  Lockers and corpses can not.  The player can move too at the moment but probably shouldn't be able to move.  Whether that gets changed or not, a vehicle will still be able to move and, without a range check, its trunk would remain open on someone's screen.

Fate-JH:
I have added damage properties to facility turrets, partially.  The process of "destroying" turrets has been truncated to one where they can be brought down to 1 health and anyone manning the turret at that time will be killed.  The turret will still be mountable and can be fired at 1 health; but, one shot of nearly anything against the turret will kill the operator.  The reason it is implemented this way is that I do not want to deal with the issue of handling how facility turrets are resusitated upon being repaired.  That will require some more time and this is supposed to be a _quick_ job.

Just make certain they get repaired.

PR #237, opened 2018-11-11
Maxes and Turrets Fixes

Fate-JH:
I'm not going to delude myself, or you, into thinking that this solution is assuredly a correct application of the packet code for `0x18` `DetailedCharacterData` streams.  Instead, for now, I will merely state that this change for MAX's is justified because it allows MAX's to work without their character data breaking down on the client's end of things.

Included is also a change to the initialization order of server objects that still need to register GUID's outside of their formal object construction.  This should handle awkwardness with many turrets, so long as the `Maps` code is good.  Let's be honest here: turrets causing otherwise stable code to break because, somehow, the same number as the external container is being applied to a contained object for no good reason is a silly explanation.  If that really is the problem statement, and it has nothing to do with mixed random-fixed selection patterns, then it must be an intentional mistake.


Mazo:
👍 Looks good to me

PR #241, opened 2018-12-07
Proximity Terminal Fix

Fate-JH:
Now with not-breaking-the-server action!

The original proximity terminal concept suffered from three major problems.  The first was that it operated on `WSA`-specific timers that require careful construction and deconstruction on top of each terminal's own internal machinery.  Obviously, it wasn't handled with enough care.  The second problem is that the terminal's timer situation was always operating on a "renew" scheduling that required receiving `ProximityTerminalUseMessage` packets to keep the system going, else it ended at a fixed delay.  Expecting this much consistency from the network while expecting both external and internal messages was too much.  The third problem was a curious issue where, through empathy, all `WSA` representing clients connected to a zone's event bus eventually upgraded to act like they were using a given proximity terminal if one player was using that proximity terminal, even though the other players did not get its curative effects,  The constant fluctuations of timers and events caused by a player trying to flutter the terminal's effects (let alone stop using the terminal normally) resulted in massive network congestion from clients struggling to figure out whether they should use the terminal or not.

The new model uses a callback system and has internal machinery only within itself.  Internally, a respective terminal tests whether a target is close enough to feel the effects of the terminal, and will determine if the target is still in a valid state where it would need the terminal's effects.  If either of these conditions fails, the terminal ends its affiliation on the user on its own.  The system synchronizes with `WSA` for player targets (like with `medical_terminal`) or with the `VehicleService` service (`such as with repair_silo`) as those are where the damage dealing code for the given target - players or vehicles - is synchronized.  This is far more organized than the previous system.

To cut to the chase, though: medical terminals, repair/rearm silos, landing pads, and Lodestar proximity repair.  Go play test.

**`MedicalTerminalDefinition`**:
The class has been completely reworked and serves as the basis for proximity terminals, for now, since all terminals currently in operation offering healing effects.

**`ZoneAware`**:
A harsh compromise on my behalf where I realized that I needed to know in which zone the Lodestar was located to dispatch events to it properly.  The class either doesn't have enough use yet or has too little use over-all and should be reconsidered.  For now, it does what it needs to do.

**`UtilityWorldEntity`**:
Internal offset positioning for utilities.  Mostly useless to us, probably, save for the Lodestar.

**`Vector3`**:
Mathematic vector rotation code has been added.  The function `closeToInsignificance` is me throwing my hands up in the air over `Float` calculations.  I'm satisfied with the result; but, it was still a "throw arms in air" moment.

**`CommonMessages.Use`** and **`CommonMessages.Unuse`**:
An optional data portion was added to the messages.

Caveats:
- The distance calulations are tricky.  A certain range exists where the client will detect being close enough to attempt the terminal; but, the server judges the terminal too far away from the target and stops the action.  This is because the client is aware of the proportions of the target's model, while the server is only aware of the target's center point.  There is no solution for this at the moment and the overhead to be incurred by this situation is not known.  For most terminals, no effect will actually be witnessed.  For the repair silos, however, they will open and close repreatedly like claws grasping for one's damaged vehicle.
- Only the previously-established proximity terminals on `z6` (Ceryshen) and `home3` (VS Sanctuary) have been properly updated to working status.  Any terminal that isn't in working order right now is in such a state probably because the location of the terminal is set to the world origin and it's impossible to move into the server-recognized activation radius.
- That said, proximity terminals now need their world coordinates defined to work properly.  The old coordinate-less syntax is there just so that the code doesn't break.  It should, however, be considered deprecated.
- Various equipment terminals have been updated to use the new style order terminal syntax.  In the case of the repair silos and landing pads, their definition classes have been split up into a dedicated repair terminal and a dedicated rearm terminal.  `RepairRearmSiloDefinition` is now a defunct class and has been removed from the project.
- The delay to reset terminal order acceptance has been moved to the end of each `case` rather than the whole `match` statement.  Why?  who knows.

Addenda:
- While working on the code for positioning the invisible lodestar repair terminals, I sought a means of visualizing where they were located and thus produced a "marker" system that rendered `AmmoBox` objects where they were centered.  The `AmmoBox` objects would update position as the Lodestar moved and turned.  The code is no longer part of the PR but the concept might be useful for other things in the future.  All it costs is a GUID and some network load per marker.
- Facility wall turrets are actually kinda OP.  lol


Fate-JH:
These other portions of the proximity terminal unit system on the `WSA` have outlived their usefulness and only exist to impose unnecessary complexity and risk.

PR #244, opened 2019-02-21
IFF/Door orientation - Building MapID/GUID changes - Ishundar full config

Mazo:
What started out as small fix to door orientations ended up a bit bigger as other issues around Building objects appeared.

1. The short version is door orientation in the UBR files isn't correct. They can be randomly orientated 180 degrees either way, which doesn't help when trying to discover inside/outside orientation. As it turns out, the IFF locks DO have their orientation set correctly. As such, the code around door orientation has been moved to IFF locks instead.

2. On top of that, there's always been some ambiguity between Building Id/GUID/ModelID/MapID. ModelID has been removed and has been simplified down to just GUID and MapId. The MapID used for BIUM packets is the order that the building appears in the MPO files, whereas the GUID is generated in a different specific order and is used in other packet types. I've somewhat hacked my way through changing the Building construction process, so it may require some follow up refactoring.

3. The final change is the full Maps.scala for Ishundar is now present (with ownership based on @SouNourS V3 playtest). Default zone has been changed from z6 to z4 and all other Zone buildings have been commented out for now, if no issues arise from the Ishundar config I'll generate all the other zones and push them in a separate PR.


Mazo:
Couple other quick fixes as Nick pointed out that spawning at a base/tower points you to the East, instead of North where all spawn doors seem to be orientated.

Another quick fix for a second issue spotted by Nick is that tech plant garage doors had the wrong inside/outside orientation. This is down to the fact that those doors are the only ones in the game where the lock does not match the orientation of the door. As such I've just added a 90 degree offset in my map generator script, with a comment explaining why it's needed.

Mazo:
Fixed a couple of missing objects such as air/ground repair & rearm terminals.

One other thing I noticed is in the game_objects.adb file amp_station, comm_station and cryo_facilities have their vehicle_terminals remapped to vehicle_terminal_combined, so that has now been accounted for.

Also, it seems that the correct terminal would be vehicle_terminal for normal ground terminals, but the current scala code uses ground_vehicle_terminal, so a workaround was added to the maps generator to conform with this but may need changing.

Mazo:
Another thing I forgot to note was that currently all spawn pads use the same VehicleSpawnPad object. This will need splitting as the different spawn pads (e.g. dropship_pad_doors and mb_pad_creation) have differing Z offsets to prevent vehicles from spawning in the ground and falling through the world.

PR #245, opened 2019-03-07
Full map configs for all above ground continents including battle islands and oshur prime

Mazo:
Pretty much what it says on the tin, all the maps generated from my C# tool.

A few things to note:

1) All the maps have been split into their own files, as a 50,000+ line Maps.scala was a bit unwiedly.
2) Additional parameters have been added to LocalObject to cut down on common repetitive lines like ObjectToBuilding
3) This PR is branched off my other branch from PR #244 as it included changes to the way buildings are constructed. I would recommend merging that one first


Mazo:
Couple extra fixes. Spawn pads now have the correct object types and orientation and have Z orient offsets (to rotate vehicles the correct direction compared to the pad orientation from) and Z offsets (to stop vehicles clipping through the pad) from game_objects.adb.lst

Maps have been updated accordingly for the above, and a few missing GUIDs for sancs have been resolved.

@Fate-JH If you wouldn't mind checking over the last commit with the duplicate ODM packet, I'm not sure if this is something old and missed when refactoring as everything else seems to use DetachFromRails or if it covers a weird edge case. Either way, it was causing duplicate ODM packets to be sent.

PR #248, opened 2019-03-30
Gating

Fate-JH:
The process of gating and vehicle gating is fairly straightforward. 

That was a lie.

- Starting with basic infantry gating, it operates no differently than respawning or reconstructing at a tube somewhere.  The "tube" in this case is only about 300m in radius.
- Following this simple procedure is the slightly more advanced procedure of moving a single-player driven vehicle which involves adjusting the avatar creation routine to retain knowledge of the vehicle while transitioning the player between zones.  Additionally, the zone loading protocol must be adjusted to not re-rebuild the vehicle since that would wipe away the avatar-level player that was created for the client.
- Following this slightly more advanced procedure is a sufficiently more complicated procedure where passengers of the previous driven vehicle are introduced to the zone.  For this operation, the driver of the vehicle operates a temporary messaging system that alerts the passengers when it is safe for them to start the same process that he went through.  The zone loading protocol is once again modified to draw all vehicle passengers but the player and to consider the possibility that the player is not the vehicle driver.
- Finally, the "sufficiently more complicated procedure" out of the way, a somewhat similar feature is procedured that allows vehicles that are treated as the cargo of other vehicles to be transported between zones.  The driver of the cargo vehicle is treated as one of the passengers of the carrier vehicle's passengers and that instigates the single-driven vehicle protocol which eventually calls the passenger protocol again.  Presumedly, vehicles that are nested as cargo of other cargo vehicles (of other cargo vehicles ...) also apply but we don't have any of those.  The zone loading protocol is, as expected, modified.

Four important technical aspects of the procedures: the driver and the passengers of the vehicle being transported, as well as the cargo vehicle being transported with the carrier vehicle, are never moved from where they are currently installed, either in their seat or in their cargo hold; whoever is the last participant in the process is responsible for deleting the ghost of the vehicle being gated from the previous zone; rather than being explicitly transported, vehicles are implicitly retained during transport using local variables; the differences between spawning, reconstruction, and gating have been bridged, producing a single consistent workflow regardless of how getting from point A to point B is planned.

__Features__
* _SpawnPoint_ and _SpawnPointDefinition_:
Originally, anything that could be spawned on was a type of class called `SpawnTube`.  That was because everything that the player could spawn on previously was a type of amenity called a "spawn tube."  With the existence of warp gates as building-wide spawnable locations, the essense of spawn positions has been abstracted into this class.  Additionally, the logic of spawn positions has been contained to this class's implementation through its specific definition, though it remains in its somewhat compromised calculations.  Being recognized as a "spawn point" requires three things - extension of this class, a globally unique identifier in some zone, and a location in some zone.

* _Building_:
The building definition system now permits different definitions.  That said, there's still only one definition of consequence at the moment - `building`.

* _WarpGate_:
Warp gates benefit the most from the variation in building definitions since more than one warp gate type exists on the continents.  Surprisingly, there's never more than two types per continent - warp gates and external cavern gates for most, and a battle island gate and the external cavern gate on Nexus.  The definitions for warp gate structures found in thre ADB contain a variety of fields that seem to mark radii and distances used in determing the ultimate spawn position.  I ignore most of that for an adequate simpified system that uses only radius.  Being recognized as a "spawn point" for warp gates involves all of the previous criteria and one other thing - a map id.

__Caveats__
* When using warp gates, unless piloting a flight vehicle, the destination will most likely drop the player a meter onto the ground.  This is a consequence of using the coordinates of the center of the spawn platform in terms of where a player can stand.  This will often lead to flippped ground vehicles when trying to resolve collision with obstructions in the game world within the warp gate bubble, such a trees.  Something else may have to be produced when we use actual in-game world coordinates for the warp gate.  For now, deal with it.
* If one becomes stranded during a failed vehicle gating when the passenger of a vehicle, or they feel the process is taking too long, they can skip out of the gating process and bail from the vehicle.  This is not just a cosmetic fix - is actually disengages the player from the vehicle, so that they are no longer occupying a seat in the background (locking it from being used).  A ghost of the vehicle that the player just exited will hang in the air above and must currently continue to exist so that fellow former passengers do not get despawned unceremoniously, should they also skip  If the vehicle's GUID is reassigned in that zone, however, then the vehicle will be despawned anyway, rednering the previous consideration moot.
* Disengaging a failed vehicle gate operation is not possible at the moment.
* The constant one second spawn penalty for being gated when the passenger of a vehicle that is levied is intended to allow the system to calm down.  It also is intended to reduce some animation quirks.  Regardless, some animation quirks remain; for example, other players will often see a player who joins a zone as a vehicle passenger manifest from out of nowhere and re-mount the vehicle while it is still being driven.

__Addenda__
* The AMS spawn system has been revamped.  Complications with the previous system stemmed from a lack of understanding about how `BindPlayerMessage` actually interacted with the client deployment screen, as well as a compulsive desire to "undraw" and "forget" the previous spawn point for no good reason.  The usual problems involved a never-updated spawn entity on the far other side of the continent remaining available or a non-spawnable entity being reported at the world origin.
* As part of the unification between reconstruction and gating,  `CSRZone` and `CSRWarp` should now properly deal with vehicles being moved when the driver tries to move around using the commands.
* In a vehicle being treated as cargo to a carrier vehicle, where that cargo vehicle has both a driver and passengers, if the driver disconnects for some reason, the vehicle they were piloting is ejected from the carrier vehicle before the passengers are kicked out.  This resolves an issue where the camera of the passengers would get stuck in the post-bail-landing animation angle if the cargo vehicle was still being treated as "cargo" at the time.
* Vehicles can now tell when they are flying without needing a height map.  This was always possible but a little research was needed to make certain this was an appropriate use of the field.  (A height map would probably still be good for server-based polling and validation.)
* A compromise between the `Actor` callback system for updating building information and a direct access method for accessing building information at client initialization time has been struck.  By "compromise" I mean we're doing it my way.
* I don't know why Girru, Ishundar starts with 0% NTU while the other facilities on that continent have full NTU.  This will probably affect my Anguta, Ceryshen test site in the future.
* The `Hackable` internal field has been adjusted to discard the object reference to the player who did the hacking and instead use theirt name s a reference back to the player.  It also now features fields such as the start time and the duration.  The service support `Actor` object has not been changed functionally, though it will update those new fields.  The duration field has been set up in a way that it can be updated later.
* The lengths of time required to respawn have changed.  Before, it was always a 10s respawn time in all situations.  Now, it's a 0s spawn time for going to one's own sanctuary, 10s for leaving sanctuary (provisional), and the respawn time is determined by the destination spawn point in all other cases.
* I have a terrible, terrible cold.


Fate-JH:
Minor update:
Performed tuning on code that handles cargo vehicles.  Asides from safety checks to the process, the most significant changes involve how a cargo vehicle is positioned when it bails, how a cargo vehicle dismounts when the carrier is landed, and how a cargo vehicle is kicked by the carrier pilot.

No progress on stranded cargo vehicle gating bailing.

Fate-JH:
I redid this branch from the ground up to utilize the other PR's as its base and then overwrote the `gating` branch.

Updates:
- Sanctuary respawn tubes for the `VT_building_*`'s now use their own definition, as compared to other respawn tubes.  This affects `home1`, `home2`, and `home3`, obviously.
- As a "may have forgotten to mention" point, the only warp gates currently set up involve two of the three on `home3` and the four normal gates on `z6`; and, they only are gauranteed to work as broadcast gates.  The two geo warps on `z6` have entries but should not be active.

PR #249, opened 2019-04-11
Exclude Equipment from Loadouts

Fate-JH:
In response to https://trello.com/c/rs2X5SqT/27-anti-infantry-max-armor-can-be-acquired-at-an-ams-terminal, loadouts can now be blocked in whole or in part.


Fate-JH:
Update for incorrectly reporting exo-suit types and subtypes in `FavoritesMessage` packets.
Update for fallback exo-suit type should the loadout the player selects not be permitted from his current certification values.

Fate-JH:
Update for prioritization of vehicle trunks over player backpacks when finding places to put purchased equipment.

PR #254, opened 2019-04-17
Gating Fixes

Fate-JH:
In response to feedback from the last `gating` PR and in an attempt to resolve some lingering concerns.

Spawn times when traveling via warp gates is now as instantaneous as possible.

Subsequent clicks on the same deployment map options will not process the additional invalid data.

Broadcast gating now works per faction rather than just for the Vanu Soverneigty.  To demontsrate, some gates have been made active for the Terran Republic only, while other gates have been made active for the VS only.


Fate-JH:
Changhe peculiarities about vehicle disownment at log-out.

PR #255, opened 2019-04-19
Door closing modifications

Mazo:
Increased the distance on the door close check to stop doors bouncing open/closed and added a check to make sure the player is alive when holding the door open


Fate-JH:
Corpses did hold doors open on Live.  Since we have access to the zone in the entry, you should be able to check if the player can be found in either `zone.LivePlayers` (if not backpack'd) or `zone.Corpses` (if backpack'd) for sufficient existence.

Mazo:
I'm assuming the advantage of checking in `zone` is because the `player` object may have released and respawned, thus the check wouldn't be correct for the old `player` avatar left behind?

Fate-JH:
I want to validate against the `zone` object's fields to avoid shenanigans.  If you want to do it another way, modify `CorpseRemover`'s `FirstJob` so that the unreleased backpack is internally moved to the world origin.

PR #258, opened 2019-04-26
Cargo Bail

Fate-JH:
Let us review the current.

As things stand, when a vehicle is deconstructed, all of its passengers are kicked out of their seats.  This includes the driver seat occupant.  If a vehicle's driver seat occupant bails, and the vehicle is a flying vehicle, the vehicle deconstructs to clear away a now inaccessible hunk of metal.  When a vehicle deconstructs, it deconstructs everything about the vehicle - its still-mounted passengers, its cargo, its trunk contents, and its utilities.  This is, of course, bad in a variety of ways, such as the example when a carrier vehicle deconstructs abruptly and it still has a cargo vehicle that stops existing, including on the client of the former cargo driver.  Further issues occur when players disconnect from the game, resulting in similar instances of game objects being deconstructed and player's being stranded in a state of nonexistence.  The entire process becoming stranded, leaving ghosts of the previous objects that no longer exist, either on specific clients or all clients, was also always a possibility.

To avert this problem, the following conditions have been implemented: 
1. if a deconstructing vehicle is cargo for some other carrier vehicle, it will dismount as cargo before the carrier deconstructs; (`VehicleRemover`)
2. passengers will not be kicked out of a deconstructing vehicle until it is no longer counted as cargo to some other carrier vehicle; (`RemoverActor`)
3. if a deconstructing vehicle is the carrier vehicle for some cargo vehicle, it will dismount that cargo vehicle before before deconstructing; (`VehicleRemover`)
4. if a dismounting cargo vehicle has no driver, that cargo vehicle will be submitted to the `VehicleRemover` for immediate deconstruction. (`WorldSessionActor`)

It is important to note that the second point is a current condition and actually expresses task priority rather than enhancements.  This is a workflow with cyclical potential.  Not being a cargo vehicle is necessary for any passengers being kicked from that same vehicle to avoid misappropriating the camera to one of a permanent bailing view, with no first person privileges.  Additionally, this covers the potential case of a carrier vehicle ferrying a driver-less cargo vehicle with only a passenger in it; and, the carrier vehicle's driver bailing.  Previously, the passenger would stop existing when the chain of parent-child entities dissolves.  Now, that passenger is ensured a proper bail before any chance of accidentally being eliminated from the game world when both vehicles are cleaned up.

__Caveats__
- We are still relying on the `VehicleRemover` to play catch-up against the fact that vehicles still can not be moved independent of a client-connected driver.

__Addenda__
- The same ownership operations extracted from deployables into the `trait` `OwnableByPlayer` is now also used by `Vehicle` entities.  The GUID-based owner field is used for transient ownership; the name-based owner field is used for more permanent ownership identification.  Half of this functionality was already native to vehicles.  Now, it is inherited by both deployables and vehicles.  It was extended to vehicles to assist in management ownership between zone loading and avatar respawn operations.
- Vehicles are now disowned more explicitly and less haphazardly.  The process of losing ownership is still kept distinct from the process of being submitted to `VehicleRemover` for deconstruction.  The only special condition should be for a vehicle in flight.
- As was mentioned, the logic for regaining ownership of vehicles on both zone loading and avatar respawn (both of which are piped though similar protocol) has been tuned.  One should retain control of vehicles in conditions where it is important.
- As a product of giving vehicles the same ownership fields as deployables, logic could be implemented that coming back to a zone with a previously-used vehicle in it allows the player to regain ownership of the vehicle in a way similar to how ownership of deployables may be retained.  As no evidence of the vehicle re-inheritance was observed on Gemini Live, unlike minor evidence of deployable re-inheritance, this logic has not been realized.
- The first four of these Addenda notes involved ownership.  So does this one.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
+++ b/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
@@ -38,4 +38,28 @@ trait OwnableByPlayer {
     }
     OwnerName
   }
+
+  /**
+    * na
+    * @param player na
+    * @return na
+    */
+  def AssignOwnership(player : Player) : OwnableByPlayer = AssignOwnership(Some(player))
+
+  /**
+    * na
+    * @param playerOpt na
+    * @return na
+    */
+  def AssignOwnership(playerOpt : Option[Player]) : OwnableByPlayer = {
+    playerOpt match {
+      case Some(player) =>
+        Owner = player
+        OwnerName = player
+      case None =>
+        Owner = None
+        OwnerName = None
+    }
+    this
+  }
 }
--- a/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
+++ b/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
@@ -1,0 +1,41 @@
+// Copyright (c) 2019 PSForever
+package net.psforever.objects
+
+import net.psforever.packet.game.PlanetSideGUID
+
+trait OwnableByPlayer {
+  private var owner : Option[PlanetSideGUID] = None
+  private var ownerName : Option[String] = None
+
+  def Owner : Option[PlanetSideGUID] = owner
+
+  def Owner_=(owner : PlanetSideGUID) : Option[PlanetSideGUID] = Owner_=(Some(owner))
+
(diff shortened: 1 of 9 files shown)


Fate-JH:
Update for landed-dismount logic, and checks for current CSR comands.

PR #259, opened 2019-05-06
Adding PSUtil Support

stalkopat:
This PR adds PSUtil Support, open to criticism as i have no idea if i integrated it in the right place, i tried to create custom Packet Types but got very confused and it felt out of place, this approach differentiates a PSUtil packet during the initial session creation, they are prefixed by the Opcode 0xf4, 1 above the Highest known Game Opcode wich is 0xf3, that way there shouldn't be any collisions with normal packets. This shouldn't affect the SessionRouting at all although I cant know for sure because im getting an exception on trying to connect onto my local server.


pschord:
I know we never followed up with this, but I don't believe we're ready to integrate this into upstream until we consolidate around a single launcher. Once we have a fully working launcher with client-side verification, then we can start doing some extras like server-side verification for mods and stuff like that. That's a ways off and is going to require some more thought.

PR #260, opened 2019-05-08
Hotspots

Fate-JH:
I actually like the visual cues PlanetSide 2 employs as hotspots
but not everyone can have nice things.
We're stuck with pulsating ten-point stars.

___Features___
__`MapScale`__
All `ZoneMap` information now keeps track of the maximum map to ground dimensions.  This is useful, maybe?

__`HotSpotInfo`__ and __`ActivityReport`__
Each hotspot is generally shared by another faction and may even be visible to all three.  As a consequence of this, each hotspot needs to keep track, individually of how relevant it is to each of the factions.  The two measure of this are `heat` - a raw count of the amount of activity that has been detected recently - and "time since last reported activity."  When determining whether to draw a hotspot, `heat` is used.  When determining whether to clear a hotspot, "time since last reported activity" is used.

__`ZoneHotSpotProjector`__
The projector manages communications between the hotspots maintained on the particular zone object and those currently known to the client, though the ones that are actually "known" to the client are only understood to be the ones currently "known" to the server.  Excepting where one empire can not see the hotspots of another empire, which is a special rule (see `Zones.HotSpots`), hotspots that are hidden for some reason do not exist.  The projector directly interfaces with the `GalaxyService` `Actor` object as a method of messaging communication.  Originally, it was intended to sit atop of the `GalaxyService` as another "support actor;" but, putting it on top of a specific zone's `ZoneActor` was much more versatile. The blanking phase where old hotspots are removed from the map is set to a steady fifteen second interval, regardless of what happens.

__`GalaxyService`__
Individual channels for each faction are now expected.  A special `Service.Join` handling case is provided.

__`Zones.scala`__
Zones keep track of the ruleset used by their projector to determine if and how long, and where hotspots should be depicted on a continental map.

___Caveats___
- Certain things one might call "combat" do not cause hotspots under this  preliminary system setup.  Currently, only cross-factional damage dealt to other players causes hotspots.  ~~Vehicles taking damage should not cause hotspots.~~  Amenities and deployables taking damage should not cause hotspots.
- Only 4095 hotspots may be depicted on a specific zone's deployment map per client.  The limit is enforced by _datatype_ - the size of the list is represented by a 12-bit integer.  Attempts to define 4095 in one packet and then send another packet with additional hotspots will find the effort futile as each hotspot packet clears the previous hotspots and draws only the hotspots reported in the most recent packet.  This boundary is noted because hard limit has been written into the `HotSpotUpdateMessage` packet `codec`.
- Also in regards to `HotSpotUpdateMessage`, the use of the `priority` field is unknown and the `scale` fiield for each hotspot remains a juryrigged  kludge trying to find meaning where none may exist.  The issue is ongoing ...
- "Heat" in `ActivityReport` is generally a worthless statistic, only indicating a level of activity equal to the amount of projectiles causing damage, per faction.  It does not change the graphic.  It does not change the graphic's size.  It does not change the hotpot's duration.  While it does generally imply that hotspots with greater `heat` can be considered well-worn regions of intense combat, no specific information about the actual form of combat can be determined just by looking at the value.
- Looking at the function assignments, no hotspots should be generated in the sanctuary-specific zones.

__Addenda__
- I say "hotspot," not "hot spot."  Even the client can be inconsistent.


Fate-JH:
Update:
- hotspots are now caused by damaging enemy vehicles
- addenda fix: cargo vehicles now see the damage ticks that their carrier vehicles suffer
- addenda fix: cargo vehicles and the passengers of cargo vehicles are now handled properly when the carrier vehicle is destroyed

PR #264, opened 2019-07-13
Vehicle jacking

Mazo:


Mazo:
One potentially important change is only the player that owns the vehicle will process VehicleResponse.Ownership. Hopefully this should resolve the issue from the playtest where vehicle ownership was changing randomly.

PR #267, opened 2019-08-23
Turret Fix

Fate-JH:
It's kinda like the vehicle terminal - stop breaking it.

The latter part of the facility turret update is now indicated by an on/off value.  The flag is not actually attached to any work in the backend; rather, it merely indicates when the `TurretUpgrader` operations believes that the turret is ready to be used.  It blocks allowing people to mount the turret and it also stops the turret weapon from rendering on client initialization of a zone.  (The turret weapon should be drawn normally by the `TurretUpdater` operations when it is ready.)  An additional change was performed on the sequencing of `AmmoBox` registrations and de-registrations to make certain that the turret knows it is ready only once the new `AmmoBox` objects are ready, and make certain that the same `AmmoBox` object isn't worked on in both operations.

`Caveats`:
* If any turret gets stuck during its update process, it will block entry and, on zone loading, appear as a turret without a weapon.  Use `/destroy` while pointing at the turret to force it to downgrade to standard setting - `None`, i.e., SGL Heavy Gattling Cannon - and that should correct the hang-up.
* We are still subject to the "immortal turret" hack which will endure until static server objects are ready to be revived from a destroyed state.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/serverobject/turret/FacilityTurret.scala
+++ b/common/src/main/scala/net/psforever/objects/serverobject/turret/FacilityTurret.scala
@@ -11,6 +11,7 @@ class FacilityTurret(tDef : TurretDefinition) extends Amenity
   with StandardResistanceProfile {
   /** some turrets can be updated; they all start without updates */
   private var upgradePath : TurretUpgrade.Value = TurretUpgrade.None
+  private var middleOfUpgrade : Boolean = false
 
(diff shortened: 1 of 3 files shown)


Mazo:
Might be worth putting a faction check on the /destroy change otherwise it may be abused in a fight to downgrade turrets from a distance by other factions?

Mazo:
Ah fair enough. I assumed that would reset the turret back to its base state.

PR #271, opened 2019-08-27
Utilities Fix

Fate-JH:
This problem should have actually affected all forms of vehicular utilities that were transported from one zone to another.  Only the AMS's equipment terminals were being called-out.  In short, while a vehicle's control object was correctly being transported over to the new zone on transfer, the vehicle's utiltities did not have the benefit of their control objects moved over to the new zone.  The control objects had to properly be destroyed before they could be properly rebuilt.

Also included are some quality checks for dropping items and picking up items.


Mazo:
:thumbsup:

PR #276, opened 2019-10-09
Teamwork

Fate-JH:
While there is no "I" in "team," you can find "me" and a "mate" in one.
I came up with that just now and it shows.

All squad operations begin with the would-be squad leader either publishing a squad structure (loadout) or issuing an invite some other player of the same faction affinity to an impromptu squad.  In the former case, another player of the same faction would either send a role bid message to indicate that they would like to join a specific role or the squad leader could, again, issue an invite some other player.  All previous operations generate an invitation event that is either responded to with an acceptance into the squad or a rejection and the original relationship between players is maintained.  Squad leaders may also search for player to fill out the squad or fill certain roles in the squad.  Players can be kicked or leave of their own violation.  Waypoints can be displayed in conjunction with the squad from the command interface on the continental map.

Basic game loops to test involve: starting squads by messing around with roles and settings; joining and leaving squads in various weays; creating, saving, and loading squad loadouts; moving around waypoints; spawning, dying, and respawning while in a squad; creating complicated squad join request queues that need to be manipulated when the squad leader changes hands (someone else is promoted).  Essentially, a lot of the ground is open for issue to be discovered.

No platoon operations are supported at the moment.  I'm sure you'll find something to break by exploring those waters.

I make no promises of stability.

___Features___
__`ReplicationStreamData`__, __`SquadDetailDefinitionUpdateMessage`__, __`SquadDefinitionActionMessage`__
The structure of these packets were updated significantly to reduce complexity of the creation of packet data and the translation from object data into packet data, including partial composition and completion.

__`Loadout`__ and __`EquipmentLoadout`__ - __`Infantry...`__, __`Vehicle...`__, __`Squad...`__
Refactored code to create super classes supporting more concrete implementations of the loadout.  `InfantryLoadout` and `VehicleLoadout` continue to exist but are connected by a new super class, and `SquadLoadout` was created from the original super class.  The division allows them all to be known as "loadouts" but makes it apparent that only the former two utilize object construction.

__`LoadoutManager`__
Common behavior for different kinds of loadouts.  This strips out functionality originally located in the `Avatar` class and makes it generic for the sort of loadout being handled.

__`Avatar`__ and __`Player`__
The unique character identifier is now a required constructor parameter for the former, and merely accessible from the latter.  Loadouts having to do with equipment (and vehicles) and loadouts having to do with squads are made separate.  The "Looking for Squad" state is also associated with a backend field.  The important changes have been worked into the `AvatarConverter`.

__`Member`__
Representative of a single role in a squad.  Contains sufficient fields to represent both the role's details and the player who joins the squad in this role.

__`Squad`__
Representative of a group of roles for players to join, specifically designating one of them as the "squad leader."  Contains sufficient fields to represent both roles and facets of the squad such as zone of activity and tasking and whether or not the squad is publically listed.

__`SquadFeatures`__
Also representative of a squad but more focused towards supporting information as opposed to cardinal aspects of the squad.  Anchors a still unrealized `SquadSwitchboard` `Actor` object that was intended to make squad messaging less complicated.

__`WaypointData`__
Represents indicators on the map that can be set by the squad leader.  Closely related to `WaypointInfo` which can be found in its packet exchange protocol for squad waypoint request packets (but different from "squad waypoint event" packets).

__`SquadMembershipRequest`__ and __`SquadMembershipResponse`__
Packets for communicating squad mership change activity.  `...Request` is dispatched by clients to suggest or promp changes in squad composition, such as invitation requests or leaving a squad.  `...Response` can be used to reset fields automatically modified by `...Request` but mainly generates event chat messages for clients.

__`WaypointRequest` and `WaypointEvent`__
Manages squad waypoints, of which five are defined.

__`SquadState`__
Updates the information about other squad members in the client squad user interface (internally called "cards") in the client.  These cards are managed through `SquadMemberEvent` packets otherwise.

__`SquadService`__
The monolithic service class that regulates all facets of squad management, invitation, creation and update relay, including waypoint control and intersects squad loadouts.  Honestly, it was never intended to be this one-stop class for squad operations.  All squads were intended to have their own `Actor` object.  Invitation queues would also have their own `Actor` objects.  The system would be parallelized neatly. The growing complexity of the interactions and invitation model required that I stick to this monolithic design until I have a much clearer picture of what was going to be necessary.

___`Caveats`___
1. Way too many to properly itemize.
2. Squad chat is not integrated due to the fact that a chat communication service is not currently built for the development branch, though one does exist for the live version.  A channel name is generated for subscription and is communicated to `WorldSessionActor`, however, to ease accommodation of the service.  Actual accommodation is outside of the scope of my work.
3. When a would-be squad member request to join a squad in a given position and the squad is not set to auto-allow these join requests, the squad leader should receive a message prompt similar to the leader sending an invitation to a would-be squad member.  This message prompt does not spawn despite my tests and replication of squad packet passing, however, so a workaround was developed to push a string to communication chat ~~and repeat it until the squad leader responds~~.  Instructions for responding are included in the message.  The actual message is hardcoded in the `WorldSessionActor -> SquadResponse.WantsSquadPosition` case and may require tweaking for readability/authenticity.
4. When displaying squad waypoints in between zone switches and respawns, a blue waypoint may become fixed somewhere on the map in a location that gravitates the point where the squad leader last respawned or was originally positioned.  It should not distract too much if it does show up.
5. Squad messages may not display in all appropriate situations, even though the appriorate action is executed.
6. The most complicated tests only involved four clients simultaneously.  Trying to deal with larger scale tests locally causes my system to suffer to the point where testing becomes strained.  As a consequence of this, expansive interactions such as the sort that would manage active (and in-wait) invitations from old squad leader to new squad leader upon promotion may not function correctly.  Despite my best efforts, a number of features are only "assumed fully operational."
7. The "Looking for Squad" marquee is transformed into a "Looking for Squad Members" message when employed by a squad leader.  All other squad members can not display this or even toggle the field.  When promoting a member to a squad leader, however, the new leader should gain the ability to broadcast this "Looking for Squad Members" message.  Currently, the new leader does not; that is, the message does not broadcast.  I did not have time to explore why.  ("Looking for Squad Members" is maintained locally in the `WorldSessionActor` instancwe for that client.)
8. The icons next to squad members on the respective tab are probably wrong.  Currently, it looks like a boomer trigger? the cab section of an advanced mobile spawn?  I did not have time to explore what packet and what fields modify them.
9. While it is possible to search existing squads and squad roles based on provided criteria, on top of not knowing what kind of search priorities existed, I do not know how to return this information to the client.  I outline this limitation twice.
10. The squad member does not seem to need to be in the first role of the squad but the actual mechanisms that support this freedom of role assignment do not exist in our packet captures.  Additionally, the sqad member assignment controls to which the squad leader should have access do not interface correctly and only the squad leader may be moved around by the squad leader currently, resulting in the above situation where moving away from the first squad role resuls in the squad becoming rogue (and later generating errors).  The UI cards for squad members have related updating hangups when this situation arises.  For those reasons, the enumerated functionality had been either limited severely or has been disabled outright.
11. Different squad invitation searches do not respect the same priority.  Proximity searches use a "best fit" method where a would-be member will try to be placed in the role that best suits their certification allowances.  Searching for "Looking For Squad" self-flagged players and generic invitations by the squad leader (and other ...) result in the earliest available role being selected.  In fact, the "LFS" role-by-role invitation method does not abided by the correct scope - it should search globally for players that are appropriate and, instead, only searches the local zone.
12. One can not execute a proximity search without first starting a squad in some capacity.  This is a known oversight.
13. Squad leaders can block players from joining a squad for the remainder of the squad's existence.  Players can block squad invitations from other players for the remainder of their current session.  Entries are also updated in the background depdning on the player's responses to rejection of invitations.  This is as much a quality oif life as anything.  As of the moment, no mechanism allows clearing these refusals without the squad dissolving or the player logging off, respectively.
14. `LocationFollowsSquadLead` currently can be set and unset but does not reflect the presentation of the squad in any known way.  I do not know how to set the "default" for this field.
15. `AutoApproveInvitationRequests` starts as active for convenience of the squad leader.  Also, I do not know how to set the "default" for this field.  If the user thinks the GUI checkbox has become disconnected from the actual state of the field, toggle it a few times to align the state.
16. Although code exists to allow players to automatically rejoin their last squad if they log out and in quickly enough, this delay is outpaced by the `SessionReaper` killing their `WSA` and that unsubscribes from `SquadService` and that forces the player out of the squad.
17. While a significant amount of effort was taken in the hopes of avoiding `DeadLetters`, it is possible to generate them by all the players of a given squad disconnecting at that same time.
18. `SquadService` currently is a lot like `AvatarService` in that it gets abused by every squad-affiliated player in the game.
19. `SquadSwitchboard` does nothing but get created and destroyed.  I know.

___Addenda___
Although it was definitiely not a working part of the original, the ability of a would-be squad member to watch changing squad details on their client has been enhanced.  Previously, if one loaded squad details, it would be a static snapshot of what the squad looked like when they requested it, lasting until they joined, asked to load a different squad's details, or asked to reload the same squad's details.

What is that <**?**> in the first commit message?


Fate-JH:
I neglected to completely switch waypoint events from the previous message-sending format to the new "`Publish`" format.  It should work now.

pschord:
Stupendous work yet again Fate! Clearly a multi-month effort. Cool to see your progression of reverse engineering in the commit log.

Fate-JH:
Repaired broken tests and change squad id management.
Please determine if changes from previous commit invalidates issues reported in Trello bug reports.

common/src/main/scala/net/psforever/packet/game/SquadState.scala
```diff
@@ -0,0 +1,101 @@
...
+/**
+  * Information about a specific squad member.
+  * @param char_id the character's unique identifier
+  * @param health the character's health value percentage, divided into 64 units
+  * @param armor the character's armor value percentage, divided into 64 units
+  * @param pos the world coordinates of the character
+  * @param unk4 na;
+  *             usually, 2
+  * @param unk5 na;
+  *             usually, 2
```
pschord: It's hard to determine if a packet field being constant warrants decoding it as a full int or as a fixed value. One leads to easier debugging when creating the packet, but moves the burden of providing the bespoke fixed values on the packet creator. Not sure if this is reasonable.

common/src/main/scala/net/psforever/packet/game/SquadWaypointRequest.scala
```diff
@@ -0,0 +1,115 @@
...
+  implicit val codec : Codec[WaypointEventAction.Value] = PacketHelpers.createEnumerationCodec(enum = this, uint2)
+}
+
+/**
+  * na
+  * @param zone_number the zone
+  * @param pos the continental map coordinate location of the waypoint;
+  *            the z-coordinate is almost always 0.0
+  */
+final case class WaypointInfo(zone_number : Int,
```
pschord: Dont we have an enum for Zones?

common/src/main/scala/net/psforever/packet/game/SquadWaypointRequest.scala
```diff
@@ -0,0 +1,115 @@
...
+  *                      0-3 for the standard squad waypoints numbered "1-4";
+  *                      4 for the squad leader experience waypoint;
+  *                      cycles through 0-3 continuously
+  *
+  * @param unk4 na
+  * @param waypoint_info essential data about the waypoint
+  */
+final case class SquadWaypointRequest(request_type : WaypointEventAction.Value,
+                                      char_id : Long,
+                                      waypoint_type : Int,
```
pschord: Int -> Enum

common/src/main/scala/net/psforever/packet/game/SquadWaypointRequest.scala
```diff
@@ -0,0 +1,115 @@
...
+  * @param waypoint_type the waypoint being updated;
+  *                      0-3 for the standard squad waypoints numbered "1-4";
+  *                      4 for the squad leader experience waypoint;
+  *                      cycles through 0-3 continuously
+  *
+  * @param unk4 na
+  * @param waypoint_info essential data about the waypoint
+  */
+final case class SquadWaypointRequest(request_type : WaypointEventAction.Value,
+                                      char_id : Long,
```
pschord: We should also consider creating a `type CharacterID = Long` typedef somewhere to improve type checking/readability

common/src/main/scala/net/psforever/types/SquadRequestType.scala
```diff
@@ -0,0 +1,34 @@
...
+  PlatoonDisband
+  = Value
+
+  implicit val codec = PacketHelpers.createEnumerationCodec(this, uint4L)
+
+  def toResponse(request : SquadRequestType.Value) : SquadResponseType.Value = {
+    val id = request.id
+    if(id < 6) SquadResponseType(id)
+    else if(id > 6) SquadResponseType(id - 1)
+    else throw new NoSuchElementException("request does not have an applicable response")
```
pschord: wtf planetside. can you explain what is happening here?

common/src/main/scala/net/psforever/types/SquadResponseType.scala
```diff
@@ -0,0 +1,33 @@
...
+  PlatoonDisband
+  = Value
+
+  implicit val codec = PacketHelpers.createEnumerationCodec(this, uint4L)
+
+  def fromRequest(response : SquadResponseType.Value) : SquadRequestType.Value = {
+    val id = response.id
+    if(id < 6) SquadRequestType(id)
+    else if(id > 5) SquadRequestType(id + 1)
+    else throw new NoSuchElementException("response does not stem from an applicable request")
```
pschord: same comment as SquadRequestType

common/src/main/scala/services/teamwork/SquadAction.scala
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2019 PSForever
+package services.teamwork
+
+import net.psforever.objects.zones.Zone
+import net.psforever.packet.game._
+import net.psforever.types.{PlanetSideEmpire, SquadRequestType, Vector3}
+
+object SquadAction {
+  trait Action
```
pschord: Please seal all traits that wont be extended to improve compiler type checking and to generate `Warning:(7, 3) match may not be exhaustive.` on partial matches
https://www.waitingforcode.com/scala-core/sealed-keyword-scala/read

common/src/main/scala/services/teamwork/SquadResponse.scala
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019 PSForever
+package services.teamwork
+
+import net.psforever.objects.teamwork.Squad
+import net.psforever.packet.game._
+import net.psforever.types.SquadResponseType
+
+object SquadResponse {
+  trait Response
```
pschord: Seal please!

common/src/main/scala/services/teamwork/SquadSwitchboard.scala
```diff
@@ -0,0 +1,142 @@
...
+          Watchers.remove(member)
+        case None => ;
+      }
+
+    case _ => ;
+  }
+}
+
+object SquadSwitchboard {
+  final case class Join(char_id : Long, actor : Option[ActorRef])
```
pschord: Do we need to make these extend from a sealed trait like the other messages?

Fate-JH:
This should correct issues with squad "listing" button on squad leave and re-creation.

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -2871,7 +3220,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
       import net.psforever.objects.GlobalDefinitions._
       import net.psforever.types.CertificationType._
       val faction = PlanetSideEmpire.VS
-      val avatar = Avatar(s"TestCharacter$sessionId", faction, CharacterGender.Female, 41, CharacterVoice.Voice1)
+      val avatar = new Avatar(41605313L+sessionId, s"TestCharacter$sessionId", faction, CharacterGender.Female, 41, CharacterVoice.Voice1)
```
pschord: Uhhh, what does this magic number mean?

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -2906,6 +3255,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
       avatar.Certifications += AssaultEngineering
       avatar.Certifications += Hacking
       avatar.Certifications += AdvancedHacking
+      avatar.CEP = 6000001
```
pschord: :godmode: orbital strikes incoming

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -3338,6 +3689,8 @@ class WorldSessionActor extends Actor with MDCContextAware {
               }
               vehicleService ! VehicleServiceMessage(continent.Id, VehicleAction.VehicleState(player.GUID, vehicle_guid, unk1, pos, ang, vel, flight, unk6, unk7, wheels, unk9, unkA))
             }
+            //vehicleService ! VehicleServiceMessage(continent.Id, VehicleAction.VehicleState(player.GUID, vehicle_guid, unk1, pos, ang, vel, flight, unk6, unk7, wheels, unk9, unkA))
```
pschord: comment looks like a copy and paste we dont need any more

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -4439,6 +4792,30 @@ class WorldSessionActor extends Actor with MDCContextAware {
             log.info(s"GenericObject: $player is MAX with an unexpected weapon - ${definition.Name}")
         }
       }
+      else if(action == 36) { //Looking For Squad ON
```
pschord: magic numbers. can we get an enum pls?

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -8898,6 +9276,71 @@ class WorldSessionActor extends Actor with MDCContextAware {
     }
   }
 
+  def SwapSquadUIElements(squad : Squad, fromIndex : Int, toIndex : Int) : Unit = {
```
pschord: Heavy business logic functions like these are excellent candidates for moving to a `private[WorldSessionActor]` separate file, if possible.

common/src/main/scala/services/teamwork/SquadService.scala
```diff
@@ -0,0 +1,3017 @@
...
+
+class SquadService extends Actor {
+  import SquadService._
+
+  /**
+    * The current unique squad identifier, to be wrapped in a `PlanetSideGUID` object later.
+    * The count always starts at 1, even when reset.
+    * A squad of `PlanetSideGUID(0)` indicates both a nonexistent squad and the default no-squad for clients.
+    */
+  private var sid : Int = 1
```
pschord: Do we have a strategy for when the squad service inevitably crashes and these SIDs reset to zero and new squads start to clobber old ones on the clients?

common/src/main/scala/services/teamwork/SquadService.scala
```diff
@@ -0,0 +1,3017 @@
...
+  /**
+    * All squads.<br>
+    * key - squad unique number; value - the squad wrapped around its attributes object
+    */
+  private var squadFeatures : TrieMap[PlanetSideGUID, SquadFeatures] = new TrieMap[PlanetSideGUID, SquadFeatures]()
```
pschord: hell yeah, Tries for speed

common/src/main/scala/services/teamwork/SquadService.scala
```diff
@@ -0,0 +1,3017 @@
...
+                }
+
+              //the following action can be performed by anyone who has tried to join a squad
+              case (_, CancelSelectRoleForYourself(_)) =>
+                val cancellingPlayer = tplayer.CharId
+                GetSquad(guid) match {
+                  case Some(squad) =>
+                    //assumption: a player who is cancelling will rarely end up with their invite queued
+                    val leaderCharId = squad.Leader.CharId
+                    //clean up any active RequestRole invite entry where we a
```
pschord: default case or all handled?

pschord: should we log it?

common/src/main/scala/services/teamwork/SquadService.scala
```diff
@@ -0,0 +1,3017 @@
+// Copyright (c) 2019 PSForever
```
pschord: I wont nitpick this file, plus it's hard to review given the size. Nice job!

common/src/main/scala/net/psforever/objects/teamwork/Member.scala
```diff
@@ -0,0 +1,89 @@
...
+  def Requirements : Set[CertificationType.Value] = requirements
+
+  def Requirements_=(req : Set[CertificationType.Value]) : Set[CertificationType.Value] = {
+    requirements = req
+    Requirements
+  }
+
+  def Name : String = name
+
+  def Name_=(moniker : String) : String = {
```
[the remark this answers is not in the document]
Fate-JH: This OOP accessor / mutator pattern is to Scala what getFoo / setFoo is to Java.  You could interact with a number of fields directly and ax the pattern, I guess.

common/src/main/scala/net/psforever/objects/teamwork/SquadFeatures.scala
```diff
@@ -0,0 +1,154 @@
...
+import net.psforever.objects.DefaultCancellable
+import services.teamwork.SquadService.WaypointData
+import services.teamwork.SquadSwitchboard
+
+class SquadFeatures(val Squad : Squad) {
+  /**
+    * `initialAssociation` per squad is similar to "Does this squad want to recruit members?"
+    * The squad does not have to be flagged.
+    * Dispatches an `AssociateWithSquad` `SDAM` to the squad leader and ???
+    * and then a `SDDUM` that includes at least the squad owner name and char id.
```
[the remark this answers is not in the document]
Fate-JH: I would have added one to the wiki if I could.
On the other hand, better practice is to always refer to the packets by name in the comments.

common/src/main/scala/net/psforever/objects/teamwork/SquadFeatures.scala
```diff
@@ -0,0 +1,154 @@
...
+  private var autoApproveInvitationRequests : Boolean = true
+  private var locationFollowsSquadLead : Boolean = true
+
+  private var listed : Boolean = false
+
+  private lazy val channel : String = s"${Squad.Faction}-Squad${Squad.GUID.guid}"
+
+  def Start(implicit context : ActorContext) : SquadFeatures = {
+    switchboard = context.actorOf(Props[SquadSwitchboard], s"squad${Squad.GUID.guid}")
+    waypoints = Array.fill[WaypointData](5)(new WaypointData())
```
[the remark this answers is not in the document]
Fate-JH: That's the number of squad-related way points I can manipulate at the moment.

common/src/main/scala/services/teamwork/SquadService.scala
Fate-JH: Fix the `SquadService` so that it doesn't crash.
When this actor does stop, though, it does try to close all squads that were active at the time (see `postStop`).

common/src/main/scala/services/teamwork/SquadService.scala
Fate-JH: I do plan to break it up.  This is just the only way I could deal with the initial setup.

common/src/main/scala/services/teamwork/SquadService.scala
Fate-JH: Indeed we should!  However, we need a better log standardization plan.  Many things that should be `trace` or `debug` are currently `info` and we need to rein in what does get logged.  The live server has already run into hassle with the size and number of the logs it produces.

Fate-JH: All should be handled.  Basically, one success short-circuits other checks, and then pools into the response for any success.  I could rewrite it without this flow if for clarity.

common/src/main/scala/net/psforever/objects/teamwork/Member.scala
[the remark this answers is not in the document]
pschord: Reading up on the scala style guidelines shows that this is the case! TIL

common/src/main/scala/services/teamwork/SquadService.scala
pschord: No I was just double checking

pschord: I wonder if we need to have more fine grained logging targets instead of the generic log severity. maybe we need to be able to select which log buffers will display to the screen. for instance, services each have their own buffer. Not something we can fix in this patch, but more thought is needed later

common/src/main/scala/services/teamwork/SquadService.scala
pschord: That's okay. when we come up with a plan to breakup WSA, that can propagate to the monolithic services.

common/src/main/scala/services/teamwork/SquadService.scala
pschord: Gotcha. Akka offers some nice redundancy / error recovery mechanisms. We should look into those cases later on

common/src/main/scala/net/psforever/packet/game/SquadWaypointRequest.scala
Fate-JH: I have an ancient "put-on-hold" ticket related to this that is me griping that we have continents that are referenced by string name (internal name) and referenced by their ordinal number (sometimes an `Int` and sometimes `PlanetSideZoneID`) and referenced by their map name and the only thing that connects them is the heavy `Zone` object,  I have at least two functions covering for conversions between these references at a higher level.  It'd be easier to just put together an enum-ish container.

common/src/main/scala/net/psforever/types/SquadRequestType.scala
Fate-JH: Since I sidestepped this idea at some point, I can probably strip this code out.  It's like this:

When I first inheritted this code from aphedox, he used a single enum for both `Request` and `Response`, though he tied them to the former (as `Request` types).  When I noted that the `Response` packets produced messages in the events chat, I decided to add them to the comments so it would be eaiser to remember when to use them.  At some point, I noticed that for certain enum values the message for value A would be better suited for value A+1 and that's when I decided to create two enumerations.  This code would just make `Request.Invite` translateable to `Response.Invite` if one had a different id than the other.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: This is part of my "dev branch stuffs," i.e., a local hack.  41605314 was the unique character identifier number for the original character we had encoded.  The first `sessionId` is typically 1.  On the live server, this whole identifier is a table row number.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: I'd like `PlanetsideAttributeMessage` to make sense myself!

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -113,6 +118,19 @@ class WorldSessionActor extends Actor with MDCContextAware {
...
+  var squad_supplement_id : Int = 0
+  /**
+    * `AvatarConverter` can only rely on the `Avatar`-local Looking For Squad variable.
+    * When joining or creating a squad, the original state of the avatar's local LFS variable is blanked.
+    * This `WSA`-local variable is then used to indicate the ongoing state of the LFS UI component,
+    * now called "Looking for Squad Member."
+    * Upon leaving or disbanding a squad, this value is made false.
+    * Control switching between the `Avatar`-local and the `WSA`-local variable is contingent on `squadUI` being populated.
+    */
+  var lfs : Boolean = false
```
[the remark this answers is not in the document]
Fate-JH: Going to change it to LFSM and update the comments.

common/src/main/scala/net/psforever/packet/game/SquadDefinitionActionMessage.scala
```diff
@@ -84,309 +414,65 @@ final case class SquadDefinitionActionMessage(action : Int,
...
+      case 26 => resetAllCodec
+      case 28 => autoApproveInvitationRequestsCodec
+      case 31 => locationFollowsSquadLeadCodec
+      case 34 => searchForSquadsWithParticularRoleCodec
+      case 35 => cancelSquadSearchCodec
+      case 38 => assignSquadMemberToRoleCodec
+      case 39 => noSquadSearchResultsCodec
+      case 40 => findLfsSoldiersForRoleCodec
+      case 41 => cancelFindCodec
+      case  2 |  6 | 11 | 12 | 13 |
```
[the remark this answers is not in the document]
Fate-JH: Mainly I don't know what to call some of them or what exactly they do.  All of them are encountered, in the packets, infrequently and others more frequently.  Most of the UI discoveries were subsequently named.

`Unknown(6)`, for example, is used at the start of the client's game, that, is, before loading squad definitions loadouts, but I don't know what to call it.
`Unknown(18)` is called at the end of many squad operations, making it an unconfirmed "completion packet," but I don't know what to call it.
The majority are called out entirely for the benefit of the packet converter, and the failure condition is intended to indicate (for me) that I've encountered something new.

common/src/main/scala/net/psforever/packet/game/SquadState.scala
Fate-JH: If it becomes a packet I revisit a lot in code, I just write a constructor that inserts those values in between the other populated fields.

common/src/main/scala/services/teamwork/SquadSwitchboard.scala
Fate-JH: We'll think about when `SquadSwitchboard` actually does something other than manage the periodic request-invite workaround.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: Done.

common/src/main/scala/net/psforever/packet/game/SquadInvitationRequestMessage.scala
```diff
@@ -0,0 +1,41 @@
...
+  * A message for communicating squad invitation.
+  * When received by a client, the event message "You have invited `name` to join your squad" is produced
+  * and a `SquadMembershipRequest` packet of type `Invite`
+  * using `char_id` as the optional unique character identifier field is dispatched to the server.
+  * The message is equivalent to a dispatched packet of type `SquadMembershipResponse`
+  * with an `Invite` event with the referral field set to `true`.
+  * @see `SquadMembershipResponse`
+  * @param squad_guid the squad's GUID
+  * @param slot a potentially valid slot index;
+  *             0-9; higher numbers produce no response
```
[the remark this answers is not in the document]
Fate-JH: Asserts, done.

common/src/main/scala/net/psforever/packet/game/SquadDetailDefinitionUpdateMessage.scala
```diff
@@ -0,0 +1,1451 @@
...
+            Attempt.Successful(guid :: true :: occupiedSquadFieldCount :: info :: HNil)
+          }
+          else {
+            //full squad detail definition protocol
+            Attempt.Successful(guid :: true :: 9 :: info :: HNil)
+          }
+      }
+    )
+  }
+}
```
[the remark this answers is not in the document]
Fate-JH: Four times.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: I need to pass the `WorldSessionActor.squadUI` `LongMap` object?

common/src/main/scala/net/psforever/types/SquadRequestType.scala
Fate-JH: Removed.

common/src/main/scala/net/psforever/types/SquadResponseType.scala
Fate-JH: Removed.

common/src/main/scala/net/psforever/packet/game/SquadDetailDefinitionUpdateMessage.scala
```diff
@@ -0,0 +1,1451 @@
...
+    requirements.orElse(Some(Set.empty)),
+    char_id.orElse(Some(0L)),
+    name.orElse(Some(""))
+  )
+}
+
+/**
+  * A container for squad position field data
+  * associating what would be the ordinal position of that field data in full squad data.
+  * @param index the index for this squad position;
```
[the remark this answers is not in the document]
Fate-JH: Done.

common/src/main/scala/net/psforever/objects/teamwork/SquadFeatures.scala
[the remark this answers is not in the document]
Fate-JH: Now influenced by `SquadWaypoints` enum.

common/src/main/scala/net/psforever/packet/game/SquadWaypointRequest.scala
Fate-JH: `SquadWaypoints` enum.

pslogin/src/main/scala/WorldSessionActor.scala
pschord: we dont need to worry about it for this patch, but these kind of functions are good to move out in the future. how we go about that, still needs thinking

common/src/main/scala/net/psforever/packet/game/SquadDetailDefinitionUpdateMessage.scala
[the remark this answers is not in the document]
pschord: 😨

pslogin/src/main/scala/WorldSessionActor.scala
pschord: Yikes. I see. Can you add a comment about that or replace with a whole number to make it more clear?

common/src/main/scala/net/psforever/packet/game/SquadDefinitionActionMessage.scala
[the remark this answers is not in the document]
pschord: Understood, my comment was mostly about duplicating the `unknownCodec(code)` inline with the rest of the cases just to avoid the large `N | N | N ... | N` syntax

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: Since we're working to unify development and live, this sort of hack (in fact, all sorts of hack) will be completely retired eventually.  For now, I'll add the comment.

common/src/main/scala/services/teamwork/SquadAction.scala
Fate-JH: Maybe we can seal them by putting classes into files with other classes?  I was using `FooAction.Action` and `FooResponse.Response` as `FooServiceMessage` and `FooServiceResponse` message parameter types, respectively.  While I sometimes might feel that this is not a good thing (for one reason or another), I believe that's why I chose to not seal them as-is.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: The comment above that line has been expanded.

common/src/main/scala/services/teamwork/SquadAction.scala
Fate-JH: Sealed.

common/src/main/scala/services/teamwork/SquadResponse.scala
Fate-JH: Sealed.

common/src/main/scala/net/psforever/packet/game/SquadDefinitionActionMessage.scala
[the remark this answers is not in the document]
Fate-JH: The cases have been broken up.

Fate-JH:
Addressing the change requests.

PR #283, opened 2019-10-16
Create PSForever config framework

pschord:
We can now parse arbitrary INI configuration entries. This will allow
server customization and testing without recompiling the server.


SouNourS:
> This will allow server customization and testing without recompiling the server.

And maybe without relaunch the server ?

pschord:
> > This will allow server customization and testing without recompiling the server.
> 
> And maybe without relaunch the server ?

Not at the moment, we can add this on a new patch. This would be nice to do from an API, but the server doesnt have this yet 😬

PR #286, opened 2019-10-23
Select Changes from My Original Securities Branch

Fate-JH:
Select changes from my original securities branch.

Features:
1. `AvatarAction` messages moved into `AvatarServiceMessage` to seal the base trait.
2. `AvatarResponse` messages moved into `AvatarServiceResponse` to seal the base trait.
3. Expanded `Avatar*.PlayerState(...)` messages to include important fields individually instead of passing the original packet around.
4. Added imports to remove specific warnings about implicit conversions (two times).
5. Sequence time assertion in `PlayerStateMessageUpstream` packet.
6. `WorldSessionActor` case for handling `ControlSync` -> `ControlSyncResp` flow has been modified. The "server time" field now should maintain a more consistent incremental value when constraining a 64-bit `Long` to a 32-bit `Long`.  The "time difference" field matches behavior observed in packet captures by increasing by one of the original received value when returned.
7. All players are explicily disallowed from generating projectile objects if that specific player is considered dead to the server at the time.  Counted infractions will be logged in a message to be displayed upon player death.


Mazo:
👍

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -149,6 +150,18 @@ class WorldSessionActor extends Actor with MDCContextAware {
 
   var timeDL : Long = 0
   var timeSurge : Long = 0
+  lazy val unsignedIntMaxValue : Long = Int.MaxValue.toLong * 2L + 1L
+  val sysTime : ()=>Long = {
+    //if the first nanoTime call is negative, subsequent calls will be only "slightly less negative"
+    //even with math.abs, each call to nanoTime would result in time counting down
+    if(System.nanoTime > 0) {
+      System.nanoTime
+    }
```
pschord: If nanotime is going to be negative, is there any reason we cant just use millis or does this require more testing?

Fate-JH: In regards to testing, we're not certain what the field calculates to begin with.  I'm just trying to accommodate the original logic under the assumption that, even if it failed, we never saw reprecussions of that failure.  At the same time, I am trying to bring it closer to behavior observed in examined packet captures.

If nanoTime produces one negative number, it will probably keep producing negative numbers for the extent of this run time.

pschord: Ah I see. When it is negative, the numbers count down. Can you return `- System.nanotime` for the negative case to make it so we're always using nanotime?

Fate-JH: After thinking about an appropriate rendition for the "if negative values" part, I came to the conclusion that `currentTimeMillis` should be more than apprioriate.

PR #294, opened 2019-12-02
Standardize constructors to have position as the first param and add positions to terminals

Mazo:


common/src/main/scala/net/psforever/objects/serverobject/terminals/Terminal.scala
```diff
@@ -180,12 +181,13 @@ object Terminal {
...
     * @param tdef the `ObjectDefinition` that constructs this object and maintains some of its immutable fields
     * @param id the unique id that will be assigned to this entity
     * @param context a context to allow the object to properly set up `ActorSystem` functionality
     * @return the `Terminal` object
     */
-  def Constructor(tdef : TerminalDefinition)(id : Int, context : ActorContext) : Terminal = {
+  def Constructor(pos: Vector3, tdef : TerminalDefinition)(id : Int, context : ActorContext) : Terminal = {
     import akka.actor.Props
 
     val obj = Terminal(tdef)
```
Fate-JH: Generic terminal object never gets assigned the position vector.

Mazo: Oops. Nice catch. Fixed.

PR #295, opened 2019-12-05
Zone-local Event Systems

Fate-JH:
Each zone now possesses each of three of its own event systems, one for avatar (player) events, one for local (environment) events, and one for vehicle (vehicle and mountable) events.  The purpose of event systems, besides communication with targets who are intended to share their experiences, is to pipe sequential activities.  Messages about event "b" for player "p" will only ever arrive after messages of event "a" for player "p," if "a" is essential for getting to "b" properly.  Although this represents a scaled increase in backend objects, the division performs a three-fold optimization.

Firstly, all events from one zone only go to that one zone.  This removes strain from multiple continents being employed at once, a situation wherein the events of players on Esamir are waiting for the events of players on Forseral are waiting on the events of players on TR Sanctuary.  Now mix in the different factions, who are all waiting on each other.  That is how the global version of these event systems worked.  Depending on how the `Actor` objects behave when it comes to timesharing processing, and managing automated scheduled events, breaking this system up by zones can be a significant difference, especially when paired with (future) computation-heavy calculations.

Secondly, the validation of zone identity is also reduced.  This makes message passing less complicated by removing the need for an extra parameter per message in many cases and the burden of constantly validating that zone identity.  In some cases, that latter issue may be unavoidable, but it is now no longer a required unavoidable situation.

Thirdly, this permits entities that are allowed to know to which zone they inhabit to directly interact with these event systems.  The previous logic of event message passing frequently made an extra, risky bounce back to the player in his `WorldSessionActor` object so that the player may then interact with the event systems and direct follow-up messages in lieu of the object that was really acting, e.g., a door changing state, etc..  Pointless paths were created just to manage this protocol, utilizing a message's `sender` as a long-term callback.  For entities that much more frequently interacted with the players, this was extra risky because, if the callback stopped being valid, it would risk the particular aspect of the system to crash upon trying to send a message to (essentially) `null`.  While sending to "all players _except a particular one_" is still a non-trivial task, sending messages to specific players as if through a unique channel just for that player that may or may not exist is easily achieved.  In short, game objects on the server are now allowed a level of triggered autonous activity.

What is this really, _really_ all in aid of?  The future dissolution of `WorldSessionActor`, bloated as it is now.

___Features___
__`AvatarService`, `LocalService`, `VehicleService`__
These service `Actor` objects now monopolize messages related to a specific zone.  The string identifier for the zone is still used as a channel name, and channel names for the messaging output are still mainly reliant on the messaging input, but most messages that explicitly report `Zone` objects or `zone_id` strings as parameters now acquiesce to the service's defined zone.

__`GalaxyService`__
A portion of the vehicle gating messaging protocol had to be transfered from the `VehicleService` object due to the latter's shift from being global to having zone locality.

__`Zone`__
Implicitly creates an `AvatarService` object, a `LocalService` object, and a `VehicleService` object that is exclusive to that zone.  Handles allow the user to send messages to those services from wherever the zone is publicly accessible.

__`PSLogin`__
The aforementioned service actors - `AvatarService`, `LocalService`, `VehicleService` - are no longer available globally through the `ServiceManager` object.  Consequently, and due to reasons of localization, `VehicleService` as an object is no longer explicitly implanted onto the `Zone` instances.  `Zone.Nowhere` is allowed to properly initialize.

__`Zone.Nowhere`__
This fake "place" is initialized to mainly support its role as the default no-zone zone in regards to interactions with the aforementioned three event system `Actor` objects. `WorldSessionActor` will always start pointed at `Nowhere` and will then switch to a proper zone.  In the process, the normal unsubscription actions will occur - the `Service.Leave` messages dispatched to the zone-local event systems - before the correct zone is subscribed - the `Service.Join` messages.

__`WorldSessionActor`__
All references to the global service actors have been removed, as per the change to `PSLogin`, and changed to the zone local service actors.  The three common channels per zone-local actors are, in order of increasing distribution scope, "player name," "faction," and "continent id."

__`AmenityOwner`__
Replaced the implicit helper object for type restricting the `Owner` field  of `Amenity` objects with a simple trait.  This affects `Building` objects and `Vehicle` objects, currently.  The real justification for this change is to allow for a manner of parent-based messaging that ignores the question of what the entity actually is and whether the entity can refer a `Zone` field to get access to event systems.

__`PainboxControl`, `ProximityTerminalControl`, `ResourceSiloControl`, `VehicleSpawnControl` (process)__
Converted to handle the new zone-local service `Actor` model.

___Caveats___
1. Many messages to the now-localized service actors, expecially those belonging to the `SupportActor` objects, e.g., `DoorCloserActor`, continue to superfluously require the `Zone` object be explicitly defined for framing autonomous event behavior.  In the future, this should be rendered unnecessary.
2. Proximity unit control still dispatches messages directly to the actor that would benefit from them, e.g., `WorldSessionActor` instances.  Even though this would be deprecated behavior as per the above concerns, the tight management of proximity unit behavior reduces the urgency of forcing compliance.  The changes that would be required to enforce this standard are obvious.
3.  I did eliminate the "sender as callback" format for message acceptance into `ProximityTerminalControl` at one point; but, I also decided to restore that format later (see previous caveat).  Despite that, both lines seem to act like they have changed.  Whitespace?
4. The more involved `Actor` tests remain finicky.  The `VehicleSpawnControl` tests, in particular, were all set to a significantly inflated timeout value just to increase the chances that they pass within any realistic test time frame.  The second time they are run, at least.

___Addenda___
1. I did significant work on the ever-unhappy `VehicleSpawnPad` spawning process.
- All references to the `AutoGuideControl` system has been eliminated as unnecessary.  It was never properly implemented in its current form and was never satisfactory in execution.  The primary cases where it was thought to be necessary (Interlink and Dropship Center) handle their own spawn routing.
- Possibility of a railless spawn has been eliminated.  This was always just a prototype system until map information could be extracted.
- The big change to all elements of this process involves using the zone-local event system rather than the global event system.  It will not be repeated for each file.
- Communication with specific players was changed from using a direct `Actor` callback to accessing the subscription of a specific player through their name via the pad's zone event system hook.  This eliminates explosions due to outdated messaging protocol being referenced but does mean that messages have to have complementary `VehicleAction` messages to reach the intended destinations.
- The queueing system uses fewer conditional checks in exchange for function literals and a `try ... catch ...` block.  Rather than consuming multiple orders by the same person and eliminating the extra ones during run-time, it handles these extraneous orders at insertion-time.
- If the vehicle spawn process gets past the `ConcealPlayer` stage, then the vehicle will be spawned and presented on the spawn pad regardless of what else happens.  Before this, multiple checks until the player was seated would allow the vehicle to be cleaned up early and the orders cycled.
- The `TryMount` request has been moved to one of the specific player calls rather than administered through the spawn pad control object.  If the local `TryMount` message fails to allow the customer to become the driver, that player is issued a `PlayerReveal` message that actually does as is intended this time.
- In general, the part of the process dealing with seating the driver is much more straightforward than previously attempts.  The order is received, the elements of the order are evaluated, an attempt is made, the process waits a bit then moves on to the next step regardless of the outcome.
- If the customer fails to become the driver, the vehicle will wait on the spawn pad for ~30s before being cleaned up and the next order processed.  If any player takes the vehicle's driver seat, the deconstruction will cancel, but the pad will continue to complain about being blocked and the next order will wait until the vehicle is moved far enough out of the way.
2. Logging levels have been reduced, many from `info` to `trace` level.


Mazo:
👍

Mazo:
@Fate-JH Merge conflicts from merging remote projectiles branch ☹️

PR #296, opened 2019-12-06
Basic lattice functionality

Mazo:
There is no particular functionality attached to this implementation of the lattice that is visible from an end user perspective, it's mostly just groundwork for future lattice related features (e.g. LLUs).

Zones hold a list of lattice links, and buildings have functions added to find their neighbours, or neighbours matching a specified faction.

Most of the changes in files are to allow buildings to have an actual name attached instead of just a GUID + MapID to allow the MapGenerator project to generate the lattice links easier, as the lattice links are stored in game_objects.adb in the following format:

`add_property map06 building_link_11 Anguta Sedna`

Pathing between nodes that are not direct neighbours is easy to do with the graph library used, however no function has been provided for this yet. A very rough example would be something akin to the below:

```scala  
val sg = continent.Lattice filter ((b: Building) => b.Faction == player.Faction)
val path = sg get continent.Building("Igaluk").get pathTo (sg get continent.Building("Akna").get)
```


Fate-JH:
Now that it's all merged (and conflicts handled through the web console), I'm going to have to go back and make certain it all still works correctly.

PR #297, opened 2019-12-09
MAX Capacitors

Mazo:
This one shouldn't need too much explanation. All three empire's MAXes will have functional capacitors.

The damage resolution for infantry has also been changed to allow for the NC max shield to absorb damage based on remaining capacitor, with the rest bleeding through to armour/health as necessary

The only caveat I can think of is the `capacitor_shield_override_time` property hasn't been implemented as I'm not sure what the original functionality was. I can guess that it might have been to stop the shield toggling on/off quickly but verification from other sources would be required.


Mazo:
Ah yes, I should have also mentioned that the isMoving change from my PR in #296 will be required first, unless the changes from #291 has fixed it already

PR #304, opened 2019-12-16
Fix VNLWorldStatusMessage for multiple worlds

pschord:
Having multiple worlds will enable a ton of great things! The fix ended up requiring a new codec as the main issue was that `encodedString` in the PlanetSide client will perform a BYTE alignment AFTER the size byte is decoded. Normally if the stream is byte aligned already all is well, but at the end of the WorldInformation structure is PlanetSideEmpire, which is ONLY 2 bits. This causes the next world in the vectorOfN to be encoded/decoded misaligned, which causes the client to crash upon reception.


Mazo:
Looks good to me, but codecs aren't my area of expertise. 👍

Fate-JH:
Would you write a test for a third world, that is an "ABCDABCD3"?

pschord:
image

PR #306, opened 2019-12-21
Increase SessionReaper timeouts and add to config file

pschord:
This should fix issues disconnecting at loading screens/zone changes as
no packets are being transmitted during this window. If the
WorldSessionsActor is also slightly overloaded, the session reaper can
drop the session mistakenly due to no outbound traffic.

Also fix-up WorldConfig.Get with better error messages along with more
tests.


Mazo:
👍

PR #309, opened 2019-12-31
Jammering

Fate-JH:
So much buzz buzz, you'd think you worked in an apiary.

This update only implements jammering **projectiles** - only the basic jammer grenades and the jammer grenade cartridges for the Punisher and for the Thumper.  The Sunderer-type (`apc`) vehicle special effect, the CUD (`command_detonater`) commander effect, and the Disruptor Mine (`jammer_mine`) effect are not implemented.

The only affected targets are: infantry (their implants and stamina), vehicles (their weapons), Boomers (explode), Disruptor Mines (fade), HE Mines (explode), sensor deployables (animate), deployed and fixed turrets, and, technically, the Aegis Shield Generator.  _All_ weapons are jammerable only for the purpose of implementing any of the previously-indicated logic.

___Features___
__`EffectTarget`__, __`TargetValidation`__
Formerly, a target designation and validation operation in service of proximity units called `ProximityTargets`.  Now, it has been rendered generic for use with projectiles.  The basic structure of a rough classification object and a clarifying validation function that accomodates `Any=>Boolean` has been retained.

__`ProjectileDefinition`__
Projectiles now have fields that flag when that type performs jammering, including target-specific fields that list the duration of the jammering effect.

__`JammableUnit`__
A `trait` that propagates fields to indicate whether the target is jammed or not.  All jammable objects are to inherit this `trait`.

__`JammableBehavior`__
A `trait` mix-in to be used by entity control objects (mainly), to mainage the two aspects of being jammered independently - the droning buzz sound effect and the status changes that accompany being jammered.  The sound flag is internal to this mix-in; see `JammableUnit` for the status flag.  The `trait` manages timers to indicate the duration of the aforementioned aspects, primarily.

__`JammableMountedWeapons`__
A specific form of the previous mix-in in service of `Vehicle` entities, `FacilityTurret` entities, and `TurretDeployable` entities.

__[object]`Deployable`__, __[object]`DeployableDefinition`__, __[object]`Control`__
This class pattern manages the structure and operation of almost all deployable objects through their promotion into "complex deployable" objects.  Only TRAPs and Telepads remain as "simple deployables" while "sensors" and "explosives" have been promoted.  Previously, certain simple deployables were supported by more-common definitions and might have not possessed a unique controlling object.  Unlike some arrangements, for most objects that have been promoted to "complex" just recently, all three of these classes inhabit a single file as opposed each having their own file.

__`PacketConverter`__
All objects that support being jammered will represent being jammered in their object create protocol.  In the case of the character select screen, however, this behavior has to be restricted.

__`WorldSessionActor`__
Stamina regeneration is now paced and occasionally punished by being able to skip "turns" once in a while.  Skipping merely means that one stamina is not restored during that period.  These turns are the number of `PlayerStateMessageUpstream` packets receieved from the client by the server, i.e., about four per second, excluding latency.

___Caveats___
1. While the ADB objects file has some hints in regards to which things are those knowable jammering sources, e.g., for this update, grenade-types, nothing indicates how the jammering affects those given units.  The "how" is missing from the record, save for personal experience and video evidence, and it is probably hardcoded.
2. Jammering entries contain shield generator references but nothing seems to happen to the Aegis Shield Generator when it is hit by a jammer grenade and packets pertaining to the effect.  The jammering is still supported, even though nothing seems to come of it.  (Perhaps it affects the physical shield upgrade for the Aegis?)  To ensure that unnecessary code paths are not traversed, the `GlobalDefinitions` mentions of the shield generator jammering conditions have been excluded from the affected projectiles.
3. When sensor-type deployables - the motion alarm sensor and the sensor disruptor - are struck with a jammer effect, their radar animations will cease for the duration of the effect.  I do not know whether this is vanilla behavior and packet examination is unreliable for confirmation.  "It looks nice."  I'll go with that until evidence is provided.  Also, since there really is no concept of "radar" right now, deploying or jammering a sensor deployable is merely for cosmetics and momentarily blocking hallways.
4. When the Spitfire turret-type deployable is hit with a jammer grenade, its health bar seems to decrease a very small amount.  This is an odd visual glitch.  It starts with 100 health and still has 100 health after being jammered.
5. By way of extrapolation, videos on Discord suggest that jamming effect time and jamming sound time are the same for all targets except for Infantry.  The effects of Infantry jamming are instantaneous and the wind down due to natural mechanism (implant initialization timer); the sound lingers.  In the case of Infantry timing, however, evidence does not point to any specific duration for the noise.  The quality of the evidence is to blame in cases - weapon discharge is much louder than the jammering buzz, and some weapons have discharge sounds that intermingle well with the buzz - but other mechanisms that are not explained in the ADB may be at work.
6. Implants are not fully functional right now and are only two are functional on Play-Live.  They start as properly initialized and will remain that way.  To support even these options, those implants are deactivated.  As compensation for the lack of initialization system, a ten turn delay when it comes to stamina regenetation has been levvied upon a jammered infantry target.
7. Activity entries for jammering do not exist, but will be added in the future.
8. All `Tool` objects are jammable.  This quality may be extended to basic `Equipment` in the future as may be required.
9. Some messaging between entity-local control objects and the zone-local event systems, jammering protocol included, is not generic and exposes/emits implementation details.
A. The ADB mentions a type of projectile for the jammer grenade cartridge called `rocklet_jammer_projectile`.  An entry for that projectile is defined elsewhere in the file, but the projectile itself is not the discharge of any weapon fire mode.

___Addenda___
1. A critical fix for remote projectile functionality has been included in this update.
2. Various logic for damage dealing and damage resolution previously divided between one or another common event object and callback path to manage synchronicity has been repurposed onto object-local control entities.  Only some remaining extremely simple deployables and Infantry units bounce damage-dealing authorities through the (appropriate) event systems to achieve synchronicity.  These efforts should be considered in-line with the inevitable goal of stripping responsibility out of `WorldSessionActor` and put them into the hands of relevant individual agencies; but, as per the purpose of this update, only jammering-affected targets had their code mitigated.
3. An inheritance conflict that previously restricted `TurretDeployable` objects from being considered fully-fledge `Deployable` objects, by definition, caused by inconvenient interaction with the inheritance path of `TurretDefinition`, has been resolved.  `FacilityTurret` objects were affected very little by this change and retain their proper identity.


common/src/main/scala/net/psforever/objects/ExplosiveDeployable.scala
```diff
@@ -13,4 +26,86 @@ class ExplosiveDeployable(cdef : DeployableDefinition) extends SimpleDeployable(
...
+    val zone = target.Zone
+    val playerGUID = zone.LivePlayers.find { p => cause.projectile.owner.Name.equals(p.Name) } match {
+      case Some(player) => player.GUID
+      case _ => PlanetSideGUID(0)
+    }
+    if(target.Health == 0) {
+      HandleDestructionAwareness(target, playerGUID, cause)
+    }
+    else if(!target.Jammed && cause.projectile.profile.JammerProjectile) {
+      if(target.Jammed = {
```
Mazo: Isn't this always going to be false due to the check on the line above?

common/src/main/scala/net/psforever/objects/ShieldGeneratorDeployable.scala
```diff
@@ -1,15 +1,121 @@
...
+  override def StartJammeredSound(target : Any, dur : Int) : Unit =  { }
+
+  override def StartJammeredStatus(target : Any, dur : Int) : Unit = target match {
+    case obj : PlanetSideServerObject with JammableUnit if !obj.Jammed =>
+      obj.Zone.VehicleEvents ! VehicleServiceMessage(obj.Zone.Id, VehicleAction.PlanetsideAttribute(Service.defaultPlayerGUID, obj.GUID, 27, 1))
+      super.StartJammeredStatus(obj, dur)
+    case _ => ;
+  }
+
+  override def CancelJammeredSound(target : Any) : Unit =  { }
```
Mazo: I assume this and `StartJammeredSound` are empty because it has no jammered sound, or is this an oversight?

If the former, perhaps a comment might be useful for anyone looking at it in the future.

common/src/main/scala/net/psforever/objects/equipment/EffectTarget.scala
```diff
@@ -0,0 +1,103 @@
...
+    def HealthCrystal(target : PlanetSideGameObject) : Boolean = target match {
+      case p : Player =>
+        p.Health > 0 && p.Health < p.MaxHealth
+      case _ =>
+        false
+    }
+
+    def RepairSilo(target : PlanetSideGameObject) : Boolean = target match {
+      case v : Vehicle =>
+        !GlobalDefinitions.isFlightVehicle(v.Definition) && v.Health > 0 && v.Health < v.MaxHealth
```
Mazo: Would these need faction checks, or is that applied elsewhere?

common/src/main/scala/net/psforever/objects/equipment/JammingUnit.scala
```diff
@@ -0,0 +1,271 @@
...
+
+object JammableMountedWeapons {
+  /**
+    * Retrieve all of the weapons on a `MountedWeapons` target object and apply a jammered status effect to each.
+    * @param target an object that can be affected by the jammered status
+    * @param statusCode the jammered status condition;
+    *                   0 for deactivation;
+    *                   1 for activation
+    */
+  def JammeredStatus(target : PlanetSideServerObject with MountedWeapons, statusCode : Int) : Unit = {
```
Mazo: Is statusCode not better as a Boolean, or are other statuses possible?

common/src/main/scala/net/psforever/objects/ExplosiveDeployable.scala
Fate-JH: `else if(!target.jammed && ...)` : this block only runs if we are not yet jammed.

`if(target.Jammed = { ... }` : set a value, the evaluation of the braced code, and return that value.  It may resolve to true and then we are jammed.

common/src/main/scala/net/psforever/objects/ShieldGeneratorDeployable.scala
Fate-JH: As mentioned in Caveat 2, the shield generator is called out in the ADB entries for jammer projectiles but it has no obvious responses to the conventional jammering packets being received.  It is also a single unit at this moment of implementation.

For example, when the vehicle is jammered, its weapons are also jammered.  Both elements emit the buzz sound but jammering the weapons is required to make them stop shooting.  The Aegis Shield Generator has no need to have any subcomponents being jammed because it has no subcomponents at the moment, hence, no reason to be jammered twice.  Overriding the `Start...` method stops the timer from being started; overriding the `Cancel...` is just for completion.  It didn't matter whether it was `Sound` or `Status`.

common/src/main/scala/net/psforever/objects/equipment/JammingUnit.scala
Fate-JH: `statusCode` gets fed into a `PAM`.  I see no reason to cast from `Boolean` when the value that is needed can directly be a `Long`.

If some reason to ever run an unscheduled `JammeredStatus(obj, obj.Jammed)` arises, I'll consider it useful.

common/src/main/scala/net/psforever/objects/equipment/EffectTarget.scala
Fate-JH: The only jammering that is faction-specific is the disruptor mine, yet to be implemented.

In terms of proximity terminals, the faction check should be in the `ProximityTerminalControl` object (though I don't think it is).

common/src/main/scala/net/psforever/objects/ExplosiveDeployable.scala
Mazo: Oops, I totally misread that as a comparison, not an assignment. As you were.

common/src/main/scala/net/psforever/objects/equipment/JammingUnit.scala
Mazo: 👍

common/src/main/scala/net/psforever/objects/equipment/EffectTarget.scala
Mazo: 👍

common/src/main/scala/net/psforever/objects/ShieldGeneratorDeployable.scala
Mazo: It might be worth adding a small explanation of that as a comment in that case, so for future reference we remember why it's overriden.

Fate-JH: Commented.

PR #310, opened 2020-01-01
GenericObjectActionMessage Field Change

Fate-JH:
A quick examination of comments in the documentation for this packet asserts that number ranges such as 60, 61, 62, 63 possess the same effect and the next number range 64, 65, 66, 67 all possess a same effect, different from the previous.  And so forth for 68, 69, 70, 71.  Based on that analysis, the numbers that matter are actually 60, 64, 68; divided by two twice they are 15, 16, 17; and, furthermore, that means the fidelity of the field extracted is too high by two bits.  This makes sense for all currently utilized data as they are all divisible by four.

Instead of a domain of possible `8u` numbers 0-255 where most of the purpose of the numbers are currently unknown or undefined, the amended domain is `6u` numbers 0-63 with only about half unattributed.


Mazo:
👍

PR #311, opened 2020-01-04
Painbox Opt-In

Fate-JH:
To quote my Trello ticket:
> All `Zone` objects have a `SphereOfInfluenceActor` object.  All `SphereOfInfluenceActor` objects have a timer that fires every 5s to partition which players, if any, are in the radius of which of the zone's `Building` `SphereOfInfluence` objects.  Every 1s, the `PainBox` objects that belong to each of a zone's `Building` objects tests which players in the radius of the `Building` object's `SphereOfInfluence` object is subject to the conditions of that `PainBox` object.  All of this system is firing even if a player is situated in VR Shooting, where there is no `SphereOfInfluence` object.  Each of the main continents will have nearly one hundred of these timers constantly running and firing.  The Battle Islands and Sanctuaries and Caverns probably add up to just over another one hundred together.  Meanwhile, the player is sitting in their faction's VR shooting gallery, where there are no `SphereOfInfluence` objects or `PainBox` objects, playing with loadouts.
> 
> This system could definitely be optimized.

Now, the server only runs the `PlayersInSOI` selection timer on the `Buildings` in a `Zone`  and only when the `Zone` actually has at least one `Player` in the `Zone`.  Meanwhile, `Painbox` elements only run their selection timer when the `Building` that owns them has a populated `PlayersInSOI` list.  In other words: no players, paused functionality.

Smaller changes include: a reinforcement of `/zone` placement rules; sorting out the swapped medical terminals on `home3`; moving around the subscriptions for the chat events system; and, painbox damage entries for vitals history.

___Caveats___
The `/zone` reinforcement is only tested for Infantry classes.  Using vehicles may result in the same issues with improper destination coordinates as before.


Mazo:
👍

PR #313, opened 2020-01-04
Vehicle fixes

Mazo:
Few small bug fixes. Sends vehicle shields/capacitor state to mounting player and allows players to dismount in a few edge cases where they could be stuck unable to dismount previously

### DIFF
--- a/pslogin/src/main/scala/WorldSessionActor.scala
+++ b/pslogin/src/main/scala/WorldSessionActor.scala
@@ -1781,8 +1781,8 @@ class WorldSessionActor extends Actor
         log.info(s"MountVehicleMsg: $player_guid mounts $obj_guid @ $seat_num")
         PlayerActionsToCancel()
         sendResponse(PlanetsideAttributeMessage(obj_guid, 0, obj.Health))
-        sendResponse(PlanetsideAttributeMessage(obj_guid, 68, 0)) //shield health
-        sendResponse(PlanetsideAttributeMessage(obj_guid, 113, 0)) //capacitor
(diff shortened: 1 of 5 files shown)


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -1781,8 +1781,8 @@ class WorldSessionActor extends Actor
         log.info(s"MountVehicleMsg: $player_guid mounts $obj_guid @ $seat_num")
         PlayerActionsToCancel()
         sendResponse(PlanetsideAttributeMessage(obj_guid, 0, obj.Health))
-        sendResponse(PlanetsideAttributeMessage(obj_guid, 68, 0)) //shield health
-        sendResponse(PlanetsideAttributeMessage(obj_guid, 113, 0)) //capacitor
+        sendResponse(PlanetsideAttributeMessage(obj_guid, 68, obj.Shields)) //shield health
+        sendResponse(PlanetsideAttributeMessage(obj_guid, 113, obj.Capacitor)) //capacitor
```
SouNourS: Maybe you can also add this : 
if(obj.Definition.ObjectId == 60){  // PTS v3
          sendResponse(PlanetsideAttributeMessage(obj_guid, 45, scala.math.ceil((obj.Capacitor.toFloat / obj.Definition.MaximumCapacitor.toFloat) * 10).toInt))
        }

Mazo: Hmm, perhaps we should be splitting out NTU capacitor from EMP capacitor, since they use different `PSAM` values

Mazo: Resolved

PR #314, opened 2020-01-05
Player Kills Game

Fate-JH:
When you die on the server, you die for real.


Fate-JH:
The player will eventually have its own control `Actor` object.  I started down this path by assigning the very first controlling object to be an instance of `WorldSessionActor` as part of a test for the concept.  When the player joins a zone, a personalized controlling object is created.  When the player releases after dying or deconstructing, that controlling object is supposed to be killed.   As `WSA` is still a functional object, however, it can not be overwritten by the actual player controlling object during the zone joining process.  I shouldn't have to draw any more of a line to explain why everything goes pear-shaped.

PR #317, opened 2020-01-06
Account and Character Database and Config Improvements

pschord:
I cherry picked the database management code from PTSv3 into master. Thanks to @SouNourS and @xTriad Jordan Nelson for their work developing this (PR #198). Due to the large amount of changes since the original PR, I had to go back in and untangle things. I tried to make a minimal amount of changes in order to keep track of the changes left from the PTSv3 branch.

* LoginSession: Loads accounts from database like PTSv3
* WorldSession: Contacts account intermediary to match login token from initial connection
* WorldSession: Loadouts and lockers are saved per character
* WorldSession: fixed crash when trying to remove boomers from deconstructed players without a GUID
* Config: Add database section
* Config: WorldServer hardcoded name/port moved to config
* Config: Now able to handle Enumeration
* Docs: Added Database setup guide to the README

### DIFF
--- a/pslogin/src/main/scala/LoginSessionActor.scala
+++ b/pslogin/src/main/scala/LoginSessionActor.scala
@@ -1,5 +1,6 @@
 // Copyright (c) 2017 PSForever
 import java.net.InetSocketAddress
+import java.net.InetAddress
 
 import akka.actor.{Actor, ActorRef, Cancellable, MDCContextAware}
 import net.psforever.packet.{PlanetSideGamePacket, _}
@@ -44,6 +45,16 @@ class LoginSessionActor extends Actor with MDCContextAware {
   var hostName : String = ""
   var canonicalHostName : String = ""
   var port : Int = 0
+
+  val serverName = WorldConfig.Get[String]("worldserver.ServerName");
+
+  // This MUST be an IP address. The client DOES NOT do name resolution 
+  var serverHost : String = if (WorldConfig.Get[String]("worldserver.Hostname") != "")
+    InetAddress.getByName(WorldConfig.Get[String]("worldserver.Hostname")).getHostAddress
+  else
+    LoginConfig.serverIpAddress.getHostAddress
+
+  val serverAddress = new InetSocketAddress(serverHost, WorldConfig.Get[Int]("worldserver.ListeningPort"))
 
   private val numBcryptPasses = 4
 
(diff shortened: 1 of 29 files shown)


Fate-JH:
In the README, the PostgreSQL download link for "Windows" points to the Ubuntu page.  It should probably point to https://www.postgresql.org/download/windows/?

pslogin/src/main/scala/LoginSessionActor.scala
```diff
@@ -25,9 +37,27 @@ class LoginSessionActor extends Actor with MDCContextAware {
...
+
+  // This MUST be an IP address. The client DOES NOT do name resolution 
+  var serverHost : String = if (WorldConfig.Get[String]("worldserver.Hostname") != "")
+    InetAddress.getByName(WorldConfig.Get[String]("worldserver.Hostname")).getHostAddress
+  else
+    LoginConfig.serverIpAddress.getHostAddress
+
+  val serverAddress = new InetSocketAddress(serverHost, WorldConfig.Get[Int]("worldserver.ListeningPort"))
+
+  private val numBcryptPasses = 4
```
Mazo: Isn't this way, way too low to avoid brute forcing by modern CPUs/GPUs?

pschord: Yes, it is. 10 would be better. https://stackoverflow.com/a/50470009 I was concerned with increasing it and it breaking the existing passwords, but looking at it more, it looks like it stores the number of rounds in the hash itself. I can increase it.

SouNourS: > 
> 
> Isn't this way, way too low to avoid brute forcing by modern CPUs/GPUs?

That brute force is non only if you have a copy ? Because with a DB not over internet + we can limite the number of try on login (3 login in 1 min => block), I'm not sure that's needed. No ? (I'm not a security expert)

pschord: The primary use is to slow down hash cracking in the event that the database is compromised. MD5 for example, can be run billions of times a second allowing for any passwords hashed with it to be cracked in a matter of seconds with GPUs. With a work-factor of 10 for bcrypt, if our DB is leaked, it will take ~100ms for each hash query. 10h/s is much better than 1000000000h/s ;)

SouNourS: Ok, in case of compromised :)

PR #318, opened 2020-01-07
Enable quiet tests using SBT config

pschord:
This should make TravisCI failing tests be *much* easier to identify as we only show the failing tests for both Specs2 and ScalaTest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ dist: trusty
 scala:
   - 2.11.8
 env:
-  - SBT_COMMAND="test:compile test packArchiveZip"
-  - SBT_COMMAND="coverage test:compile test coverageReport"
+  - SBT_COMMAND="test:compile quiet:test packArchiveZip"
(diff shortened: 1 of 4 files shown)


pschord:
Around a 62.6% reduction in lines
Before: https://travis-ci.com/psforever/PSF-LoginServer/jobs/273178198 (5,599 lines)
After: https://travis-ci.com/psforever/PSF-LoginServer/jobs/273320130 (2,089 lines, around ~1500 visible)

pschord:
Adding lockback 1,721 lines (69%)
Adding no resolve 1,488 (73.4%)
Adding missing option to scalatest 1,245 (77.6%)
Removing headers from Specs2 tests 1,007 (82%)

Nice! Much easier to figure out what went wrong. We should look into getting rid of the service actor spam (or fix this if it is a bug). One issue with this configuration is that without headers, it's slightly harder to see which Specs2 test failed

Mazo:
👍

PR #319, opened 2020-01-08
Object "Staleness"

Fate-JH:
After a long hard discussion with my rubber ducky, I have decided that unregistered objects should no longer lose the globally unique identifier they have been assigned by the server during registration.

Same as before, if the same unique identifier number is checked against the zone-wide collection of objects mapped to numbers, it will still not point to that entity anymore.  That is a basic requirement of unregistration that must remain fulfilled.  This is not a fair test to constantly have to make, however, so two methods of determining when an entity is outside of its expected scope have been instituted.  This degree of "staleness" is internalized by an entity itself in two ways.  The first is a pre-existing condition called "`HasGUID`" which merely flags whether a valid unique identifier number has ever been applied to the entity.  The second is an explicit distinction in "valid" and "stale" unique identifier containers that are expressed by the object when it is requested, the former for registration and the latter for the default condition and for unregistration,  By linking the two class/states of the unique dientifier in such a way that removes unnecessary distinction between the classed, all entities will always report an identifier number if one has been initially registered in such a way that it remains useful to the majority of the server.

The primary purpose of reporting like this is to create fields that can be filtered or pattern matched to isolate and detect invalid identifier numbers and invalid objects.  Currently, no functionality is implemented to take advantage of this safeguard.  The more short-term demonstration of this change should be a decline in zone loading failure, specifically the kind of issues that arise due to an entity going out of scope when such the entity's zone is loading for some avatar.  The window for a stale object to exist should be so small that it will realistically never run the possibility of overriding a valid object that has been re-assigned the unique identifier - the runtime of a completed functional call.

Most of the time, no change to server ops should be noticed due to this update.  Fewer crashes due to `NoGUIDException`, maybe.

___Features___
__`PlanetSideGUID`__
The class was moved to `types`, the directory where packet / object cross-over structures are expected to be found.  The class is also no longer just a `case class` but is an `abstract class` with two subclasses that form part of the backbone for staleness implementation - the `case class`es `ValidPlanetSideGUID` and `StalePlanetSideGUID`.  A great massaging of the classes is then applied to ensure that the code does not need massive reviews to deal with this change to such a fundamental building block.

__`IdentifiableEntity`__
The internal mechanisms of this class were completely rewritten, gutting out the old "hidden container" approach in exchange for a re-assigning function literal approach.  The basic impact and handling of the class is exactly the same, save for any changes made pertaining to a demonstration of "staleness", and the change to the fate of the unregistered globally unique identifier number.

__`NoGUIDException`__
~~While no changes were made to the exception itself,~~ the purpose of the exception is now one that raises concern when a globally unique identifier has never been set on an object.  This indicates only the initial registration is unfulfilled.

___Caveat___
The code went crazy when I moved `PlanetSideGUID` from `packet.game.CharacterInfoMessage` to `types`.  That's why 342 files have been changed!  Refactoring might be a powerful IDE tool but it's pretty rubbish when it comes to responsibly handling class import statements.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/Deployables.scala
+++ b/common/src/main/scala/net/psforever/objects/Deployables.scala
@@ -3,7 +3,8 @@ package net.psforever.objects
 
 import net.psforever.objects.ce.{Deployable, DeployedItem}
 import net.psforever.objects.serverobject.PlanetSideServerObject
-import net.psforever.packet.game.{DeployableInfo, DeploymentAction, PlanetSideGUID}
+import net.psforever.packet.game.{DeployableInfo, DeploymentAction}
+import net.psforever.types.PlanetSideGUID
 import services.RemoverActor
 import services.local.{LocalAction, LocalServiceMessage}
 
--- a/common/src/main/scala/net/psforever/objects/ExplosiveDeployable.scala
+++ b/common/src/main/scala/net/psforever/objects/ExplosiveDeployable.scala
@@ -9,8 +9,7 @@ import net.psforever.objects.definition.converter.SmallDeployableConverter
 import net.psforever.objects.equipment.JammableUnit
 import net.psforever.objects.serverobject.PlanetSideServerObject
 import net.psforever.objects.vital.{StandardResolutions, Vitality}
-import net.psforever.packet.game.PlanetSideGUID
-import net.psforever.types.Vector3
+import net.psforever.types.{PlanetSideGUID, Vector3}
 import services.avatar.{AvatarAction, AvatarServiceMessage}
 import services.local.{LocalAction, LocalServiceMessage}
 
--- a/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
+++ b/common/src/main/scala/net/psforever/objects/OwnableByPlayer.scala
@@ -1,7 +1,7 @@
 // Copyright (c) 2019 PSForever
 package net.psforever.objects
 
-import net.psforever.packet.game.PlanetSideGUID
+import net.psforever.types.PlanetSideGUID
 
 trait OwnableByPlayer {
   private var owner : Option[PlanetSideGUID] = None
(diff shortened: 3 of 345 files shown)


pschord:
Thanks for separating the refactoring commits from code level changes. I like the algebra we have with Stale vs Valid. Should help write more correct code when dealing with these types. One concern I have is the ability for a valid GUID to "eat" assignments when it's already assigned ("ignore subsequent mutations using a valid GUID" test). I feel that reassigning a GUID to an object is likely to always be a programmer error and that we should strictly enforce the immutability of GUIDs by preventing their reassignment (through removing the setter method and instead relying on constructor params only) or if that is infeasible, throwing something to hint to the devs that "this section of code is making wrong assumptions of GUID/object lifetimes".

Fate-JH:
The code path that intersects our common object registration process already checks `HasGUID` on objects before it acts on them and returns an unsatisfying "early success" in the case that it is (`\objects\guid\actor\UniqueNumberSystem.scala`).  The inability to set the GUID to something different without invalidating it mirrors being unable to register the object a second time without unregistering it.  Should I modify that condition as well?

pschord:
> The inability to set the GUID to something different without invalidating it mirrors being unable to register the object a second time without unregistering it. Should I modify that condition as well?
I personally think so! If the operation is a no op, then it shouldnt ever be performed. If it is performed, there is code somewhere accidentally (?) registering things twice. Maybe this will catch errors now or down the road when creating/registering new objects

Fate-JH:
For my own sanity, I'm not going to leverage the constructor parameter route.  I'd have to rewrite the constructor for every game object and create a new relationship for objects-by-themsleves and objects-with-guids, and then those changes would have to filter down to entirety of the rest of the server code.  All in all, significant retooling.  I felt was fine to do something like that for `Avatar` and the  `CharId` field since, no matter where it is or what zone it is in, `CharId` is both vital and immutable.

pschord:
Fair enough. Would be touching a lot of code that has been tested well. Up to you if you think it's worth adding a visible error when reassigning things. Could catch some mess ups down the line

Fate-JH:
Amended:
`GUIDException` is the base class for expressing exceptions related to GUID's.  The only two that exist at the moment are the classic `NoGUIDException` (which indicates access before registration) and `AssigningGUIDException` (stale assignments and redundant assignments).

pschord:
Looks great! Will be interesting to see if we catch any bugs

PR #320, opened 2020-01-10
Few more small vehicle related tweaks

Mazo:
Not a lot of explanation required here. The work @Fate-JH did with GUIDs means the previous workaround for cancelling jacking on deconstructed vehicles can use the vehicle GUID still, and don't allow dead vehicles to be revived.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/Vehicle.scala
+++ b/common/src/main/scala/net/psforever/objects/Vehicle.scala
@@ -145,7 +145,9 @@ class Vehicle(private val vehicleDef : VehicleDefinition) extends AmenityOwner
   }
 
   def Health_=(assignHealth : Int) : Int = {
-    health = math.min(math.max(0, assignHealth), MaxHealth)
+    if(health != 0) {
+      health = math.min(math.max(0, assignHealth), MaxHealth)
+    }
     health
   }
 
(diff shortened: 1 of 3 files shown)


common/src/main/scala/net/psforever/objects/Vehicle.scala
```diff
@@ -145,7 +145,9 @@ class Vehicle(private val vehicleDef : VehicleDefinition) extends AmenityOwner
   }
 
   def Health_=(assignHealth : Int) : Int = {
-    health = math.min(math.max(0, assignHealth), MaxHealth)
+    if(health != 0) {
```
pschord: Lets make an `isDead` bool instead (might need to check it elsewhere in the object).

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -4947,7 +4946,7 @@ class WorldSessionActor extends Actor
                 case GlobalDefinitions.nano_dispenser =>
                   //TODO repairing behavior
                   if (!player.isMoving && Vector3.Distance(player.Position, obj.Position) < 5) {
-                    if (obj.Health < obj.MaxHealth) {
+                    if (obj.Health < obj.MaxHealth && obj.Health > 0) {
```
pschord: Use `!isDead` here

Mazo: Done

common/src/main/scala/net/psforever/objects/Vehicle.scala
Mazo: Done

PR #321, opened 2020-01-10
Database Login Errors

Fate-JH:
Are these messages useful?

I would think so.  Without two specific database messages being visible (anywhere, let alone in the log), I would not have been able to determine how to properly GRANT privileges to progress through first account creation and then character creation.  One might argue, "at least only once, if even that;" but, that would mean they were still useful that one time.

### DIFF
--- a/pslogin/src/main/scala/LoginSessionActor.scala
+++ b/pslogin/src/main/scala/LoginSessionActor.scala
@@ -110,6 +110,9 @@ class LoginSessionActor extends Actor with MDCContextAware {
         val serverTick = Math.abs(System.nanoTime().toInt) // limit the size to prevent encoding error
         sendResponse(PacketCoding.CreateControlPacket(ControlSyncResp(diff, serverTick, fa, fb, fb, fa)))
 
+      case TeardownConnection(_) =>
+        sendResponse(DropSession(sessionId, "client requested session termination"))
+
       case default =>
         log.error(s"Unhandled ControlPacket $default")
     }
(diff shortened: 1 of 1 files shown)


pschord:
Im hesitant to add more verbose logging of exceptions to our auth flow...Im worried that we'll expose hashes/plaintext passwords in logs. Can you match on the specific exceptions from the library (like here https://github.com/psforever/PSF-LoginServer/blob/master/pslogin/src/main/scala/Database.scala#L40) and print a useful message instead of the whole thing

Fate-JH:
Would limiting the results to `SQLException` and `... .GenericDatabaseException` messages be considered sufficiently safe?

pschord:
I'm not sure. Maybe you can inject a syntax error into a query to see if it spits back the VALUES?

Fate-JH:
I tried breaking the syntax in the prepared statement of `createNewAccount` by making a typo in the table name.    I also tried to break the syntax for the same prepared statement by modifying one of the reserved words (RETURNING -> RETURN) in a separate test.  Both times, the error message did not display the values from `VALUE`, save for the account name, and that is already reported in the log.

Should we handle the `TeardownConnection` message? because we do get hit with a `TeardownConnection` that we don't do anything with from time to time.

pschord:
> I tried breaking the syntax in the prepared statement of `createNewAccount` by making a typo in the table name. I also tried to break the syntax for the same prepared statement by modifying one of the reserved words (RETURNING -> RETURN) in a separate test. Both times, the error message did not display the values from `VALUE`, save for the account name, and that is already reported in the log.

Great. LGTM 


> Should we handle the `TeardownConnection` message? because we do get hit with a `TeardownConnection` that we don't do anything with from time to time.

Yes might as well do `sendResponse(DropSession(sessionId, "client requested session termination"))` (sessionrouter API)

PR #328, opened 2020-01-13
Generator

Fate-JH:
It's a generator.

___Features___
Just Anguta's generator, for now.

___Caveats___
It does nothing.

___Addenda___
The generator terminal.  It also belongs to Anguta and does nothing.


Fate-JH:
I'm not writing tests until there's somerthing worth testing.

PR #346, opened 2020-03-31
Destroy and repair

Fate-JH:
VAG?
VVN, VDG.
VVS.

This update has too many layers to discuss, many of them fortunately declarative rather than imperative.  The core aspect is that facility amenities that previously didn't take damage can now take damage and all amenities that could be damaged and repaired can now be repaired. It's boring to say that in so few words and it breaks too much from the normal pull request format that would take far too long to tabulate,  A system overview, though it may be verbose and technical, seems in order.  Most people can probably skip down the the "Caveats" section and miss nothing.

___Features___
__Overview__
All game objects that can be damaged or repaired have  viltality (`Vitality`).  Vitality encompasses all forms of statistical changes that can be tracked.  The most basic vital statistic is, of course, health, and that determines one aspect of whether the target is organically considered destroyed or not.  The most important related object state is whether that object is `Destroyed` which is an affect on how the game object is treated by the server.  Other things treated as vital statistics include "armor", "stamina", and even "capacitor (energy)".  Refer to `ShieldGeneratorControl` for example how "shields" can be worked into a damage system that only normally involves "health".

All damage activities or repair activities are filtered from decoded packet data from one specific client to the server.  After processing and target reification, the result is dispatched internally to the control agency of the affected target and, after being sorted, then communicated to the rest of the system - the other connected clients - via messages sent to the server's event services.  A control agency is the what the game object uses to make internal changes to that object's state independent of the frame of reference of any given user.  It can be utilized to enforce synchronized changes that are perceived across mutliple users where necessary.  If no individual control agency is employed, one of the game's event-distribution services is suitable.  A record of the previous activity is then retained by the target.

The process of damaging or repairing is dependent on the agency's knowledge of what object it is meant to control and is generally structured as a series of steps that build upon each other.  Basic operations are treated as "entities" and perform only specific actions.  What we call "facility amenities", on the other hand, build on "entity" actions with extra actions when treated as "amenities."  These contextiual actions are so forth layered for some mountable targets and for vehicle targets.  Explosives, players, and telepads operate on separate rules.

__Damage__
Damage is controlled by the object general `damageable` property which allows for processing of `Vitality` properties.  Some other general properties are health-related value properties and others are condition-related properties.  `MaxHealth` is the value property that determines the ceiling for object health, for example.  `DamageByFriendlyFire` is the conditional property that allows damage originating from allied sources (accidental, perhaps).

All projectile-based damage originates with one of three packets being received - `HitMessage`, `SplashHitMessage`, or `LashMessage`.  The important fields these packets indicate are a projectile and a target, both of which are defined by the user from which the projectile originated (who sends the packet).  Supplemental information is also sent in a variety of fields, but these are the only fields that can not be inferred by the server alone.  The projectile game object is located and the target game object - plural "objects" in the case of `SplashHitMessage` - is located and the two are associated into a resolution object, occasionally, also referred as "historical," explaining the activity.  It further produces a formulaic application that explains how the target is to be affected by the projectile.  This formulaic application is sent to select targets' control agency (`Vitality.Damage`) for use in the object's syncrhonized envirnoment.

If damage is done to the target, but not enough damage to destroy the target, "damage awareness" resolution messages are produced that provided the target feedback about the successful but detrimental interaction.  If the target is destroyed in that interaction, appropriate resolution messages are sent for its "destruction awareness" to both the server in general and that user in specific.  The previous damage calculations can also be rewound before any resolution messages are sent if additional checks reveal mistakes in the application or the basic interaction.

Damaging also factors in changes from a previous update, jammering, since that is also a result of being hit by projectiles.  Being affected by such a projectile inflicts the status; being destroyed, clears the status.

__Repair__
Repair is controlled by the general property `repairable` and related properties define how the object is affected.  The most important property is  RepairIfDestroyed` which allows a property to be repaired across a theshold, changing its object state from dstroyed to not-destroyed. `RepairRestoresAt`, equally important, determines that threshold amount of health for the object change.  The game object definitions list file, colloquially "ADB", specifies this ability to be repaired as `canBeRepairedByNanoDispenser`.  The implement utilized by the repair activity is thus clarified.

The results of a successful repair attempt are always the same: the target has an increased amount of health.

All repair attempts are interpretted by the client as its user brandishing the said tool-weapon in front of a damaged game object.  This corresponds to an `ItemUseMessage` packet that signifies various elements of the exchange.  As before, supplemental fields exist, but are generally ignored.  The important fields are the target information and the tool-weapon information.  The tool-weapon can be determined by checking what is in the user's hands at the time, without need for using the packet data, but the check is verified against what the packet reports because "no tool-weapon used" is a valid object interaction.  In the past, the act of repairing was facilitated by pairing the target with the tool-weapon at the point where the packet is received and performing the whole process right there, despite the interuption of I/O and the need to re-write the code for each object case because ...

The now-employed method sends a common useage message (`CommonMessages.Use`) to the target object control agency, coupled with information about the tool-weapon being used, and allows the target to decide how to resolve the interaction.  More information validation follows, whether conditions for the repair action are applicable, and then the said activity is performed.  The resulting vital statistic change propagates into potential object changes and this all leads to the construction of messages that are dispatched to the expose those changes to other users through the event services.

___Features___ (partially itemized)
__Explosives, Players, and Teledpads__
These three objects are unique in context of the damage and repair activities previously outlined.
- Explosives can not be repaired but do follow a refined damage procedure that is integrated with a superficial jammering procedure.
- Tool assisted "repairing" to a player object does springboard from the same `UseItemMessage` packet as described above, but it involves the use of either a BANK tool-weapon to increase "armor" or the use of a medical applicator tool-weapon to increase "health".
- All telepad objects do not have individual control agencies and are currently synchronized on one of the event services.  They have not been changed as far as this update is concerned.

__Terminals__
- Types: equipment  (not Ancient-type), vehicle  (not Ancient-type), certification
- Can be destroyed
- A destroyed terminal can not be accessed
- Can be repaired
- Will start working again at 50%

__Spawn Tubes__
- Can be destroyed
- Can not be damaged by friendly fire
- Sanctuary spawn tubes in non-tower buildings are indestructible
- Destroyed spawn tubes will update the facility information window
- When all tubes are destroyed, the facility can not be used as a spawn point
- Can be repaired
- Will start working again at 50%

__Generator__
- Can be destroyed
- Can not be damaged by friendly fire
- Destroyed generators will update the facility information window
- Will explode in ten seconds upon destruction, killing all players within 14m
- Can be repaired
- Will start working again at 50%

__Implant Terminal__
- Can damage (see Caveats)
- A destroyed terminal can not be accessed
- Kills any player using the terminal while the terminal is destroyed (vehicle-like behavior)
- Can repair (see Caveats)
- Will start working again at 50%

___Caveats___
1. Generator animation, i.e., the spinning and crackling, is tied to the facility state.  ~~The only cosmetic change to a destroyed generator is the "Destroyed" descriptor added to the hover text.  For this reason, it is also impossible to know when a generator is bound to explode.~~  Give it a try.
2. The Implant Terminal is in a weird state.  Formally, it is comprised of two components - the implant terminal mech (model) and the implant terminal interface (window).  Both have interaction points - upper (mech) and lower (interface) - and both are supposed to be damageable and repairable.  For the purposes of this interaction, the state of the hood is unrelated to the position of the interaction points.  Destroying the mech portion of the terminal stops players from being able to use it until repair.  Destroying the interface portion does not.  Even if there is a simple way of testing this, there is no simple way of communicating to players when the interface needs to be repaired.  Several issue exist with the components when attempting to make it work properly one way or another, with unintuitive requirements being a major one.  For this reason, the primary damage and repair point has been assigned to the upper interaction (mech) and it has been given 1500 health, the sum of both the interface and the mech.
3. The implant terminal hood, if a player is mounted in the terminal at the time it is destroyed, will not lift back up until some one mounts the repaired terminal again.  It may bob once visually on some clients.  These are mearely visual glitches for players who are in the zone at the time.
4. Destroying explosive deployables by shooting at them causes them to smoke before they disappear.  This is a trigger effect.  This may not be vanilla behavior.
5. Destroying sensor-type deployables should cause a secondary explosion to occur near the top-mid portion, the thicker part of the pole.  This is a triggered effect.  If it does not align properly somehow at some angle, please report the location of the build surface for testing.  This may not be vanilla behavior.

___Addenda___
1. The implant terminal mech has a deprecated `Constructor` function that is currently used as its primary server object builder function.  To facilitate proper operation, the position of the interface component is copied for the position of the mech, though this may not be true.  A `Constructor` function that includes position information has been provided and is expected to replace all old mech creation.  @Mazo
2. If you build or rebuild the server, you will see a significant number of deprecation warnings.
3. Additionally, as mentioned in a previous update, the only generator that is properly setup is the one in Anguta, Ceryshen.  Other bases will need their generators and the location of those generators in their map files.  Please double-check that the position of the Anguta generator is correct as well. @Mazo
4. While transferring repairing from the primary `WorldSessionActor` `UseItemMessage` case, I found slicing up the existing cases in regards to maintain existing hacking protocol unsatisfactory.  Since my ultimate plan was to move those over too, I did just that, using the same `CommonMessages.Use` messaging pattern that I did with repair activities and started work on my "Progress Bar" feature.  Some code for a will-become `HackingDefinition` class was also written.  All of the "step functions" as well as the "finalize" functionality that were used for hacking individual objects was also removed from `WorldSessionActor`.  At that point, I realized I had started working on the WRONG SYSTEM but, given that I had done so much, I didn't want to leave the conversion half-finished.  So I left it at 3/4 finished.  It still works.
5. Due to the above change, corpse revival no longer has to juggle target health and target state until it is complete.
6. ~~As mentioned above, destroying the generator updates the facility information window.  It actually updates it three times: once when the generator is destroyed (condition "Critical"), once when it explodes ("condition "Destroyed"), and once when it is in working order again (condition "Normal").~~
7. As mentioned above, destroying the spawn tubes updates the facility information window.  When they are all destroyed, the information window indicates that and the option on the deployment map vanishes; when at leats one is repaired, the information window indicates that and the option on the deployment map is visible again.
8. With the changes made to damage, friendly units can no longer be jammered by allies, unless they are hacked.  No jammable unit can be hacked (or infected) in such a way that this can be tested.  Allied units will still not be jammed.
9. Resolved low-level issue with hastily cleaning up vehicle wreckage causing unreleased players who died in wreck to see the world vanishing around their corpse position.
10. Closed unfortunate lingering database connection.
11. The Beamer's alternate fire mode only does armor damage against other players (standard exo-suit).  Is that correct?
12. Only the TelepadDeployable remains as a "`SimpleDeployable`" and, in a few updates, may no longer be that.  Class structure for deployables will change slightly at that time.
13. The Resource Silo "Use" test constantly complains about having an error in it, despite passing completely.


Fate-JH:
**Update**  The generator now becomes _Critical_ at ~50% health after damage.  At 0 health, it will explode in 10s and, upon explosion, proceed to the _Destroyed_ condition.  No damage or repair can affect the Generator until it has exploded.  The Generator will show no sign of repair until ~50% where it will be restored to _Normal_ condition.  Damaging the Generator back to 0% prior to restoring _Normal_ condition will not cause a subsequent explosion.

Whenever the Generator takes damage and is not destroyed, it will broadcast a message to all players in the SOI that is is taking damage.  When the Generator starts the explosion coutdown, that will also broadcast a message warning players, specifically to evacuate the Generator room.  After exploding, the Generator will broadcast no other messages until it it restored to _Normal_ condition.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
```diff
@@ -95,69 +177,74 @@ class PlayerControl(player : Player) extends Actor
       super.CancelJammeredSound(obj)
     case _ => ;
   }
+
+  def RepairValue(item : Tool) : Int = if(player.ExoSuit != ExoSuitType.MAX) {
+    item.FireMode.Modifiers.Damage0
+  }
+  else {
+    item.FireMode.Modifiers.Damage3
+  }
```
Mazo, line 186: Isn't this always going to hit the first case of Damage0, unless I'm missing something where a player in a MAX can use a Tool to repair something?

Fate-JH: Damage and repair actions resolve in a target-centric context, i.e., the `player` here is the one being repaired, the one being managed by the `PlayerControl` agency, not the person doing the repairing (`user`).

Mazo: Ah yeah of course, that makes more sense. Even then, why the different behaviour is present isn't immediately obvious. MAXes have a different repair rate than other armour?

Fate-JH: The way we had figured these damage/repair values works like this:
`damage0` - vanilla infantry
`damage1` - armor
`damage2` - aircraft
`damage3` - mechanized infantry
`damage4` - unknown (used for amenities and some deployables)

Fate-JH: Realistically-speaking, however, the `bank` does not naturally benefit from this.  The `nano_dispener` does.  This part is included for customization potential.

NickPSF: > The way we had figured these damage/repair values works like this:
> `damage0` - vanilla infantry
> `damage1` - armor
> `damage2` - aircraft
> `damage3` - mechanized infantry
> `damage4` - unknown (used for amenities and some deployables)

damage1 is base amentities and ground vehicles
damage4 is BFRs/battleframe robotics

Fate-JH: Changed.

common/src/main/scala/net/psforever/objects/serverobject/resourcesilo/ResourceSiloControl.scala
```diff
@@ -30,7 +30,14 @@ class ResourceSiloControl(resourceSilo : ResourceSilo) extends Actor with Factio
 
   def Processing : Receive = checkBehavior.orElse {
     case ResourceSilo.Use(player, msg) =>
-      sender ! ResourceSilo.ResourceSiloMessage(player, msg, resourceSilo.Use(player, msg))
+      if(resourceSilo.Faction == PlanetSideEmpire.NEUTRAL || player.Faction == resourceSilo.Faction) {
+        resourceSilo.Zone.Vehicles.find(v => v.PassengerInSeat(player).contains(0)) match {
```
Mazo, line 34: Does this now require the player to stay inside the ANT to continue depositing NTU? Because you could get out and leave it running on live.

common/src/main/scala/net/psforever/objects/serverobject/structures/BuildingControl.scala
```diff
@@ -35,43 +37,62 @@ class BuildingControl(building : Building) extends Actor with FactionAffinityBeh
         building.Amenities.foreach(_.Actor forward FactionAffinity.ConfirmFactionAffinity())
       }
       sender ! FactionAffinity.AssertFactionAffinity(building, faction)
+
+    case Building.AmenityStateChange(obj : SpawnTube) =>
+      if(building.Amenities.contains(obj)) {
+        SendMapUpdate(allClients = true)
+      }
+
+    case Building.AmenityStateChange(obj : Generator) =>
```
Mazo, line 46: Can this be merged with the above case by passing in a more generic object class since the two are identical otherwise?

common/src/main/scala/net/psforever/objects/serverobject/resourcesilo/ResourceSiloControl.scala
Fate-JH: Is `ResourceSilo.Use` a one-and-done message, or does it need to keep being sent to keep depositing NTU?

Mazo: Can't remember off the top of my head. I want to say it repeats, but I'd have to test it to check.

common/src/main/scala/net/psforever/objects/serverobject/structures/BuildingControl.scala
Fate-JH: For now, they're the same.  You could technically make them part of the same case with `obj : Amenity`; but, I'd still then want to check what kind of `Amenity` they are so we could use these cases for triggered facility updates.

common/src/main/scala/net/psforever/objects/serverobject/resourcesilo/ResourceSiloControl.scala
Fate-JH: A quick check of `WSA` suggests that the process of discharging into a silo uses a `Cancellable` timer.

Mazo: Double checked and you're right, it's one use message and then the timer kicks in. Closing.

PR #373, opened 2020-04-18
It sucks.

Fate-JH:
Make sure to test this out before running with it.  I do have another solution, potentially, but it will probably break tests that I don't want to waste time fixing right now.


Fate-JH:
It sucka s little less now.

PR #374, opened 2020-04-18
Swap surge into first implant slot and change darklight for audio amp

Mazo:


Fate-JH:
Can't I just set up my implants the way I want them? :P

NickPSF:
> Can't I just set up my implants the way I want them? :P

Yes, _you_ can!

Issue #372, opened 2020-04-18
Spawning - Instant Action - Spawn Tubes - Instant Action will often times cause players to get pushed out of the Spawn Tube into a bad location


NickPSF:
Instant Action will often times cause players to get pushed out of the Spawn Tube into a bad location.

Players get stuck or dropped below the world.

NickPSF:
milestone_fixes-3af9c15692

Issue #358, opened 2020-04-17
Gating - Warpgates - Various Issues with Warpgates


NickPSF:
TR can only get to Solsar from Sanctuary, and capturing Solsar does not grant you additional links.
NC has no functioning Warpgates, they all "one-way" instantly warp you back to Sanctuary Spawn
VS can only get to Ceryshen from Sanctuary

I think the current Test Live (51200) does something special to handle this. @SouNourS would know.

NickPSF:
Fixed in milestone_fixes-3af9c15692

Issue #342, opened 2020-02-02
Gating - Warpgates - Geowarps - Geowarps do not warp players to Cavern zones


NickPSF:
Geowarps do not warp players to Cavern zones

NickPSF:
The cave timers appear to be somewhat functional client side? Is there anything we can do here?

Mazo:
Currently no, they would need a lot of work to be properly functional. The only reason the timers are working client side on 51200 is because Sounours has added a few lines to send an extra packet with essentially an arbitrary timer.

51000 has no cave timers.

NickPSF:
Any hacky solution for this currently?

Mazo:
Beyond the hacky "turn off the geowarps"?

Not yet. I'll try and take a look at some point but until proper cave rotation is in I doubt we can do better than sending them to the currently linked cave.

Mazo:
Also it's probably for the best they stay off for now, until we find a fix for the zoning crash from the turrets.

NickPSF:
Sending them to the currently linked cave is good enough for me.

Dethdeath:
Geowarp zoning is much more stable now, but you can still get a client crash every once in a while. There's also an issue where you can get sent back into the cave instantly when zoning to a regular continent. This bypasses the "you have left the protective shield around the warpgate" message and action which is usually required to go back through.

Fate-JH:
Going to close this because the core issues is more or less solved.
Issues that arise from the geowarp gates working should become their own tickets.

Issue #364, opened 2020-04-17
Commands - Regular Users - On Master CI 51000, regular users can use /zone and /warp, despite not being a GM


NickPSF:
On Master CI 51000, regular users can use /zone and /warp, despite not being a GM.

NickPSF:
Needs to require the same permissions as !list, etc.

NickPSF:
Fixed in milestone_fixes-3af9c15692

PR #376, opened 2020-04-19
Milestone fixes

Mazo:
Few fixes for milestone issues

#342 - Cavern warpgates are disabled for now
#358 - All warpgates are broadcast warpgates now
#372 - Instant action spawn point now respect overrides like z offsets (Hopefully this doesn't break anything else @Fate-JH)

On top of that, an unreported bug is fixed where resource silos weren't showing their values correctly after the server was started. All resource silos also now start with 100% NTU.

### DIFF
--- a/common/src/main/scala/services/vehicle/support/VehicleRemover.scala
+++ b/common/src/main/scala/services/vehicle/support/VehicleRemover.scala
@@ -1,6 +1,7 @@
 // Copyright (c) 2017 PSForever
 package services.vehicle.support
 
+import akka.actor.ActorRef
 import net.psforever.objects.Vehicle
 import net.psforever.objects.guid.{GUIDTask, TaskResolver}
 import net.psforever.objects.zones.Zone
@@ -26,7 +27,9 @@ class VehicleRemover extends RemoverActor {
     entry.zone.GUID(vehicleGUID) match {
       case Some(vehicle : Vehicle) if vehicle.HasGUID =>
         val zoneId = entry.zone.Id
-        vehicle.Actor ! Vehicle.PrepareForDeletion()
+        if(vehicle.Actor != ActorRef.noSender) {
+          vehicle.Actor ! Vehicle.PrepareForDeletion()
+        }
         //escape being someone else's cargo
         (vehicle.MountedIn match {
           case Some(carrierGUID) =>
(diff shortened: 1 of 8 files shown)


Mazo:
Another fix added for #364

Mazo:
Another fix added for #378

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -4329,55 +4332,59 @@ class WorldSessionActor extends Actor
...
-              deadState = DeadState.Alive //must be set here
-            case _ => //seated in something that is not a vehicle or the vehicle is cargo, in which case we can't move
-              deadState = DeadState.Alive
+        case (true, pos) =>
+          // continent.Id == "c1" || continent.Id == "c2" || continent.Id == "c3" || continent.Id == "c4" || continent.Id == "c5" || continent.Id == "c6" ||
+          if (player.isAlive && (admin || continent.Id == "z8" ||
+            continent.Id == "tzshtr" || continent.Id == "tzcotr" || continent.Id == "tzdrtr" ||
+            continent.Id == "tzshnc" || continent.Id == "tzconc" || continent.Id == "tzdrnc" ||
+            continent.Id == "tzshvs" || continent.Id == "tzcovs" || continent.Id == "tzdrvs")) {
+            deadState = DeadState.Release //cancel movement updates
```
Fate-JH, line 4369: What's all this, then?

Up there with CSRZone too.

common/src/main/scala/services/vehicle/support/VehicleRemover.scala
```diff
@@ -26,7 +27,9 @@ class VehicleRemover extends RemoverActor {
     entry.zone.GUID(vehicleGUID) match {
       case Some(vehicle : Vehicle) if vehicle.HasGUID =>
         val zoneId = entry.zone.Id
-        vehicle.Actor ! Vehicle.PrepareForDeletion()
+        if(vehicle.Actor != ActorRef.noSender) {
+          vehicle.Actor ! Vehicle.PrepareForDeletion()
+        }
```
Fate-JH, line 32: I'll allow it, though the bulk of this file will be going out the window.

pslogin/src/main/scala/WorldSessionActor.scala
Mazo: Ported directly from Soun's PTSv3 branch. I think it's to allow non-admins to warp to certain zones like vr training or old oshur.

common/src/main/scala/services/vehicle/support/VehicleRemover.scala
Mazo: Yeah I figured when I saw you were getting rid of VehicleRemover, but at least it should work as a temporary bandaid

PR #391, opened 2020-04-22
The Facility's Generator Is Currently under Attack!

Fate-JH:
The facility's generator is currently under attack!


Fate-JH:
The facility's generator is currently under attack!

PR #395, opened 2020-04-27
Add kamon.io metrics reporting

Mazo:
If you want to use this make sure you get an api key from kamon.io and put it in application.conf, and enable the Kamon.Active flag in worldserver.ini


Fate-JH:
Please add blank lines to the end of worldserver.ini.dist and application.conf.

PR #396, opened 2020-04-29
Vanu turret actor message spam reduction

Mazo:
So, one of the things plugging in Kamon.io flagged up almost immediately was this:

image

62.1k messages for FacilityTurretControl per minute? That can't be right...

image

Oh, Right.... I forgot about that.

So this PR addresses that. Vanu turrets will be notified when their weapon is discharged, and will begin the recharge timer at that point, and subsequently cancel the timer when the ammo is fully recharged. No more actor message spam

Also included is a quick fix that stops a log warning from firing when a splash damage hits the world itself.


Mazo:
Fast tracking this merge since @NickPSF has tested and confirmed it still seems to work as expected, and @SouNourS will need this merged for when he pulls in the kamon stuff

PR #401, opened 2020-05-01
Kamon tweaks

Mazo:
Just a couple of fixes for Kamon, and an attempt at reducing the logspam coming from oshi.util.FileUtil on PTSv3


config/logback.xml
```diff
@@ -39,4 +39,6 @@
 
   <!-- Logger specific overrides -->
   <logger name="com.github.mauricio.async" level="OFF"/>
+  <logger name="oshi.util.FileUtil" level="OFF"/>
+  <logger name="o.h.platform.linux.LinuxDisks" level="OFF"/>
```
pschord, line 43: This last one isnt right as logback will abbreviate the full package name. if the oshi package is giving us issues, just block the whole package (`oshi`) instead

pschord: https://stackoverflow.com/a/48659858/5768099

Mazo: Thanks @pschord I wasn't sure what package it was since it seems to be abbreviated, was just hoping that would work. Has been amended.

Issue #397, opened 2020-05-01
Implants - Stamina Regeneration - When spamming an Implant on/off, it never registers as being off, and continuously drains Stamina until the player is out of Stamina and can no longer regenerate it until dying and respawning


NickPSF:
When spamming an Implant on/off, it never registers as being off, and continuously drains Stamina until the player is out of Stamina and can no longer regenerate it until dying and respawning.

**Expected Result:** Spamming an Implant on/off will not progressively add additional drains to Stamina.

**Video:** https://streamable.com/cwemdl and https://streamable.com/x2akg4

NickPSF:
Added a new video. https://streamable.com/x2akg4 (only needs to turn Implant on/off too quickly once to bug out a character's Stamina)

NickPSF:
Issue no longer occurring on corresponding CI test server

PR #415, opened 2020-05-03
Fix implant stamina draining permanently when the client tries to activate an already active implant

Mazo:
See #397


NickPSF:
Fix confirmed

PR #417, opened 2020-05-03
Painbox v2

### LINKED ISSUE #371: Painfields - Amp Station & Tech Plant - Z-axis of Pain Field is too high for these Facilities
Amp Station & Tech Plant Z-axis of Pain Field is too high for these Facilities.

Lowering them will help more than harm, the area in Basement that will be affected by Pain Field is way more relevant and way less traveled than the Lobby of these two Facility types.

Let's also change damage to 2, from 10.

Mazo:
All pain fields were previously radius based, even those that were not supposed to be. I'm somewhat convinced they were originally line of sight based on live, but we can't replicate that since our server has no 3d model information.

Because of the above, I came to a workaround involving nearby amenities and bounding boxes. Essentially on startup a non-radius painfield will find amenities nearby, and create a bounding box around them. From that we can determine if a player is within the spawn room or not.

The caveat being that this method won't work for anything other than above ground spawn facility spawn rooms. Currently this isn't a big deal as generators pain fields aren't implemented, and I've disabled non-radius pain fields in caves as an acceptable "temporary" state.

One notable trick for the bounding box calculation is instead of rotating the bounding box to match the base rotation the player coordinates are instead rotated. This allows for the bounding box to stay XY aligned and vastly simplifies the logic for checking if the player is within the bounding box

Fixes #371

### DIFF
--- a/common/src/main/scala/net/psforever/types/Vector3.scala
+++ b/common/src/main/scala/net/psforever/types/Vector3.scala
@@ -326,4 +326,11 @@ object Vector3 {
       z
     )
   }
+
+  def PlanarRotateAroundPoint(point: Vector3, center: Vector3, radians : Float) : Vector3 = {
+    val x = (Math.cos(radians) * (point.x - center.x) - Math.sin(radians) * (point.y - center.y) + center.x).toFloat
+    val y = (Math.sin(radians) * (point.x - center.x) + Math.cos(radians) * (point.y - center.y) + center.y).toFloat
+
+    Vector3(x, y, point.z)
+  }
 }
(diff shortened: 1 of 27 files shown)


common/src/main/scala/net/psforever/types/Vector3.scala
```diff
@@ -326,4 +326,11 @@ object Vector3 {
       z
     )
   }
+
+  def PlanarRotateAroundPoint(point: Vector3, center: Vector3, radians : Float) : Vector3 = {
```
Fate-JH, line 330: So, after realizing that this is z-axis rotation that should be in the same vein as function `Rz`, I double checked and discovered that `Rz` is (probably) written incorrectly when calculating the y-coordinate value.  Shouldn't be fixed, yet, as I'm expecting the wrong behavior as the correct result in some cases.

PR #423, opened 2020-05-09
Validate usernames on account creation like PSFPortal

pschord:
I have not tested this patch yet. Could someone give it a go?


Mazo:
Tested, seems to be working fine.

A fun fact, turns out PacketCodingActor can't handle non US-ASCII characters in usernames (e.g. £) and the client will just hang on authenticating, but that's fine for now.

`[ INFO] PacketCodingActor - Failed to unmarshal the inner packet of a SlottedMetaPacket: Failed to parse game packet 0x01: credential_choice: US-ASCII cannot decode string from '0xa3a3a3'. Data : ByteVector(32 bytes, 0x01030000000f0000008b44656320203220323030394180a3a3a3816154000000)`

PR #424, opened 2020-05-09
First Time Events

Fate-JH:
“Multiple exclamation marks,' he went on, shaking his head, 'are a sure sign of a diseased mind.”

There's currently no point to removing the yellow bang marks through natural environment interaction if the database is not ready to commit to their memory.  Let's just not show them for now.

___Caveats___
I may not have them all.  I certainly have most of them, anyway.

___Addenda___
I was flying around Hunhau and found a lot of terminals that were not connected to the building in which they were installed.   I really do not know how to call these "non-facility" slums but the specific room had an overhead walkway area.


NickPSF:
Let's merge it

PR #429, opened 2020-05-13
Let's Move Item, Together

Fate-JH:
At first, it was only a remote electronics kit that did not like playing by the rules.  It violated boundaries and placed itself wherever it wanted, overlapping whatever other object's personal space that it wanted.

For a while, it seemed to settle, and no grave concerns could be perceived.

Then, like wildfire, the flagrant abuse of trust and encroachment began, as if every inventory was possessed by a malevolent spirit of disorder!  Things were going everywhere they should not have been in the backpacks of all players.  The bullets were getting in the rocket launchers.  The medkits were getting into the bullets.  Items were jumping out of people's hands, lost somewhere between the hand and ground, to be replaced by another item of unfair distribution.  Rifle and shell alike were stacking up to the high heavens, beyond what was reasonable.  The item blamed the container and the container blamed the item, each only arguing that one was not right.  In all this tumult, the world ended many times.

And, now, like a thunder crack, the codebase is shattered and rebuilt.  All objects must now exist in harmony.

And that's enough of that malarkey.

Serious now.  I have made all item insertion and removal processes require some synchronization between the source container and the destination container to ensure that all moved entities are properly shifted from one spot to the next without violating numerical limits or bounded limits and without eradicating another entity already in the same position.  The change crept over to affect every aspect of container item manipulation, since it's all just a syncrhonized insertion or removal.  Buy equipment, sell equipment, drop equipment, pick up equipment, move items from one hand to the next, move items from one container to the next, request loadouts: try all of these behaviors and more and tell me where any of it goes pear-shaped.

But do not tell me that the vehicle terminal is broke again.

___Features___
__`Containable` and `ContainableBehavior`__
The central logic for handling synchronized container activity.  The main selling point is the handling of `MoveItemMessage` logic for and between: `Player` objects, `Vehicle` objects, and `Locker` objects (the storage component of `Lockers` anyway, not the kind you find in spawn rooms).  These are the only object types that support container interaction in the game.  A fast-forward approach when moving items from one hand to the other exists and a tiered structure allows for object-specific callbacks to be written, greatly benefitting the variety of perspectives offered by the affected containers.  The basic put cases and removal cases, however, offer a springboard for writing whatever sort of inventory logic is necessary.

__`WorldSession`__
Specific write-ups for logics for manipulating items and inventory have been carefully transitioned out of `WorldSessionActor` itself and placed here.  While this does result in an exceptional amount of logic that hooks in resources that were collected for the benefit of `WorldSessionActor` to facilitate the same operations that `WorldSessionActor` had performed on its own, the resulting item manipulations function smoothly enough.

__`LockerContainer`, `LockerEquipment`, and `LockerContainerControl`__
Originally, `LockerContainer` was a flat simple object designed only to handle the portable inventory portion of a user's locker.  It was `Equipment` since it was technically in an "equipment slot" between the holsters and the formal backpack inventory.  The arrangement is ill-suited for `ContainableBehavior` adaptation so `LockerContainer` objects now exist to wrap around `LockerEquipment` objects and extend to the latter these new `Containable` capabilities.

__`ZoneGroundActor`__
All messages for items coming and going from the ground previously needed to swing back to the `WorldSessionActor` that contracted the pick up or drop action.  Originally, this was just for logging messages; but, later, this endpoint had to be leveraged to make certain `BoomerDeployable` objects and `BoomerTrigger` objects worked correctly.  To avoid constantly passing `ActorRef` references to that player endpoint whenever interacting with the ground, the process was simplified in a way that eliminated the last endpoint.  The callbacks for the `ContainableBehavior` objects now handle anything that must be collected or deposited with care, such as `BoomerTrigger` items.  The Ground is much more generic now and deals only with managing the objects that are available, no longer being used for depiction.

__`InfantryLoadout` and `VehicleLoadout`__
Both are now functationally sorted either by the player that will affected by the loadout or by the vehicle that will be affected by the loadout, respectively.  The only communication they have with `WorldSessionActor` is for visualization and entity registration.

__`Avatar`__
As a sacrifice to the god of technical debt, the way items were logged as purchased at certain times had to be changed.  Previously, `WorldSessionActor` maintained the said purchasing cooldowns, mainly for `Vehicle` objects and for equipping the mechanized assault exo-suit and did so at the time of callback.  This way fine when that given player's `WSA` was the recipient of the terminal message that brough the goods and services to the player's feet, then unloaded and set up the player's backpack.  The inventory now being handled internally by individual control agencies, it was necessary to localize the cooldown timers into the yuser's avatar (the part of the player character that persists as long as the user is logged in).  Additionally, this system was extended to be made capable of ingesting "use" timers (such as seen with `Medkit` items).  Game objects are still remembered by their global object id; exo-suits are kept separate, remembered by the organization of their `Enumeration` object and, then, the different mechanized assault exo-suits separate from normal exo-suits.

__`OrderTerminalDefinition.Tab`__
To allow for terminals to send targetted messages directly to the control agency that will sort the products of the message - individual beings of equipment for `BuyEquipment`, or loadouts information - different tabs now direct to which of a number of targets its output should be dispatched.  For example, infantry loadout information can be sent directly to the infantry player that will use it, while vehicle loadout information will be sent directly to the vehicle that will use it.  (For the moment, only the affected inventory logic is redirected.  Other outlet for this redirection is not explored.)

___Caveats___
1.  `TaskResolver` objects and the `ask` pattern mix as per needs of the logic.  As these are both stark contrasts of synchronicity in the Akka system, they don't visually blend together well, even if they work ...
2.  While most of the extraction moved into `WorldSession` might have a return type, only the ones that return `TaskResolver.GiveTask` require that return.  To run the code, this object has to be passed to--you guess it--a task resolver object.  Some of these functions also pass in their own reference to a task resolver object for the purposes of registering or unregistering an entity in the course of execution or as recovery from a failed execution.  Putting that aside, most of the return that is `Future[Any]` is underutilized.  The primary purpose also yields to internal logic that is administered but never properly reported, e.g., removing an entity from a slot (outer) and, upon successful removal, unregistering it (inner).  In any case, making the output of these functions more useful for statusing the outcome is a definite place to grow.
3.  The error catching is not where I'd like it to be, and this is especially true for instances where `ask` logics might go into a request timeout (`AskTimeoutException`).  That's a risk... almost everwhere.  I am actually wary of writing logic for these failures until I see them in the wild.  I hope the 5s timeout countdown makes it extremely unlikely to encounter such an issue from a task not being completed.  The main issue warded by current methods involve not blocking future terminal interaction.
4.  Anything that can be inherit `Container` interaction must be a `PlanetSideServerObject` type of game object to take advantage of `ContainableBehavior`.  The logic for handling an object without a control agency but still being a container doesn't currently exist.
5.  It's not a problem that the case for `VehicleLoadout` doesn't actually return the mounted weapons of the loadout for depiction.  Although mounted weapon swapping actually works in all ways except the UI, failure of the UI upon update happens regardless of how the matter is currently sorted and is a showstopper for the feature.  We can't advise people to keep getting in and out of their vehicle to correct the issue.  Ignoring "old" and "new" weapon data and just updating all weapon magazines is the safest thing to do in the meanwhile.  The vehicles that are known to swap their mounted weapons aren't yet implemented anyway.
6.  `MoveItemMessage` handling code has a 1s timeout for each stage of its operation (`ContainableBehavior`), measly compared to the 5s timeout for other forms of item manipulation (`WorldSession`).  This is the result of me being miserly earlier in the development of this branch but paranoid during later development and not actually reflective of a well-planned out strategy.  Do tell me which of these groups of behaviors are more prone to failure - basic move item requests, or more complicated item manipulations - that which is given tigher constraints or that which is granted more lenient time budgeting.
~7.  The mappings for purchase and use cooldowns are not going to be where they are now, in the `object` of `WorldSessionActor`, so one should not get comfortable with them being there.  They're eyesores.~ This is now located in `Avatar` and also accessible through `Player`.

___Addenda___
1.  For anything that looks like it should have been changed to reflect the new way that objects are added or removed from containers, that pathway most likely is considered existing at the terminal end of the lifespan of that entity and is not governed by manual item manipulation.  The more expedient, straightforward approach is acceptable until failure becomes apparent.
2.  For the part where the page/tab system for order terminals allows local redirection of the response message, a more optimal solution would be to let external configuration of the redirection be allowed using function literals.  It's not necessary right now but shall be considered in the future.
3.  The repair rearm silos in sanctuary flanking the HART buildings are still busted as far as I can tell.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,8 +48,8 @@ lazy val commonSettings = Seq(
     "org.scala-graph"            %% "graph-core"    % "1.12.5",
     "io.kamon"                   %% "kamon-bundle" % "2.1.0",
     "io.kamon"                   %% "kamon-apm-reporter" % "2.1.0",
-    "org.json4s"                 %% "json4s-native" % "3.6.8",
-  ),
+    "org.json4s"                 %% "json4s-native" % "3.6.8"
+  )
 )
 
 lazy val pscryptoSettings = Seq(
--- a/common/src/main/scala/net/psforever/objects/Avatar.scala
+++ b/common/src/main/scala/net/psforever/objects/Avatar.scala
@@ -58,6 +58,12 @@ class Avatar(private val char_id : Long, val name : String, val faction : Planet
 
   private var vehicleOwned : Option[PlanetSideGUID] = None
 
+  private var lastUsedEquipmentTimes : mutable.LongMap[Long] = mutable.LongMap[Long]()
+
+  private val lastUsedExoSuitTimes : Array[Long] = Array.fill[Long](ExoSuitType.values.size)(0L)
+
+  private val lastUsedMaxExoSuitTimes : Array[Long] = Array.fill[Long](4)(0L) //invalid, ai, av, aa
+
   def CharId : Long = char_id
 
   def BEP : Long = bep
@@ -219,6 +225,47 @@ class Avatar(private val char_id : Long, val name : String, val faction : Planet
   def VehicleOwned_=(guid : Option[PlanetSideGUID]) : Option[PlanetSideGUID] = {
     vehicleOwned = guid
     VehicleOwned
+  }
+
+  def GetLastUsedTime(code : Int) : Long = {
+    lastUsedEquipmentTimes.get(code) match {
+      case Some(time) => time
+      case None => 0
+    }
+  }
+
+  def GetLastUsedTime(code : ExoSuitType.Value) : Long = {
+    lastUsedExoSuitTimes(code.id)
+  }
+
+  def GetLastUsedTime(code : ExoSuitType.Value, subtype : Int) : Long = {
+    if(code == ExoSuitType.MAX) {
+      lastUsedMaxExoSuitTimes(subtype)
+    }
+    else {
+      GetLastUsedTime(code)
+    }
+  }
+
+  def SetLastUsedTime(code : Int) : Unit = SetLastUsedTime(code, System.currentTimeMillis())
+
+  def SetLastUsedTime(code : Int, time : Long) : Unit = {
+    lastUsedEquipmentTimes += code.toLong -> time
+  }
+
+  def SetLastUsedTime(code : ExoSuitType.Value) : Unit = SetLastUsedTime(code, System.currentTimeMillis())
+
+  def SetLastUsedTime(code : ExoSuitType.Value, time : Long) : Unit = {
+    lastUsedExoSuitTimes(code.id) = time
+  }
+
+  def SetLastUsedTime(code : ExoSuitType.Value, subtype : Int) : Unit = SetLastUsedTime(code, System.currentTimeMillis())
+
+  def SetLastUsedTime(code : ExoSuitType.Value, subtype : Int, time : Long) : Unit = {
+    if(code == ExoSuitType.MAX) {
+      lastUsedMaxExoSuitTimes(subtype) = time
+    }
+    SetLastUsedTime(code, time)
   }
 
   def Definition : AvatarDefinition = GlobalDefinitions.avatar
--- a/common/src/main/scala/net/psforever/objects/Avatar.scala
+++ b/common/src/main/scala/net/psforever/objects/Avatar.scala
@@ -57,12 +57,24 @@ class Avatar(private val char_id : Long, val name : String, val faction : Planet
   private var lfs : Boolean = false
 
   private var vehicleOwned : Option[PlanetSideGUID] = None
-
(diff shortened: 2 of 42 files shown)


Fate-JH:
Expanded completion conditions for loadout tasking (and some others).  May inflate timeout for those tasks, if necessary.

Fate-JH:
Dragons will not destroy your purchasing power.

Mazo:
Looting an item from a corpse will cause a client disconnect.

Fate-JH:
Grave robbery is no longer punishable by certain death.

Fate-JH:
I git wrong sometimes.

PR #433, opened 2020-05-14
Unowned Vehicle Decay

Fate-JH:
Restoring decay to vehicles that become unowned.

Made `PlayerSource` cling to life less.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/vehicles/VehicleControl.scala
+++ b/common/src/main/scala/net/psforever/objects/vehicles/VehicleControl.scala
@@ -2,7 +2,7 @@
 package net.psforever.objects.vehicles
 
 import akka.actor.{Actor, ActorRef, Cancellable}
-import net.psforever.objects.{DefaultCancellable, GlobalDefinitions, SimpleItem, Vehicle, Vehicles}
+import net.psforever.objects._
 import net.psforever.objects.ballistics.{ResolvedProjectile, VehicleSource}
 import net.psforever.objects.equipment.JammableMountedWeapons
 import net.psforever.objects.serverobject.CommonMessages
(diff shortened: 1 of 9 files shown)


Mazo:
In addition to what Dethdeath mentioned:

`On 61007 when a vehicle loses its owner it starts smoking and its lights start flickering as if damaged. When you start damaging these vehicles, their health appears 
 to be desynced. They do still take damage and blow up, but health shows as full until they die.`

The vehicle also doesn't update the mounting symbols for player's still in the zone, so it looks like you can't mount it as the driver even though the driver has left the zone. It does still let you mount though.

On a related note I set up a test with two characters.

Character A: Owns a Wasp. Driver seat set to empire, trunk is locked.
Character B: Owns a Mosquito. Driver seat set to empire, trunk is locked.

Character A:
image

Character B:
image


Now Character A get's in the Mosquito, taking ownership of it.

Character A:

image

Character B:

image

The Wasp is now unowned, and the trunk is showing as open Character B (who now owns no vehicle). However you can't actually open the trunk.

common/src/main/scala/net/psforever/objects/vehicles/VehicleControl.scala
```diff
@@ -73,11 +73,31 @@ class VehicleControl(vehicle : Vehicle) extends Actor
...
+            vehicle.Zone.GUID(player.VehicleOwned) match {
+              case Some(v : Vehicle) =>
+                v.Actor ! Vehicle.Ownership(None)
+              case _ =>
+                player.VehicleOwned = None
+            }
+            LoseOwnership() //lose our current ownership
+            GainOwnership(player) //gain new ownership
+          }
+          else {
```
Mazo, line 97: Should this else block not be one level higher, so it's the else block for `obj.Seats.values.exists(_.isOccupied)`?

Fate-JH:
Yeah, I know about the mount glyphs.  Can't figure out how to update them in real time, even though I'm sending the packets that should be updating them in real time each time.  This affects the driver seat moreso than the others.

I also don't know why it switches over to its slightly damaged model animation upon becoming ownerless.  Not sending the packet that updates to "ownerless" and relying on the floor update glyphs (that don't work reliably either) might be a way to go but that doesn't have interaction tests since we've always used the no-owner packet.

common/src/main/scala/net/psforever/objects/vehicles/VehicleControl.scala
Fate-JH: If `obj.Seats.values.exists(_.isOccupied)` fails, the player that tried to mount the vehicle had failed to mount the vehicle.  We shouldn't stop the decay timer in this case.  Additionally, `seat_num == 0` just goes through a little extra driver-specific code that ultimately does the exact same thing as the `else` clause.  I will do a bit of switch up with this, however, so comments will be added for clarification.

Mazo: Ah yeah, that's a good point. I was thinking about the case where someone is seated in something other than the driver seat, do we still want to start the decay timer in that case?

Fate-JH: No, any player occupancy in the vehicle should suspend the decay timer.  That's why `TryDismount` checks `!decaying && obj.Owner.isEmpty && obj.Seats.values.forall(!_.isOccupied)` (are we already decaying? are we unowned? is no one seated anywhere?) before trying to restart it.

Mazo: Ah that's all good then. I missed the bit in `TryDismount`.

PR #436, opened 2020-05-14
Doors should no longer close if any alive player is still within range

Mazo:
Hold the door, please.

It turns out if you had a backpack in the zone the `DoorCloseActor` thought you were still dead, so the door would unceremoniously slam shut in your face repeatedly, until your backpack despawned.

Instead, it will now stay open until you die, or move away. If that happens and another player is nearby, they will take over chivalrous duties of holding the door open until no live players are within range of the door, at which point it will close.


pschord:
@NickPSF can you test this change?

NickPSF:
Issue no longer occurring on corresponding CI server

PR #437, opened 2020-05-15
Fix Dockerfile

jgillich:
The scala-sbt image no longer has a latest tag so we must
provide a tag.

I've picked Java 8 and the latest SBT / Scala version, hope that's right. This is what's available: https://hub.docker.com/r/hseeberger/scala-sbt/tags (java_sbt_scala)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM hseeberger/scala-sbt
+FROM mozilla/sbt
 
 EXPOSE 51000
 EXPOSE 51001


pschord:
This Dockerfile doesn't really work past building the server. You cannot run the server without a DB. See our PSFCI https://github.com/psforever/PSF-ServerCI/blob/master/docker/Dockerfile to see what we do for building automatically. Also mozilla/sbt seems to be maintained better these days.

jgillich:
I think you'd usually run the db in another container, no? Updated to mozilla/sbt.

pschord:
Normally yes, but I dont know anyone who is using a Dockerized version of our server. I guess they may want to have the DB separate to avoid recreating it each time

jgillich:
I am using it. I want to contribute to PSFPortal, but first I need a containerized dev environment, and this is one of the components.

PR #445, opened 2020-05-16
Let's Move Item, Together, Again

Fate-JH:
There is no novel fancy flourish to this PR message.

#429


Mazo:
Still something weird happening with implants not finishing their initialization after dying + respawning on this branch.

image

Sometimes surge will initialize. The others I've not seen initialize, except on first spawn where everything seems to work normally.

image

Fate-JH:
It's weird but the only thing changed about implants is the loop counter structure (`for(0 to array.length-1)` to `array.indices.foreach`, both of which resolve to the same 0-n range).  Despite all the blue/green lines in the diff, the substance of the existing implant code does not appear changed.  It's the formatting that became weird.

I have a branch - `stamina` - where I'm looking into the specifics of whay implants are doing what they're doing now and extending some of their features such as cancelling before zone changes or exo-suit changes, and why stamina finds reason to stop regenerating.  That branch incorporates the changes in `sync-move-item`.

Mazo:
> why stamina finds reason to stop regenerating

Have you managed to reproduce that? After I fixed it originally the only time I've heard people complain about that is when they didn't realise there was a delay after running out of stamina before it recharges, or right after getting hit with an EMP (if I remember right)

Fate-JH:
The actual reason is quite easy to understand.  Your client will not dispatch `PlayerStateMessageUpstream` and `VehicleStateMessage` at the same time and, if it's sending neither `PSMU` packet nor `VSM` packet, then it's probably sending a `ChildObjectStateMessage` packet.  (It will typically pair `VSM` with `COSM`, but will also send `COSM` in situations where `VSM` is not applicable.)  Each of these serves a specific purpose - "is infantry," "is the driver in a vehicle," and either "has a mounted turret (driver or not)" or "is seated in a non-driver position."  There may be a little more to it where `COSM` may not be valid either but I'm still exploring that.

The short of it is that only `PSMU` handles stamina.

Mazo:
Oh I see what you mean, it doesn't regen when in a vehicle yeah. I thought you meant it would stop regenerating as infantry.

Fate-JH:
I'm going to have to manually repair and merge this branch due to all the changes in `PlayerControl` and, of course, `WorldSessionActor`.  I was trying to properly rebase it and git got to the point where it completely wrecked the format of `PlayerControl`.

Fate-JH:
Rebase complete.  I think it all works again but, for the sake of it, I'd like for there to be another testing of an inventory operations and of as many login situations as possible.

There is one situation I am not entirely happy about that I know about and involves a shot of the deployment map when moving around vehicles using `/warp`.  That's a very niche problem so I'll only look into it if issues similar to that appear when it comes to tradition vehicle gating.

Relogging should work correctly.  Relogging when dead or released should resolve into just respawning you at the most convenient location, hopefully closest to wherever you last died.  Originally, the code was supposed to send you directly to the deployment map - for deployment - but that was not panning out in the intended manner.  A chance that relogging in certain situations may cause significant delay was encountered during testing but the case that was causing the delay is believed to have been eliminated.  If it continues, please inform me.

I haven't played with tests and will probably fix those if they don't work as-is.

PR #459, opened 2020-05-21
Add some logging for suspicious HitMessages

Mazo:


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -6213,6 +6213,13 @@ class WorldSessionActor extends Actor
         case Some(hitInfo) =>
           ValidObject(hitInfo.hitobject_guid) match {
             case Some(target : PlanetSideGameObject with FactionAffinity with Vitality) =>
+              val hitPositionDiscrepancy = Vector3.DistanceSquared(hitInfo.hit_pos, target.Position)
+              if(hitPositionDiscrepancy > 7500) {
```
pschord, line 6217: Can we create a new section in the world config for "antihack" and start adding these kinds of variables in there? We need to eventually enable "enforcing mode" and be able to tune these without restarting the server.

Mazo: Yeah, that's a good idea. I presume the config loader doesn't currently support hot reloading?

PR #461, opened 2020-05-24
Update to Scala 2.13 & Akka 2.6.5

jgillich:
Most dependencies required updating for Scala 2.12 compatibility so I went ahead and updated all of them to the latest versions. Some of those are major version upgrades. Server works fine as far as I can tell, and I think there's only a single failing test. But there are tons of warnings (not sure how many of them are new).

If it is desired we could also go straight to Scala 2.13, but [scodec-akka](https://github.com/scodec/scodec-akka) is not available for that and its repo has been archived.

TODO:

- [x] Ensure all tests are passing
- [x] Figure out what is happening with the zone maps and drop the workaround


jgillich:
Failing test is `objects.ZoneTest`, maybe there's actually something wrong with the zone maps. But I'm completely lost in the Travis log, can someone spot where the failure is?

Fate-JH:
The failures for in the Travis build are exceptional.  Is this the "doesn't play well with 2.13?"

Mazo:
[scala213fix.zip](https://github.com/psforever/PSF-LoginServer/files/4676287/scala213fix.zip)

Patch file attached to fix build errors if you could apply it to your fork and push to the PR @jgillich 

I didn't look at any of the warnings beyond a cursory glance. Those will need attention at some point.

jgillich:
Nice work! I've committed it in your name (btw, if you PR my branch it will get included here as well).

jgillich:
https://scastie.scala-lang.org/2gTIxaX2TqiofoQU4kM6XA

It works if you switch to Scala 2.11, but it seems like using await this way is no longer supported. There were a bunch of changes to futures in 2.12, but nothing stands out to me as being responsible for this.

jgillich:
Somebody pointed out to me that this actually works in a REPL, even in Scala 2.13. I have no idea what's happening.

Mazo:
I found this issue: https://github.com/scala/bug/issues/10782

Which then points to this one: https://github.com/scala/bug/issues/9076

> Dave Gurnell (davegurnell) said (edited on Mar 1, 2017 10:35:25 AM UTC):
> Seth suggested I post this recreation here. Try this on a 2.12.1 REPL:
> 
> ```scala
> import scala.concurrent._
> import scala.concurrent.duration._
> import scala.concurrent.ExecutionContext.Implicits.global
> 
> // This works fine:
> val x = Future(42)
> Await.result(x, 1.second)
> 
> // This throws a TimeoutException:
> Await.result(Future(42), 1.second)
> ```


It's not an exact match for our issue, but it's pretty damn close.

jgillich:
They specifically talk about it only affecting the REPL. But that code you posted also reproduces the error in Scastie. But the fix was only for the REPL. I guess we can conclude this is a Scala bug then?

Would it be a big deal to run the map init synchronously until we figure out a solution?

jgillich:
Alright, so thanks to @Mazo this should be good to go. I've cleaned up the commits a little.

jgillich:
I also had a chat with some people on the Scala Gitter, contains some interesting information about the futures issue in case anyone is curious: https://gitter.im/scala/scala?at=5ecc1041f0b8a2053ab1bf5b

Mazo:
It's probably worth us getting a few people on a play test for this PR to make sure there's no weird side effects before merging.

PR #466, opened 2020-05-26
Add docker-compose file

jgillich:
Makes it possible to get the server up and running with all its dependencies via a single `docker-compose up`.


Mazo:
Nice. Looks good to me.

jgillich:
^ I realized PSAdmin is supposed to be TCP

PR #467, opened 2020-05-26
Refactor MultiPacketCollector#Bundle

jgillich:
Bundle was never called in any place other than BundleOption.
This refactors BundleOption into Bundle and removes BundleOption.
Besides being more efficient, it no longer has the possibility of
swallowing unrelated exceptions.


jgillich:
Update because I "fixed" a typo from "decalred" to "decalared" :facepalm:

PR #476, opened 2020-05-28
Interim

Fate-JH:
When an object with its own control agency leaves a zone that control agency, which was mostly likely owned by the zone, must be retired and it must be ushered into the new zone.  This mainly affects players and vehicles as they are the only entities that transfer between zones and have their own control agency.  Due to vehicles having utilities and some workflow making it almost necessary that all such objects be assigned their control agency upon creation (a form of zone introduction, even for sessile server objects), this case is actually much more common than anticipated.  This interim between controlled and uncontrolled was typically punctuated with the use of an akka "`null` test" - `ActorRef.noSender`.  That's actually a flagrant lie.  I did not know that `ActorRef.noSender` was equal to `null` when I assigned it the importance.

Under perfect conditions, a game object should not be visible in the zone if it is not already registered to that zone and, if necessary as a server object, is assigned a control agency.  Problems arise the latter case is assumed but not actually true, and the resulting of sending a message to `null` is not pretty.  The default value of the default "no-control" `Actor` must also be an `Actor`.  It must be something that exists while the entity is in an unstable state.  It must be realistically comparable in that one could determine that unstable state exists for a given entity.  The nature of how it operates beyond that is not as important as the fact that it exists.

___Features___
__`Default`__
`Default` evolved from `DefaultCancellable` and still contains the `Cancellable` global object to be used as the placeholder for all `Cancellables`.  It now also includes a defaulted `Actor` to be used as the placeholder for `Actor` references.  This is the aforementioned interim agency.  It starts a `ActorRef.noSender` (not a great compromise) and, when initialized at the start of an `ActorSystem`, equates itself to a indirect messaging service for `deadLetters`.

__`WorldSessionActor`__
The looping test for `case SetCurrentAvatar` has been expanded to include a wait period, two tries, and then a complete retry of the whole system.  There's no way out of the system once it starts, but this is much more thorough.  To facilitate the complete try, a new flag was created to allow the zone to forcibly reloaded.  (Prior to this, a zone would only load if it was detected as different from the zone where the user started.)

___Caveats___
1.  As mentioned in a rant on the Trello ticket, this method of default `Actor` destoys a bit of information - the actual destination of the message.  The best that can be done is sending a `DeadLetter` describing the default `Actor` as the destination.  That's not wrong, but it's also not correct.  COmpared to being completely wrong by being `null` and breaking something on message send, however, it's still better.
2. As mentioned above, it's possible to get stuck in a neverending loop of failing to get into a zone.  As unsacrastic as it sounds, this is a positive change from it being possible to get stuck in a neverending loop of failing to get into a zone.
3.  Timings and the number of acceptable attempts at one test or another may require.

___Addenda___
Moved to a PR comment because it's that significant,


Fate-JH:
**A solution that eliminates the need for the interim period altogether does exist**.  The entity with the control agency must have a global scope where all of these control agencies are fostered.  Only the entity's internal knowledge of which zone it is located and which zone's list of entities that entity can be found be updated.  This is actually how it was done for vehicles a long time ago (in server dev time), before this control agency was made to spurt from the occupying zone's specific `Zone`_Foo_`Actor` and then jump from one to another on transition.  I don't really want to do this for reasons of internal organization; but, it does pose a very lucrative opportunity.

Fate-JH:
Another two things I want to point out: updating seems to have causes me issues.  I can not longer run individual tests (due to a method not being found) and I can not attach breakpoints to code.  This makes solving simple issues such as the above tests failing a real hassle.

common/src/main/scala/net/psforever/objects/zones/ZonePopulationActor.scala
```diff
@@ -148,39 +182,53 @@ object ZonePopulationActor {
...
+    * @param playerMap the mapping of `Avatar` objects to `Player` objects
     * @param corpseList a list of `Player` objects
-    * @return true, if the `player` was added to the list;
-    *         false, otherwise
+    * @return a `Tuple` of two flags;
+    *         the first is whether the player was turned into a corpse or not;
+    *         the second is whether the player was found in the zone before being turned into a corpse
     */
-  def CorpseAdd(player : Player, corpseList : ListBuffer[Player]) : Boolean = {
+  def CorpseAdd(player : Player, playerMap : TrieMap[Avatar, Option[Player]], corpseList : ListBuffer[Player]) : (Boolean, Boolean) = {
```
Mazo, line 193: Is it worth converting corpseList to a TrieMap also, since it's passed around a fair bit, just to negate the chance that at some point it might be modified by multiple threads?

common/src/main/scala/net/psforever/objects/zones/ZonePopulationActor.scala
```diff
@@ -63,13 +82,28 @@ class ZonePopulationActor(zone : Zone, playerMap : TrieMap[Avatar, Option[Player
       }
 
     case Zone.Corpse.Add(player) =>
-      CorpseAdd(player, corpseList)
+      val (playerIsCorpse, playerInZone) = CorpseAdd(player, playerMap, corpseList)
+      if(playerIsCorpse) {
+        if(!playerInZone) {
```
Mazo, line 87: It's a really minor gripe, but you could combine the two if statements to reduce nesting

common/src/main/scala/net/psforever/objects/zones/ZonePopulationActor.scala
Fate-JH: Perhaps, but what would the key be?  It can't be any specific information related to the user's avatar because he is allowed multiple corpses to the same player avatar.

Mazo: The GUID of the corpse perhaps? As long as they're cleaned up properly there shouldn't be any conflicts.

Fate-JH:
Okay, weird bug.

1. Suicide
2. Respawn (corpse 1 created)
3. Suicide
4. Respawn (corpse 2 created)
5. Observe both corpses
6. Corpse 1 cleans up
7. Corpse 2 cleans up
8. Warnings in chat that keep repeating:
- can not decode an unmarshallable control packet `Control30`
- "<player> failed to load anywhere"

While I assume this has something to do with the loop, I'm a little puzzled.  I have no clue what the first message is as `CryptoSessionActor` is catching it coming from the client.  We do indeed have a `Control30`, but it has no decryption method.  The second message is an error message that happens if the target fails to be registered.

This only occurs if there are two or more corpses of the same person at the same time and it usually starts after the last one cleans up.

PR #479, opened 2020-05-30
QoL: Squad Fixes

Fate-JH:
The original squad leader's health and armor were not properly being scaled to the number range 0-64.  The field they were being pushed into is `7u` (0-127) so health, usually a maximum of 100 still fit, but armor value could go up to 200 and that would be a problem.  Squad member self-updates would have smoothed out the number but there are no updates until the second user makes the proto-squad into a proper squad and that is where the break occurs.  I'm not going to say that this solves all of our squad problems, but it does solve some of our squad problems.

The squad UI cards will now only display health and armor related to the member reported in them and not the vehicle in which that member is sat.

### DIFF
--- a/common/src/main/scala/services/teamwork/SquadService.scala
+++ b/common/src/main/scala/services/teamwork/SquadService.scala
@@ -2230,8 +2230,8 @@ class SquadService extends Actor {
     val leadPosition = squad.Membership(0)
     leadPosition.Name = name
     leadPosition.CharId = player.CharId
-    leadPosition.Health = player.Health
-    leadPosition.Armor = player.Armor
+    leadPosition.Health = StatConverter.Health(player.Health, player.MaxHealth, min=1, max=64)
+    leadPosition.Armor = StatConverter.Health(player.Armor, player.MaxArmor, min=1, max=64)
     leadPosition.Position = player.Position
     leadPosition.ZoneId = 1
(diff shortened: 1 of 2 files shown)


common/src/main/scala/services/teamwork/SquadService.scala
```diff
@@ -2230,8 +2230,8 @@ class SquadService extends Actor {
     val leadPosition = squad.Membership(0)
     leadPosition.Name = name
     leadPosition.CharId = player.CharId
-    leadPosition.Health = player.Health
-    leadPosition.Armor = player.Armor
+    leadPosition.Health = StatConverter.Health(player.Health, player.MaxHealth, min=1, max=64)
+    leadPosition.Armor = StatConverter.Health(player.Armor, player.MaxArmor, min=1, max=64)
```
Mazo, line 2234: From a quick glance at `StatConverter` it will likely fail on cloakers. They have zero armour and it has no check for `max > 0`, leading to a division by zero on the final line

Fate-JH: That's what the first case is for - `if(health < 1) 0`.

And I know it needs renaming.

Mazo: For some reason it hadn't occured to me than if "MaxArmour" is 0, "Armour" will also be.

As you were. 😑

PR #481, opened 2020-05-30
Zone specific property override config files

Mazo:
This is a fairly basic implementation of config based property overrides, and hopefully shouldn't require too much explanation.

For now, I've only added a few example properties since @NickPSF is intending to make changes to them, I'll let him add whatever he wants to them and we can merge that in later.

One thing of note is the dirty reflection hack in `ObjectClass.scala` to allow me to resolve object strings to their integer representation (e.g. lasher -> 430). I'm sure there's a better way of doing that...


Mazo:
One thing I should have mentioned: The overrides will only work clientside. No changing things like max health or damage values.

PR #483, opened 2020-06-05
XToolsV2

Mazo:
As the original XTools was written way back in 2016 by @SouNourS, and has since been hacked at by me to add UTF8 decoding and parallel processing it was due a bit of refactoring attention.

It's now been refactored to be a bit easier to follow the logic, moved into it's own subproject under `tools/decode-packets` (good idea @jgillich) and it's also a pretty useful tool for new developers to have access to when cloning the repo in order to view packet logs in a more human friendly output, hence including it within the main repo.

A copy of `xtoolspar` is included for historical reference.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,7 +108,8 @@ lazy val decodePackets = (project in file("tools/decode-packets")).
     libraryDependencies ++= Seq(
       "org.scala-lang.modules"     %% "scala-parallel-collections" % "0.2.0"
     )
-  )
+  ).
+  dependsOn(pslogin, common)
 
 // Special test configuration for really quiet tests (used in CI)
 lazy val QuietTest = config("quiet") extend(Test)
--- a/build.sbt
+++ b/build.sbt
@@ -49,6 +49,7 @@ lazy val commonSettings = Seq(
     "io.kamon"                   %% "kamon-apm-reporter" % "2.1.0",
     "org.json4s"                 %% "json4s-native" % "3.6.8",
     "com.typesafe.akka"          %% "akka-stream"   % "2.6.5",
+    "org.scala-lang.modules"     %% "scala-parallel-collections" % "0.2.0"
   )
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -109,7 +109,7 @@ lazy val decodePackets = (project in file("tools/decode-packets")).
       "org.scala-lang.modules"     %% "scala-parallel-collections" % "0.2.0"
     )
   ).
-  dependsOn(pslogin, common)
+  dependsOn(common)
 
 // Special test configuration for really quiet tests (used in CI)
 lazy val QuietTest = config("quiet") extend(Test)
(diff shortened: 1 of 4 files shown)


tools/decode-packets/src/main/scala/XToolsV2.scala
```diff
@@ -0,0 +1,157 @@
...
+import net.psforever.packet.PacketCoding
+import scodec.bits._
+import scodec.Attempt.{Failure, Successful}
+import java.nio.file.{Files, Paths, StandardCopyOption}
+
+import scala.io.{Codec, Source}
+import util.control.Breaks._
+import scala.collection.parallel.CollectionConverters._
+
+object XToolsV2 {
```
jgillich, line 13: This and the file should probably match the subproject? No matter what we call it.

tools/decode-packets/src/main/scala/xtoolspar.scala
```diff
@@ -0,0 +1,274 @@
...
+import net.psforever.packet._
+import scodec.Attempt
+import scodec.bits._
+
+import scala.io.{Codec, Source}
+import scala.collection.parallel.CollectionConverters._
+
+object Xtoolspar {
+
+  def main(args: Array[String]): Unit = {
```
jgillich, line 19: Do we have to keep this around? If we wanted to preserve a copy we could make it part of the git history, but there's no reason someone would use this over the improved version, right? Because when you run `sbt decodePackets/run` right now you get a `Multiple main classes detected` warning, I don't think you're supposed to have multiple mains per project.

build.sbt
```diff
@@ -102,5 +102,14 @@ lazy val common = (project in file("common")).
...
   settings(pscryptoSettings: _*)
 
+lazy val decodePackets = (project in file("tools/decode-packets")).
+  settings(commonSettings: _*).
+  settings(
+    libraryDependencies ++= Seq(
+      "org.scala-lang.modules"     %% "scala-parallel-collections" % "0.2.0"
+    )
+  ).
+  dependsOn(pslogin, common)
```
jgillich, line 112: I think `pslogin` is not necessary

tools/decode-packets/src/main/scala/xtoolspar.scala
Mazo: Interesting, I was running it through IDEA so wasn't getting those errors. The thinking behind keeping the original (for the moment) would be so you can run both and compare the outputs ensure there's not any unexpected differences between the two.

build.sbt
Mazo: Good spot, copypaste error.

tools/decode-packets/src/main/scala/xtoolspar.scala
jgillich: It lets you select which one you want to it's not a big deal, we can keep it for now.

tools/decode-packets/src/main/scala/XToolsV2.scala
Mazo: In light of the discussion below, shall we rename it once we're happy to get rid of the original `xtools`?

jgillich: Sounds good :shipit:

PR #485, opened 2020-06-06
Stamina / Implants

Fate-JH:
"How's your bioelectric level?"
"It's always at a hundred percent. I like to stay prepared."

Stamina management has been modified.  Stamina regeneration used to be limited to being on foot, but will now properly recharge when the driver or passenger of a vehicle.  The timing has been changed from a consistent pace - once every 500ms (2 per 1s) - to one that applies a benefit if crouched - once every 750ms when standing or sitting (4 per 3s), and once every 500ms when crouched.  Actually, the unit of measurement is "client update ticks" and that's ideally one tick every 250ms.

In changing stamina, implants were also modified somewhat.  The workflow for having new implants installed on a player and having old implants uninstalled from a player has been moved from `WorldSessionActor`, onto that player's control agency.  Implant timers now serve two purposes.  They are the initialization timer, first.  They are the drain timer, second.  Information about when the implant started its last initialization is retained to determine whether they should be consider initialized, and zero'd in case where they should start over.  They will restart their initialization process if they are disabled by fatigue before becoming fully initialized.  Uninstalled implants do not cause the client to crash on avatar setup.

___Features___
No new classes.  Keep reading.

___Caveats___
1. Passive implants do not have special cases which should re-activate them at the earliest convenient times, in the cases where they should havbe re-activated.  The one time Enhanced Targetting turns on after it first initializes seems to be client-driven.
2.  I _think_ implants turn off at all appropriate times but its worth checking more than once.

___Addenda___
1. From a technical perspective, stamina no longer updates just by changing the field on the player _entity_.  The player _control agency_ needs to be told directly to update.  Please do not place `Actor` messaging in the basic object classes, if possible.
2. From a technical persective, schedulers and `Actors` are distinct constructs.  If you start a scheduler within an `Actor`, and stop the `Actor`, the scheduler will continue to run.  Even without an `Actor`, schedulers will persist through a variety of situations where one might think it should not to matter anymore.  Please stop all schedulers when they are no longer necessary and rememebr the references passed along in a scheduler are only important while those references are valid.  In short, please don't forget that you sent `Player.ImplantInitializationComplete(...)` messages if the player control agencies of their destination is no longer running.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
+++ b/common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
@@ -40,7 +40,7 @@ class PlayerControl(player : Player) extends Actor
   private [this] val damageLog = org.log4s.getLogger(Damageable.LogChannel)
 
   // A collection of timers for each slot to trigger stamina drain on an interval
-  val implantSlotStaminaDrainTimers = mutable.HashMap(0 -> Default.Cancellable, 1 -> Default.Cancellable, 2 -> Default.Cancellable)
+  val implantSlotTimers = mutable.HashMap(0 -> Default.Cancellable, 1 -> Default.Cancellable, 2 -> Default.Cancellable)
   // control agency for the player's locker container (dedicated inventory slot #5)
   val lockerControlAgent : ActorRef = {
     val locker = player.Locker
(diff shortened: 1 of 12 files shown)


Fate-JH:
Three tests report errors but, due to changes to version updating, I need to wait until Travis has a go at them and I can examine the logs.  My breakpoints aren't always being tripped anymore and I can not run merely a single, sprecific test.  Even if I know the file, if I can not reason my way around the complaint just by looking at the code, I need to wait until another testing authority has a wack at it.

Mazo:
> The timing has been changed from a consistent pace - once every 500ms (2 per 1s) - to one that applies a benefit if crouched - once every 750ms when standing or sitting (4 per 3s), and once every 500ms when crouched.

Where has this come from? All rexo footage we found from live showed a consistent 500ms stamina tick, no matter what stance the player is in.

> The one time Enhanced Targetting turns on after it first initializes seems to be client-driven.

It also turns itself on from things like zoning, or moving between warpgates on the same continent.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
```diff
@@ -877,13 +880,16 @@ object PlayerControl {
...
-  def HandleDamage(target : Player, cause : ResolvedProjectile, damageToHealth : Int, damageToArmor : Int, damageToCapacitor : Int) : Unit = {
+  def HandleDamage(target : Player, cause : ResolvedProjectile, damageToHealth : Int, damageToArmor : Int, damageToStamina : Int, damageToCapacitor : Int) : Unit = {
     val targetGUID = target.GUID
     val zone = target.Zone
     val zoneId = zone.Id
     val events = zone.AvatarEvents
     val health = target.Health
     if(health > 0) {
+      if(damageToStamina > 0) {
+        target.Actor ! Player.DrainStamina(0)
```
Mazo, line 891: Are we passing 0 to force a stamina update to be sent to the client? If so, could we not move the actual stamina drain from `ResolutionCalculations` to here?

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
```diff
@@ -669,20 +702,194 @@ class PlayerControl(player : Player) extends Actor
...
+      if(wasFatigued) { //reactivate only if we were fatigued
+        player.Implants.indices.foreach { slot => // Re-enable all implants
+          player.Zone.AvatarEvents ! AvatarServiceMessage(player.Name, AvatarAction.SendResponse(Service.defaultPlayerGUID, AvatarImplantMessage(player.GUID, ImplantAction.OutOfStamina, slot, 0)))
+          if(!player.ImplantSlot(slot).Initialized) {
+            ImplantInitializationStart(slot)
+          }
+        }
+      }
+    }
+    player.Zone.AvatarEvents ! AvatarServiceMessage(player.Name, AvatarAction.PlanetsideAttributeToAll(player.GUID, 2, player.Stamina))
```
Mazo, line 735: Do all other players need to know about the stamina state of a particular avatar with `PlanetsideAttributeToAll`? I don't think there's anything that relies on you knowing how much stamina another player has

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Mazo: Actually, since this calls `StaminaChanged` that has a check to make sure the stamina has changed before calling `UpdateStamina`, so it's doing nothing?

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
```diff
@@ -760,6 +967,10 @@ object PlayerControl {
     target.Actor ! JammableUnit.ClearJammeredSound()
     target.Actor ! JammableUnit.ClearJammeredStatus()
     events ! AvatarServiceMessage(nameChannel, AvatarAction.Killed(player_guid, target.VehicleSeated)) //align client interface fields with state
+    //implants off
+    target.Actor ! Player.StaminaChanged(-target.Stamina)
+    target.Stamina = 0
```
Mazo, line 972: This feels weird, having to set the Stamina, but then also send off a notification that the stamina has changed to `PlayerControl`, and is probably very human error prone in that it's very easy to forget to do both things in other areas of the codebase.

That was my original thinking behind having the the `Stamina` setter also call out to `PlayerControl` by itself, we would never forget to do so.

Food for thought?

pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -4163,25 +4104,22 @@ class WorldSessionActor extends Actor
...
-        if(player.skipStaminaRegenForTurns > 0) {
+      if(deadState == DeadState.Alive) {
+        val regenTick = if(is_crouching) 2 else 3
+        if(player.skipStaminaRegenForTurns > 0 && upstreamMessageCount % 2 == 0) {
           //do not renew stamina for a while
           player.skipStaminaRegenForTurns -= 1
         }
-        else if(!isMovingPlus && player.Stamina != player.MaxStamina) {
-          player.Stamina += 1
+        else if(player.Stamina != player.MaxStamina && !isMovingPlus && upstreamMessageCount % regenTick == 0) {
```
Mazo, line 4120: From previous comment:

>Where has this come from? All rexo footage we found from live showed a consistent 500ms stamina tick, no matter what stance the player is in.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
```diff
@@ -51,102 +54,35 @@ class PlayerControl(player : Player) extends Actor
...
-          }
-        } else if (player.Fatigued && currentStamina >= 20) {
-          player.Fatigued = false
-          player.Implants.indices.foreach { slot => // Re-enable all implants
-            player.Zone.AvatarEvents ! AvatarServiceMessage(player.Zone.Id, AvatarAction.SendResponseTargeted(player.GUID, AvatarImplantMessage(player.GUID, ImplantAction.OutOfStamina, slot, 0)))
-          }
+        player.Stamina += amount
+        UpdateStamina()
+
+      case Player.StaminaChanged(changeInStamina : Int) =>
```
Mazo, line 80: Do we need both `DrainStamina` and `StaminaChanged` when they seem to do the same thing, except one has a check that the stamina has actually changed before calling `UpdateStamina`?

Mazo:
Implants aren't draining stamina at all when active

Fate-JH:
Yeah, I saw that on one of the previous comments, where `DrainStamina` became `+=` when it should have been `-=`.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: It doesn't matter if it's `PlanetsideAttributeToAll` or `PlanetsideAttributeToSelf` because it's only going to one channel - `player.Name`.  Also, no `if` check at the receiving `WSA`.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: We have do this for other state changes too.  For things like health updates, for example, I think you're not appreciating how the workflow is set up to occur at the tail end of a damage packet resolution.  If you set a player's health to zero at any time he doesn't just die.  You can send the update packet to immobilize him on the clients; but, to actually kill the player, the `Die` message has to be sent server-side.  Otherwise, the server will determine incorrect actions must be performed during respawn operations.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: I had thought so too but then I figured you probably have a good reason to want to update the value of stamina exactly when you do in `ResoCalc`.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: Wiki.
That's also code from before ... well, long before.  You guys had that video dive session after that mess with the first time we merged inventory things.  I don't know if you ever got back to me with stamina-related answers.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Mazo: Nothing more than it seemed like a logical place to put it and it worked, since the `Stamina` setter would call out to `PlayerControl` no matter where the stamina was updated

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Mazo: I did miss that it was sent to `player.Name`, but is it worth changing to `PlanetsideAttributeSelf` anyway to make the intent a bit clearer? Looking at it in the future might be confusing that it's sent to one player, but is using `PlanetsideAttributeToAll` implying that all players should get it

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Mazo: Actually, am I missing something here? Aren't we assinging `player.Stamina` twice in this example?

`Player.StaminaChanged` calls `StaminaChanged` which has the following line:
`val afterStamina = player.Stamina += changeInStamina`

So we're assinging the `changeInStamina` to `player.Stamina`, then assinging the result of that to `afterStamina`, but then also setting `target.Stamina` to `0` right after

Fate-JH: The player is already dead by accounts of line 965, so they won't regen stamina anymore after that point.  The do, however, need to propagate messages that turn their implants off and uninitialize them.  `StaminaChanged` is one message that can do that and it used to be connected to that mutator.  I'll eliminate the mutator call for `player.Stamina` on line 972 but the client update to 0 stamina on line 993 should remain.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: I do have a "I'll never get to it at this rate" ticket for doing stuff to the events messaging.

`...Self` also has an `if` check on the receiving end which is made pointless by ability to send to that player's specific channel.  There is a use for `...Self` - if that specific player, by GUID, should receive the message, but the player has respawned so it no longer matters they they get the message; or, if you only know the GUID of the player, but didn't track down their name.  That hedges against latency and timers.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: In any event, we should be able to modify entity fields without forcing client updates.

common/src/main/scala/net/psforever/objects/avatar/PlayerControl.scala
Fate-JH: If you want, I can exchange `DrainStamina(Int)` into a plain `UpdateStamina()`.

Mazo: Would replacing `Player.DrainStamina` calls with `Player.StaminaChanged` not do essentially the exact same thing?

Fate-JH: The calls to `DrainStamina` would still have to be changed to `StaminaChanged`, and with an negated value.  But, when it comes to the line `target.Actor ! Player.DrainStamina(0)` being changed to `ChangeStamina` ...
```
val beforeStamina = player.Stamina //50
val afterStamina = player.Stamina += 0 //50 + 0 = 50
if(beforeStamina != afterStamina) { //50 != 50, false
  UpdateStamina() //doesn't get called
}
```

Mazo: Right I see what you're saying now, I guess yeah, renaming `DrainStamina` to just call `UpdateStamina` makes more sense, especially since it's not really being used to drain stamina at all, just trigger an update.

Mazo: As an extra thought: Alternatively an optional boolean on `Player.StaminaChanged` to always send an update

PR #490, opened 2020-06-09
Graverobbing

Fate-JH:
Corpses should be safely lootable.
Messages were getting published to the original player's channel and not to zone channel.  That's why looting your own corpse worked but looting other peoples's corpses didn't.


NickPSF:
Looks fine on the CI server

PR #491, opened 2020-06-09
Reviving Players at a Distance

Fate-JH:
... should not be possible anymore.
I make no promises about anything else working.


NickPSF:
Looks good on the CI server

PR #492, opened 2020-06-09
Xtools v3

jgillich:
> * Use command line interface instead of hard-coded paths
> * Takes in .gcap files instead of parsed gcapy files
> * No longer needs temp files
> * Add pack config to allow system-wide installation
> * Slightly change output format from [C|S] -> [C|s] to C [<<<|>>>] S
> * Add some documentation to README

@Mazo

Also added two small but unrelated commits that aren't worth making separate PRs for, just don't squash merge. The changes to the docker image should reduce its size by a lot. The config path does change to something kind of bad `/usr/local/config` because sbt-pack doesn't output a proper folder structure, but I think nobody uses it at the moment so I/we can take care of that later.


jgillich:
To clarify on the output format, instead of:
```
#1 @ 0.111600s C -> S

or 

#1 @ 0.111600s S -> C
```
where the position of `C` and `S` is swapped, it renders as:

```
#1 @ 0.111600s C <<< S

or 

#1 @ 0.111600s C >>> S
```

where only the direction of the arrows changes. It's easier to read that way, I think?

tools/decode-packets/src/main/scala/DecodePackets.scala
```diff
@@ -0,0 +1,165 @@
...
+    val builder = OParser.builder[Config]
+
+    val parser = {
+      import builder._
+      OParser.sequence(
+        programName("psf-decode-packets"),
+        opt[String]('o', "out-dir")
+          .action((x, c) => c.copy(outDir = x))
+          .text("Output directory"),
+        arg[File]("<file>...")
```
Mazo, line 30: Does this work if you pass in a folder instead of individual files? I'd rather not list 655 files in one argument 😄

tools/decode-packets/src/main/scala/DecodePackets.scala
```diff
@@ -0,0 +1,165 @@
...
+     opts.files.foreach { file =>
+      if(!file.exists) {
+        println(s"file ${file.getAbsolutePath()} does not exist")
+        sys.exit(1)
+      }
+    }
+
+    opts.files.par.foreach { file =>
+      val parsedFile = s"gcapy -xa ${file.getAbsolutePath()}" !!
+      val out = new ByteArrayOutputStream()
```
Mazo, line 61: JVM will almost certainly run out of memory if you're processing a folder of all the available gcaps and holding the entire file contents in memory until the end to flush it. Some of the resulting files are in the 300-600MB range.

That was one of the reasons I used the temporary folder, as if anything went wrong during processing for that file you could just clear the temp folder (without having to dig through thousands of log lines to see which files were being processed at the time), then rerun XTools and it wouldn't reprocess already done files when combined with the check to see if the file already exists in the output directory

tools/decode-packets/src/main/scala/DecodePackets.scala
```diff
@@ -0,0 +1,165 @@
...
+
+     opts.files.foreach { file =>
+      if(!file.exists) {
+        println(s"file ${file.getAbsolutePath()} does not exist")
+        sys.exit(1)
+      }
+    }
+
+    opts.files.par.foreach { file =>
+      val parsedFile = s"gcapy -xa ${file.getAbsolutePath()}" !!
```
Mazo, line 60: The option to pass in pre-processed gcapy files would be nice addition. This actually wouldn't run on my machine because I run gcapy through a Linux instance using WSL, but do all my other dev through Windows.

tools/decode-packets/src/main/scala/DecodePackets.scala
jgillich: `*.gcap`? The shell should handle that just fine. Also more flexible since you can do stuff like `**/*.gcap` to convert everything in subfolders.

tools/decode-packets/src/main/scala/DecodePackets.scala
jgillich: Yea I'll add an option

tools/decode-packets/src/main/scala/DecodePackets.scala
jgillich: Do you you how many threads it uses? The default thread pool is equal to the number of processors, so if using that, it might be fine?

I wanted to test this by running it over all captures, but I'm getting a lot of errors. Could be related to my Python 3 port of gcapy though, not sure yet.

jgillich: How common are errors anyhow? I didn't really consider them to be normal tbh that's why it's just decoding everything every time.

Mazo: Yeah I believe it's equal to logical processors. Which on my machine is 16 threads. Worst case scenario taking the biggest files all running at the same time that's 2.53GB for the gcapy output, and 5.92GB for the decoded output, plus overheads.

tools/decode-packets/src/main/scala/DecodePackets.scala
Mazo: I guess the question there is does file globbing work in the various ways to run it? Could be run through sbt shell directly, a terminal (windows, or linux) or through IDEA

jgillich: Linux and Windows terminal all support globs. No idea if it works in IDEA, if it runs stuff in a terminal then it should. Sbt console is a Scala REPL, so no, I don't think so. The idea was to just install it to your PATH so you can call it via the terminal from anywhere.

tools/decode-packets/src/main/scala/DecodePackets.scala
jgillich: Ok so I ran it over all 363 captures and it consumed 3.7GB RAM at most. But I only have 8 logical processors so your estimation is fairly accurate.

So hmm, we could a) limit the number of threads to 4-8 (or make it dynamic: free GB / 500mb) b) have an option to set threads or c) go back to temp files.

I'm not a fan of writing to disk unnecessarily, and you usually don't gain performance by doing that either. But it's probably better than writing to swap..

Mazo: Errors in the actual decoding are fairly rare, but in the event they do happen being able to just exclude that file and not reprocess everything is nice since it does take a while. 

Primarily the errors come from outside factors, for example, the main error that I tend to come across though is actually disk space, on a space constrained SSD it's a bit of a pain to try and clear some space, and have to rerun all over it repeatedly until you have enough space for the whole data set

Mazo: At least with the BufferedWriter you can tweak the length of the buffer it uses, so perhaps increasing that somewhat will alleviate the majority of the performance cost of writing to disk more frequently. Could set the buffer to 16MB or something?

And the temporary files technically aren't temporary really. It's the final output file, just in a temporary location that gets moved so it's not as if the file is being written twice.

Mazo: Also limiting the number of threads seems like a bad solution. The process is primarily CPU bound more than it is disk bound, so that'll slow things down greatly.

jgillich: Yea you're right, revert to BufferedWriter will be up in a sec

jgillich:
Restored BufferedWriter. Also two new flags:
```
Usage: psf-decode-packets [options] <file>...

  -o, --out-dir <value>  Output directory
  -p, --preprocessed     Files are already preprocessed gcapy ascii files (do not call gcapy)
  -s, --skip-existing    Skip files that already exist in out-dir
  <file>...
```

README.md
```diff
@@ -148,32 +156,64 @@ sbt packArchiveZip # creates a single zip with resources
...
   - Join the #code channel and ask any questions you have there
 
 Chord is the lead developer and you can contact him on Discord as Chord or by email [chord@tuta.io](mailto:chord@tuta.io). Discord is preferred.
 
+## Tools
+
+### decodePackets
+
+The decodePackets program can be used to decode GameLogger `.gcap` packet captures. Requires
+[gcapy](https://github.com/psforever/gcapy) to run.
```
Mazo, line 196: Probably worth updating this with a quick note about preprocessed files, and a couple examples of the new flags for how to run it below?

tools/decode-packets/src/main/scala/DecodePackets.scala
jgillich: I also timed the performance and it's maybe 5 or 10 percent that is gained by doing it all in memory. Doesn't matter at all.

tools/decode-packets/src/main/scala/DecodePackets.scala
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+     opts.files.foreach { file =>
+      if(!file.exists) {
+        println(s"file ${file.getAbsolutePath()} does not exist")
+        sys.exit(1)
+      }
+    }
+
+    val tmpFolder = new File(System.getProperty("java.io.tmpdir") + "/psf-decode-packets")
```
Mazo, line 67: Since this will be buried away from the user in some system temporary folder is it worth clearing contents of that folder if it exists before starting a run, in the event the previous run failed for some reason?

I'm not sure what the default behavour of FileWriter is when the file already exists

PR #499, opened 2020-06-17
Disconnect

Fate-JH:
"Your account has been logged out by a Customer Service Representative."

Of the forms of disconnect known:
1. Menu Exit
This is the count down log-out.  10 for Sanctuary and Friendly regions, 20 for Neutral regions, and 30 for Enemy regions.  Once the user logs out at the completion of the countdown, character persistence is also dropped.  Quitting is technically a form of zoning so all conditions that would cancel Instant Action or a Recall to Sanctuary action will also cancel quitting.
2. Client window close, ALT + F4, process kill, crash, closing through the launcher
Client termination is instantaneous.  Persistence for the existing player character who was in the game world is maintained for the normal amount of time so re-logging with the same character before then should not banish the user back to Sanctuary.
3. Kicking
A form of disconnect.  Most of the aspects of this moderator-driven disconnect are similar to the previous entry, asides from the length of time being customizable and an inability to return to game with this same player character despite trying to log back in.  Player characters remain completely vulnerable, kicked out of any vehicle they previously occupied, and will lay dead for an extended duration of time if they are killed.

___Features___
No new classes or entities.  Everything piggybacks off of or clarifies existing functionality.

___Caveats___
1. A number of the string resources that would had to have been written for the interactions between different zoning methods and `DeadState` statuses don't seem to exist.  The expected references don't resolve.  Custom messages are implemented in their place.  This was already in practice for some cases for Instant Action and for Recall to Sanctuary.  Almost all of the hard-coded messages are employed when first receiving the message to zone or to quit.
~~2. Measures have been erected to disallow the cadaver of a kicked player to turn into a corpse but a guarantee for all cases requires additional testing.~~
~~3. Players who are kicked still count as population on a continent, unless filtered in all instances where population is counted.~~ @jgillich

___Addenda___
1. The command syntax for kick now supports a time field.  After declaring the individual to be ejected, a number of seconds of persistence can be entered.  The persistence monitor will ensure that the individual's player character will remain in the game for that amount of time to bar the user from re-entry.  Writing anything but a numerical time, as long as something is written, results in the persistence reverting to its default value.  There is no pardon, save lowering the time and logging out.
2. The time until use for medkits is presented as an integer again.
3. Stop kicking yourself by accident.

### DIFF
--- a/common/src/main/scala/net/psforever/objects/zones/Zoning.scala
+++ b/common/src/main/scala/net/psforever/objects/zones/Zoning.scala
@@ -10,7 +10,8 @@ object Zoning {
     val
     None,
     InstantAction,
-    Recall
+    Recall,
+    Quit
     = Value
   }
 
@@ -34,6 +35,8 @@ object Zoning {
     final val None = TimeType(20, "None")
     final val Enemy = TimeType(30, "Enemy")
   }
+
+  final case class Quit()
 
   object InstantAction {
     final case class Request(faction : PlanetSideEmpire.Value)
--- a/common/src/main/scala/net/psforever/packet/game/GenericActionMessage.scala
+++ b/common/src/main/scala/net/psforever/packet/game/GenericActionMessage.scala
@@ -26,9 +26,10 @@ import scodec.codecs._
   * 24 - message: you have been imprinted (updates imprinted status; does it?)<br>
   * 25 - message: you are no longer imprinted (updates imprinted status; does it?)<br>
   * 27 - event: purchase timers reset (does it?)<br>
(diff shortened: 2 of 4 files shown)


Fate-JH:
After conferring with Jakob, I went ahead and did it anyway.  Player characters who are kicked will be removed from the game session in ~60s; but, the length of time for which they will be kicked can be longer.  This is maintained entirely by the persistence monitor.

Additionally, due to the logout changes, it is no longer necessary to force players who are kicked into corpses (which would have entailed cleanup).  Most of the code in effort to catch those cases, however, remains in place.

Issue #482, opened 2020-06-04
!hack <faction> does not completely hack all of the buildings in a Caverns Zone


NickPSF:
!hack <faction> does not completely hack all of the buildings in a Caverns Zone

Mazo:
Seems fine on 61000 after #484

NickPSF:
@Mazo Still occurring in Adlivun (c3) for both AT Plants.

Mazo:
Resolved

PR #500, opened 2020-06-21
Fix ugd03 GUIDs + Sanctuary repair silos

Mazo:
What was a small bug assigning two of the equipment_terminal GUIDs in ugd03 ended up in a rather large refactoring of the AmenityExtractor code after I got fed up with it being a complete spaghetti mess after being adjusted from learning how different systems interacted with GUIDs.

It's by no means good, but it's better.

As a nice bonus, repair silos in sanctuaries are now also assigned GUIDs properly, and a small fix was added to allow them to be used since they have no owner, they also have no faction affinity.

Corresponding PS1Tools PR: https://github.com/GeekOfWires/PS1Ubr/pull/21


Mazo:
As a note, this should fix #482 properly this time.

PR #504, opened 2020-06-28
Correctly show ANT capacitor on zoning

### LINKED ISSUE #470: ANT - NTU Capacitor - When zoning an ANT with NTU, the NTU Capacitor display will show it empty when entering the zone, despite having NTU
When zoning an ANT with NTU, the NTU Capacitor display will show it empty when entering the zone, despite having NTU.

Mazo:
You should no longer need to dismount and remount to get the correct capacitor to show when going through warpgates.

One caveat is zoning seems to have an issue with packets being lost currently, but it seems to work going from z4 -> home3

Fixes #470

### DIFF
--- a/common/src/main/scala/net/psforever/objects/Vehicle.scala
+++ b/common/src/main/scala/net/psforever/objects/Vehicle.scala
@@ -209,6 +209,14 @@ class Vehicle(private val vehicleDef : VehicleDefinition) extends AmenityOwner
     NtuCapacitor
   }
 
+  def NtuCapacitorScaled : Int = {
+    if(Definition.MaxNtuCapacitor > 0) {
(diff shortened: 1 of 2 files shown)


Fate-JH:
Anything that's only apparent in the vehicle UI and not in the object create packet needs to be updated in the `HandleSetCurrentAvatar` function or else it will not update on things like `CSRWarp`.
I'm doing something to make certain the `HandleSetCurrentAvatar` function always gets called when working with vehicles and `CSRWarp`.

NickPSF:
Works on CI server, though was unable to test /warp

Mazo:
I did test /warp locally and it worked fine for both capacitor and shields.

PR #508, opened 2020-06-29
Persistence #1 featuring quill

jgillich:
Note: Tested and appears to work well, but it's a large change so we don't necessarily have to merge this right now. I have a lot more planned, so we could leave this open and I'll add stuff as I go. Up to you guys.

Configuration in quill uses the typesafe.config library. Since Akka also uses it, I wonder if it wouldn't make sense to dump the current config parser / ini config and use it for everything. The config library also supports environment variables, which is nice to have for container deployments. Maybe something we can consider in the future.

There are multiple places where there are fewer `ActionResultMessage.Fail(x)` codes used. I'm not sure what they stand for, but I've tried to keep them as they were. But the old implementations accounted for some situations that I don't think are possible, like an insert with `RETURNING` not returning any results (but also not failing).


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -9239,39 +9118,24 @@ class WorldSessionActor extends Actor
...
+    }
+
+    val result = for {
+      loadouts <- ctx.run(query[persistence.Loadout].filter(_.charactersId == lift(owner.CharId)).filter(_.loadoutNumber == lift(line)))
+      loadout <- loadouts.headOption match {
+        case Some(loadout) =>
+          ctx.run(query[persistence.Loadout].filter(_.id == lift(loadout.id)).update(_.exosuitId -> lift(owner.ExoSuit.id), _.name -> lift(label), _.items -> lift(items)))
+        case None =>
+          // FIXME id is long in objects but int in schema
+          val charId: Int = owner.CharId.toInt
```
jgillich, line 9131: A lot of IDs in the game objects are longs. Shouldn't really be necessary, do we make them int or do we make the db values longs?

Fate-JH: Character ID is always represented as 32-bits in the packets and you can't extract that into an `Int`.

jgillich: But `Int` is 32 bit..?

Fate-JH: `Scala` `Int` is signed.  `scodec` treats a stream of bits as unsigned.  It has to take a 32-bit number and push it into the datatype with the next highest bit length to preserve the value.
The `uint32` transcoder resolves into a `Long` value.

jgillich: Ah TIL there's no `UInt` in scala. Seems like postgres doesn't have it either, I'll make them longs in the schema when I'm done with migrations.

Fate-JH: If you wanted to write a 32-bit -> `Int` transcoder, you'd use `bool :: uint(31)` and manually factor for the sign and two's complement with an `xmap`.

jgillich: Right, could also do that, although I think a custom encoder would require a custom type as well. I'll look into it.

jgillich:
Migrations are done, they will automatically apply on startup. Caveat is that you can never go back, so we have to be cautious when rolling back things on the production environment (we do regular backups, right?).

Existing databases have to be initialized once using the [Flyway CLI](https://flywaydb.org/getstarted/firststeps/commandline) using the `baseline` command.

Mazo:
>(we do regular backups, right?)

Err..... @SouNourS ?

SouNourS:
Since it's "dev" server, no backup right now :(

jgillich:
Might be good to set that up, even if it's just backing up to local disk.

Thoughts on making account names case insensitive? I think the case sensitivity confuses a lot of people, especially since the server will just create missing accounts without feedback.

NickPSF:
@pschord

scp-mb:
Case insensitive account names is definitely a required change.

We'll have to manually comb through the existing accounts though, merge where safe to do so if we can confirm both accounts belong to one user, and work out what to do with the rest.

Mazo:
Case insensitive account names is definitely a required change.

We'll have to manually comb through the existing accounts though, merge where safe to do so if we can confirm both accounts belong to one user, and work out what to do with the rest.

jgillich:
I think I can do it in a way that is backwards compatible. New signups are all converted to lowercase, but logins check for the exact name first and then name.toLower.

pslogin/src/main/scala/WorldSessionActor.scala
Fate-JH: Er, actually, I forgot.  There are transcoders for signed numbers, I just never use them. The native `intL32` goes into an `Int`.
Most shorter fields utilize their full bit length and numbers are almost never negative.  In terms of character id, the only difference between using `Int` and `Long` is the absurd number of unique characters that can be created.

jgillich:
It occurred to me that we may not want to want to store all usernames lowercased in case we want to display them in their original form. This however means that only one match is actually password checked.

Given three accounts: tESt, Test, TEST

If you log in with `test`, you can hit any of them. But if you enter their exact names, you will get the right one.

But since it's no longer possible to create more than one of these moving forward, it should be fine.

jgillich:
Rebased. The last commit is reproducible with `sbt scalafmtAll` for those who don't feel like reviewing 200k changed lines.

jgillich:
Also got rid of unused imports and fixed usernames still being lowercased.

jgillich:
f5b35ee could use some testing. I noticed some weirdness regarding the address the server is bound to, for some reason I could only connect via my LAN address (the reason I never fully discovered though, the default appears to be the loopback address)? Then I noticed that the worldserver.Hostname setting only affects LoginSessionActor forwarding the address to the client and not the address the sockets are bound to. The change brings them in line.

jgillich:
@NickPSF Here it is, the `/capturebase` command in its full glory. Well, almost, the timer doesn't actually work yet. And sometimes hacking the whole continent takes two more than one attempt, but !hack probably had the same issue and the underlying cause should be somewhere else. And if you have time to test other chat commands to make sure I didn't break anything, that would be appreciated.

The chat handler has been greatly simplified into a single match. CSRZone and CSRWrap are still separate because I really don't want to move code back into WSA. But they're a good example of why we can't effectively break up WSA at the moment, they are still only partially separate because everything's reliant on shared local variables. What's worse are shared global variables, which is what we're doing with zones at the moment, and that definitely needs to be fixed for db persistence. There may be a solution here that could address both of these issues.

jgillich:
TIL resolving conflicts through Github's UI create a merge commit. Guess I'm not doing that again.

Anyway, would be nice to get this in before I have to rebase again because that's a bit of a PITA. =)

Fate-JH:
841 files changed!?

If this is the sort of formatting we're going with in the future, I _need_ you to share your IDEA linter rules with me.  I know I'm going to break at least four of them without trying. Especially the lack of space between the variable/function and the colon, and the same for adding that space after every conditional and loop operator.

jgillich:
I don't use IDEA, but the formatting uses scalafmt: https://github.com/psforever/PSF-LoginServer/blob/8ed62ceaaafcc9aeacb3dbf4bbeb867f7aa230c9/.scalafmt.conf

The settings are close to default, but we can definitely tweak them further.

IDEA supports scalafmt out of the box: https://www.jetbrains.com/help/idea/work-with-scala-formatter.html

jgillich:
For reference, scalafmt configuration options: https://scalameta.org/scalafmt/docs/configuration.html

I'm flip flopping a little on the align preset. `more` looks nice in some places but sometimes the inserted whitespace is a lot when you have variables with very long names. Maybe we should dial it back a bit and use `some`.

`maxColumn = 120` is also debatable. 80 is the default because that way you can fit two files side by side with a 1080p resolution. With 120, some characters at the end are cut off. But as everyone has a widescreen display nowadays and I like using that space. But maybe 100 is better, not sure.

jgillich:
Oh man, I just realized that [`getLocalHost`](https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getLocalHost()) doesn't return `127.0.0.1` but the public address of the primary interface instead.

So https://github.com/psforever/PSF-LoginServer/pull/508/commits/f5b35ee1e7b467b36a6134d98422d835dbccede7 changes the default to `127.0.0.1` unless `worldserver.Hostname` is set, in which case that is used for all ports (except psadmin). Does this make sense? I'm thinking that maybe there should be a listen address (usually 0.0.0.0) and then another setting for to use for forwarding to the world server (the public IP, what `worldserver.Hostname` used to be).

jgillich:
Ok, done. There is now a `bind` setting for the bind address and a `public` setting for the public IP / hostname. And I moved everything to typesafe.config. PSFCI needs an update for this, coming soon.

jgillich:
https://github.com/psforever/PSF-ServerCI/pull/2 is up.

build.sbt
```diff
@@ -47,6 +47,8 @@ lazy val commonSettings = Seq(
     "com.typesafe.akka"          %% "akka-protobuf-v3"        % "2.6.6",
     "com.typesafe.akka"          %% "akka-stream"             % "2.6.6",
     "com.typesafe.akka"          %% "akka-testkit"            % "2.6.6" % "test",
+    "com.typesafe.akka"          %% "akka-actor-typed"        % "2.6.6",
+    "com.typesafe.akka"          %% "akka-cluster-typed"      % "2.6.6",
```
jgillich, line 51: Oops didn't mean to include this. But I got some typed actors coming so we may as well.

jgillich:
I realized that naming the user configuration `application.conf` is not a ideal because that's the built-in file and configs have a override order (reference.conf < application.conf < user config). So I made the new name of the config `psforever.conf`.

Fate-JH:
SUDDENLY

PR #510, opened 2020-07-04
Deployment and NTU

Fate-JH:
"How does this system work anyway?"
> "Nanites."

"But why ..."
> "Nanites."

&nbsp;
The main purpose of this effort was moving code out from `WorldSessionActor`.  The details of this code involved the ability to tranfer nanites to and from the Advanced Nanite Transport, and the ability to deploy and undeploy vehicles, that being a necessary step for the nanite movement.

Somewhere along the way, it all went pear-shaped.

___Features___
__`TransferBehavior` and `TransferContainer`__
A hopeful abstraction of the functionality for an entity control agent that can induce transfer from one container object to another.  The main purpose is to keep track of whether that transferred resource is incoming (`Charging`) or being offloaded (`Discharging`) as it treats these as mutually exclusive states.

__`Ntu` and `NtuContainer`__
Ntu.  NTU.  Nanite Transfer Unit.
Defines a transferrable material as well a storage trait and messages to be used by ...

__`NtuStorageBehavior`__
Simple behavior mixin for managing stored nanites and for writing custom responses to different messages.  All control agencies attached to objects that storage and move nanites adapt these operations.  The affected include the following objects and their control entities: `ResourceSilo` and `WarpGate`, and required for `Vehicle` to allow for functionality of the Advanced Nanite Transport.  The difference between this mixin and `TransferBehavior` are that this focuses only on NTU exchange and cdoes not care about limiting to only incoming transfers or only outgoing transfers.

__`AntTransferBehavior`__
Behavior mixin that facilitates the transfer of nanites as a recourse from `ResourceSilo` entities into the Advanced Nanite Transport vehicle entity from the perspective of the Ant.  Operates the panel and particle animations.

__`DeploymentBehavior`__
This behavior mixin already existed, but has been enhanced with direct override opportunities to eliminate code that required callback to `WorldSessionActor`.  Those overrides are realized in `VehicleControl`, as well as the elimination of redundant functionality previously intersecting other parts of the code.  Clean-up of deployed vehicles has, especially, been streamlined to be centered around a single attempt at undeploying.  In case of having been destroyed, cleanup tried to invoke undeploying operations but instead fell back on a redundant case involved in the actual deconstruction of the vehicle.

___Caveats___
1. The layers of abstraction just for Ntu transfer originally were implemented for future expansion when building the NTU siphon and armor siphon weapon-arms for battleframe robotics, but I jumped the gun and presumed the kind of messaging pathways that would be used by the client in regards to those arms.  Obviously, this has been the first chance we've had to actually test the arms in any real environment since no resource silos or enemy vehicles existed in the Gemini virtual reality driving range.  Unfortunately, it was only after I started the "I forgot to check" tests for the weapon-arms that I remembered my effort in the previous battleframe robotics branch where they were not spawning correctly and I actually lacked clear-cut object create packet data for them.  Faced with this sethback, the code is being shipped as-is, since it works.
2.  Though moving deployment code out from `WorldSessionActor`, some deployment change code related to Routers remains in `VehicleService`.  For now, do not worry about it.
3.  When deploying, an ANT will always look for a warp gate to charge from, even if it is deploying in front of a resource silo.  The resource silo wll contact the ANT and ask to discharge into it, and that prompts the ANT to search for the resource silo.  The only reason this system works is that no resource silos exist within the sphere of influence of any warp gates.
4.  The amount of nanites being transferred from warp gates to the ANT is hard coded in the ANT's request.  The amount of nanites being transferred from the ANT to the resource silo is also hardcoded in the ANT's reply.
5.  Code format standardization has not been attempted for any new files or any changed files.  I am waiting on the other branch where _everything_ has been formatted to be merged first.


Fate-JH:
Need to make a modification with how `ResourceSiloControl` flags state changes.

PR #516, opened 2020-07-14
 ChatActor

jgillich:
This removes roughly 1k LOC from WorldSessionActor and moves
them to a new ChatActor. That was the initial goal anyway, but it
wasn't that simple. There was no clear location to put this new actor,
I didn't want to put it in pslogin since it isn't part of the "login server"
(and neither is WSA). But since the new actor would have to talk to
WSA and common does not depend on pslogin, I had a choice of
putting more actors in pslogin or putting everything in common. I
chose the latter.

ChatActor and SessionActor (formerly WorldSessionActor) now live
in `common/actors/session`. Since WSA also depends on other
actors in pslogin, most of the pslogin code was moved to either
common/login or common/util. PsLogin as the main entry point
remains in pslogin since having the main code compile to a library
has some advantages, and it will allow us to produce binaries
for distinct login/world servers in the future if desired. For a second
take, I'd suggest moving common to /src in the root directory.

This change is enabled by a new immutable `Zone` object that is
passed from SessionActor to ChatActor. Most of its members are
still mutable references, and the code at the moment does depend
on this being the case. Changes to the session object in
SessionActor are forwarded through a SetZone message to
ChatActor. As we split out more code into actors, we could
use EventBus or typed Topic's instead.

Also included is a reworked ChatService that was converted to a
typed actor and uses the built-in Receptionist facility for service
discovery. By receiving the session object from ChatActor, it can
be much smarter about who to send messages to, rather than
sending all messages to everyone and having them figure it out.
But as this session object is not updated, it can only use static
properties like player name and faction and not fluid properties
like position.

<hr>

Thoughts on the new file structure? As mentioned, I thought adding more stuff to pslogin doesn't make much sense since that module is way more than a login server already. And we have a lot of actors mixed in with `objects` where they don't really belong, that's why I think making a new `actors` package with all non-service actors is a good idea. I also stripped the World from WorldSessionActor since 90% of our code is technically part of the world server anyway.


Fate-JH:
Would you change "`ChatMessage`" back to `ChatMsg`?  It was named that way because that's the internally recognized name of the packet.

jgillich:
It seemed weird to me to have `ChatMsg` but also `ChatMessageType`. Use `ChatMsg` for both?

Fate-JH:
Well, one follows from the packet, the other just collects its purpose values.  And Chord probably found those by examining the decompiled string table for the client and making some kind of determination.  If you think it would serve consistency that we keep them the same.

jgillich:
Doesn't matter that much, I reverted it.

PR #517, opened 2020-07-14
Add command line interface

jgillich:
Comes with a flag to run flyway baseline automatically

@Mazo To set the bind address as an argument you'll now have to use `--bind 1.2.3.4`

Automatic baselining has to be enabled with `--baseline-on-migrate`. Migrations are always being run, but there's a flag to turn that off.


Mazo:
Both commands worked just fine for me, so I'd say that's a good enough test to merge.

PR #520, opened 2020-07-20
Maelstrom

Fate-JH:
Across the troubled time of base defense, people always need a Maelstrom.

___Features___
__`ChainLashMessage`__
Conveys the chain lash effect to the listed targets.

__`GenericObjectActionMessage`__
Catching this packet while holding a Maelstrom is evidence that the Malestrom's vortex whip has been discharged, but that no projectile was generated because no damageable target was struck.

___Caveats___
1. Chain lashing is a server-side calculation and can hit players through thin walls.  It collects every target within 5m, friend or foe, instructs clients to play the chain lashig effect on them, and deals damage to them.  Only targets players.
2.  The chain lashing effect is applied at the time of `HitMessage` packet arrival rather than actual damage-dealing.  It was purposefully done this way to take advantage of the locality of mass target selection.  The message could easily be distributed at the time of damage but it would be another packet per target.  It might be necessary, for the purposes of simulation, to move it into proper damage-dealing routines eventually.
3.  I believe the Maelstrom is doing way too much damage at the moment.  This excess is entirely the fault of the damage model for the vortex whip (primary fire) hijacking calculations for the chain lashing (secondary fire and tertiary fire).  To be fair, nothing about this weapon's damage model makes sense - what should be lash damage, or at least splash damage, is actually direct hit damage.  To fix this excess, a modification of the whole damage resistance model must be executed.  I was planning on doing that anyway.

___Addenda___
Calculating the length of time since the last recorded login time was throwing errors and that was frequently killing player selection.  Specifically, "`UnsupportedOperationException`: Cannot convert to Seconds as this period contains months and months vary in length".  This is a Joda-Time issue with the representation of months and extracting the number of seconds from an unspecified month.  If I'm not mistaken, character login times were stored as POSIX Long values before this change.  To quickly patch the issue, I just substituted the number of seconds in a day upon catching an exception.  This is just a patch, however, and an actual analysis of what is really necessary for this field should be performed.


pslogin/src/main/scala/WorldSessionActor.scala
```diff
@@ -10107,7 +10076,7 @@ class WorldSessionActor extends Actor with MDCContextAware {
     val faction = target.Faction
     target.ExoSuit = ExoSuitType.Standard
     target.Slot(0).Equipment = Tool(GlobalDefinitions.StandardPistol(faction))
-    target.Slot(2).Equipment = Tool(GlobalDefinitions.suppressor)
+    target.Slot(2).Equipment = Tool(GlobalDefinitions.maelstrom)
```
jgillich, line 10079: :point_up:

Fate-JH: And that's how you get a Maelstrom for testing.
Will be undone before merge.

Fate-JH: I modified the damage aspect of the damage resistance model system.  Now, instead of a package deal where things such as distance calculations, degradation due to distance, and damage type selection are all a function of the target, ONLY damage type selection is a function of the target.  That is to say: the target will tell the projectile what it is and what kind of damage it takes, and that's it.  Degradation models have been moved from the target to the projectile, into a system of stackable `DamageModifiers`.  A few objects now contribute modifiers to the final damage calculation and should be useful for modelling other dynamics.

For now, though, the Maelstrom.  The secondary and tertiary projectiles should calculate radial damage degradation properly.

PR #522, opened 2020-07-22
Persistence #2 Featuring Factions

jgillich:
Buildings will now persist their faction in the database. At least
that's what I want you to believe this change is.

What it actually is: A rework of InterstellarCluster and groundwork
for further reworks.

InterstellarClusterService: This is the old InterstellarCluster, but
as a service (it has always been one in secret). It was converted to
a typed actor and it now handles all spawn point requests.

ZoneActor: Basically ZoneControl, but as a typed actor. It's more of a
stub right now, the eventual goal is to have it own the `Zone` object
rather than the other way around.

BuildingActor: BuildingControl, but as a typed actor.

Also includes some minor improvements to ChatActor and sets
SupervisorStrategy.restart for all typed actors (which is the default
for classic actors, but not for typed actors - we may want to get more
sophisticated here in the future).

<hr>

Also included is ChatActor from #516 since this builds on top of it.


jgillich:
One test failure remaining, and I got no clue why. :confused: 

```
 [info] VehicleControlPrepareForDeletionMountedCargoTest:
[info] VehicleControl should 
[info] - if with mounted cargo, eject it when marked for deconstruction *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (500 milliseconds) while expecting 7 messages (got 0)
```

Fate-JH:
> One test failure remaining, and I got no clue why. confused
> VehicleControlPrepareForDeletionMountedCargoTest:
Assuming what I'm seeing it your PR is accurate to your plight:

Travis says there's a `NullPointerException` on line 570.  It's most likely `zone.Transport`.  Note that Travis also complains about this for the other test VehicleControlPrepareForDeletionMountedInTest which also use the zone's vehicle list actor, which used to be set up during a proper init message.

jgillich:
I'm staring at that. There's absolutely no way `zone.Transport` can be null, the default is `Default.Actor` and it's only set once in init using `context.actorOf` (which should also not return null..hopefully).

Although I wonder if this has got something to do with it:

Screenshot from 2020-07-23 03-04-48

The parameters match the signature exactly so this is very weird...

Mazo:
>There's absolutely no way zone.Transport can be null, the default is Default.Actor and it's only set once in init using context.actorOf (which should also not return null..hopefully).

Brace yourself.

image
image
image
image
image

Fate-JH:
Under normal circumstances, `Default.Actor` is initialized once near the start of `PSLogin`.

This was the least degenerate thing that I could think of where an actor could be set to a default that is still comparable, but also not `null`, and also not require I completely unravel the actor-on-its-object-in-its-actor pattern.  I mean, I will have to completely unravel that ouroboros of technical debt eventually; just never "today".

jgillich:
But if you don't make it "today", the day may never come. It's not like everything can be fixed in a day, but you can start today, slowly. :wink: At least that's what I'm doing.

And yea, I totally missed that the initialization doesn't actually happen without calling `Default.apply`. :man_facepalming:
I think there's a timing issue there, simply sending a message to the actor seems to have fixed the null error.

jgillich:
New PR at #527

PR #529, opened 2020-08-20
Aggravation

Fate-JH:
> In a hacked medapp, one hundred incendiary gel supply
> Infantry throwing loose y'all, getting splashed on the stairs
> Folks are screaming, set on fire
> It was so aggravating until the grenades started to explode
> I spy a cloaker in the green
> _(Burn, baby, burn)_
> Plasma inferno

___Features___
__`Aura` and `AuraContainer`__
Particle effects (colored smoke) that can be emitted by a target to indicate affliction with some sort of previous damage and ongoing damage over time.  At the moment, only infantry targets emit these effects.

__`AuraEffectBehavior`__
The logic by which particle effects can emitted naturally by a target for a specific duration.  When the duration is over, the particle effects are removed from the target.  At the moment, only infantry targets emit these effects.

__`AggravatedInfo` and `AggravatedDamage`__
The logic for simulating aggravation from a projectile impact - damage over time, potentially with an accompanying aura effect.  Aggravation is typically received as a form of `SplashHitMessage` but modifications to the data allow for both `Direct` aggravation and `Splash` aggravation.  Not much really distinguishes these two types of damage, however; whenever a projectile has both types represented, the same sort of data is featured in both.  As targets that are affected by aggravation and targets that are not affected by aggravation exist, a special collection of target validation checks, similar to those used for jammer grenades and proximity terminals, are employed.  Specifics about how each field actually contributes to damage calculations is actually not well understood.  For example, the field `vanu_aggravated` is set on the Starfire and Aphelion Starfire, both VS anti-air technology, but not set on the Comet or any other anti-air weaponry.  Information is stored directly on the projectile definition.

__`AggravatedBehavior`__
The logic by which aggravation is generated for any target.  The validity of the target is established once to make certain the initial damage is applied appropriately, and then the validity of the target is confirmed again to start the damage over time aspect.  Upon each tick of aggravation, calculations that affect the damage specific to its sort of aggravating effect are performed by a tailored copy of the original ballistic information being piped back into the target's damage logic pipeline.  The damage over time aspect can end naturally, when the canned logic runs out of application, or will end naturally when the target has been destroyed.  This continuing application of damage over time is referred to as "burning".

___Caveats___
1.  Information about how projectiles with aggravation information operate can not easily be gleamed from the packet captures.  Not only is this due to the format complexity of the captures causing damage information to bleed into one another, but no specific damage testing packet seems to have ever been generated for each weapon system.  The primary measurement of whether the damage being performed is sufficient is measured in the observed number of damage applications per shot and the observed number of shots to kill (STK).  Special thanks goes to Youtuber TheLegendaryNarwhal who hosts weapon demonstration videos in the virtual reality ranges for each empire, vintage 2012.
2.  The multipliers used to generate aggravated damage (damage over time) have been loaded into `DamageModifiers` stored on the projectiles that inflict the aggravated damage themselves.  External logic sets up fields regarding the projectile's resolution and its quality in order to determine which modifiers to apply at which stage of the process.  Additionally, since the metric for most projectile damage was perceived STK, and not everything about the transformations to the damage values based on their fields is knowable, a lot of the logic for calculating aggravated damage seems arbitrary.  It is arbitrary.
3.  Only infantry, vehicles, facility turrets, and one-man field turrets are susceptible to damage over time.  As mentioned above, only infantry display the aura effects conveyed by damage over time aggravation.
4.  The weapon systems affected by this mechanic are **Plasma Grenades**, weapons that use **Plasma Cartridges**, both fire modes of the **Dragon**, the **Starfire**, and the **Comet**.  Normal explosive damage should be unaffected.  Flak, while it was examined, is also unrelated as it uses a completely different mechanic to explain its damage model, and is unchanged.
5.  Being "aggravated" in the sense of using `AggravatedDamageMessage` seems to purely be an Infantry affair.  The packet is perfectly capable, on the other hand, of causing any person or piece of equipment to cry out in pain if applied.  It's freaky.
6.  ~~Short aura effects will cancel out the incomplete remainder of long aura effects.  I will probably fix this later.~~
7. Due to the way aggravation plays into the reporting system for damage, I chose to unwind the inheritance structure of the `Damageable` mixins and that caused a re-evaluation of the damge procedures for a number of objects that can be damaged.  Please try to break a lot of things during testing.

___Addenda___
1.  The Dragon is a beast if it makes contact.
2.  `DamageFeedbackMessage` was given a second glance in the course of analyzing the aggravation effect of damage over time but still remains unused.  It probably works as some combined form of `PlanetsideAttributeMessage` and `HitHint` depending on which player receives it.
3. Burn that mother down.


Fate-JH:
Suppressed aura effects in certain instances (Comet splash), the Starfire will burn through any valid target, radial damage calculation change.

PR #530, opened 2020-08-22
Implied Swap Item

Fate-JH:
The Beamer tried to escape while its host lay dying.  Where did the Beamer think it was going?
Didn't it know that that we'd catch it again?

___Features___
The client no longer crashes when transforming players into corpses (when using the previous reproduction steps).

___Addenda___
What happened to the free hand item when the player died before today?  It's a mystery.


jgillich:
Attempting to fix the test server...

PR #572, opened 2020-09-03
Spud Gun

Fate-JH:
The only charging weapon in the whole game!

Fire in quick succession for about 20-21 damage per projectile.  Hold down the trigger for about a second to reach a damage output of 95.  If done as soon as the full charge is ready, the shot will still only cost one point of ammunition.  After about one second of "charge", but without discharge, the weapon will begin consuming additional points of ammunition to maintain charge but without further damage increase ("overcharge"), and the actual discharge will still cost a point of ammunition on top of that.  Consuming all the ammunition in overcharge penalty before shooting will result in no weapon discharge (and the need to reload).

The spiker is currently set as the default weapon for the first pistol slot.  The first box of ammo has been made suitable for the Spiker (though that's unnecessary since you can always just get more from a terminal).  These changes will be undone before the PR is merged.

___Features___
__`ChargeDamage`__
Simple field for maintaining charge damage.  The full process for charging requires the current projectile define a `ChargeDamage` field.

__`ChargeFireModeDefinition`__
Simple fire mode definition to retain specific information about the charging process.  Was being prepared in this manner in hopes of eventually introducing `case class` modification but never fully developed.  The full process for charging requires the current fire mode be a `ChargeFireModeDefinition` fire mode.

__`DamageModifiers.ChargeFireMode`__
Calculation for producing damage information related to the charge.  The minimum damage output is always achieved.  Normal damage output is additional damage on top of that and is subject to charge modifications.  The full process for charging requires the current projectile to have this `DamageModifier` applied.

__`DamageModifiers.RadialDegrade`__
Changes to incorporate minimum damage radius for radial degrade.

__`FireModeDamageModifiers`, `StandardDamageProfile`, `SpecificDamageProfile`__
Dissolves the former class into anonymous instances of one or other of the latter.

___Caveats___
1.  The client will consume ammunition on its own once it has determined that the Spiker has reached full charge.  The server is supposed to back this up with packet data and its own timer to ensure that the ammunition value remains consistent but, due to any sort of communication issues, it's possible the spiker UI may begin disputing a single point of ammunition during the overcharge process.  The final ammunition value _will_ be consistent with the magazine.
2.  Overhcarge can allow you to shoot the weapon while it has zero ammunition remaining (showing in the UI).  This may or may not be proper weapon fire behavior.
3.  Technical nitpick.  Weapon timing log needs to be more unified in the future.  A weapon's internal understanding of when it was last fired is in nanoseconds.  The client-facing understanding of the weapon's fire cycle is in milliseconds.  I am uncertain whether either of this information is where it should be.
4.  Technical nitpick.  Fire mode information is not where I want it right now.  For example, fire modes that do not belong to shotguns do not need to know about their `Chamber` (pellet count) field, but it's required to be baked into the basic fire mode since ammo slots need to know about pellet count to conduct weapon discharge accountability.  Important to this update, fire modes that retain information about charge mechanics are mere "data holders".  The logic for charge is much more in `SessionActor` and `DamageModifier` than anywhere else.

___Addenda___
Please try to format enumeration values to be either each on their own line or all on one line.


Fate-JH:
I just thought of something.  Do you think the charge mechanic also applies to the damage radius?

PR #578, opened 2020-09-13
Lightweight Locker

Fate-JH:
The player's locker is a bit of deadweight.  Classically, the player carries it around - more or less in his back pocket - waiting to be opened through interaction with a physical `Locker` component (as seen near spawn rooms, medical rooms, etc.).  All while it is carried around it is loaded in the background, with every object in it consuming its representation in global unique identifiers for whichever zone the player occupies.  When switching zones, it would be registered and loaded, per respawn, and then it wqould be unregistered when the zone is left.  (If it was re-registered everytime the player respawned, then that's even worse!).  And if that physical `Locker` is never touched by the player, all of that equipment is left to weigh the server down.  

A lot of juggling has been implemented to ensure that the locker needs to be registered and unregistered as little as possible, and this update is in the same effort. To imagine, the traditional method provides that all game objects are registered to a single contiguous unique number system (broken up in pools of numbers, but still contiguous).  Each player's locker has all of its items listed within that same sequence, even if it takes 300 GUID's to represent 150 grenades per player!  This constant dent into the unique number system per zone is unnecessary when the objects don't even need to be immediately accessible to the client like ammunition in the player's inventory (to update the count in their holster icons).

To that end, the locker system now utilizes a fixed number of registered fake objects that are intended to indicate the GUID of potential locker-borne equipment.  The locker inventory object has its own unique number system that shares the same GUID's of the zone-registered fake objects.  All players see the same fixed range of GUID's in the zone unique number system but that GUID is redirected to that player's own locker for content.  Before an item is placed into a locker, it's zone-GUID is unregistered and, when inserted, the locker-GUID is registered.  Before an item is removed from a locker, it's locker-GUID is partially unregistered so that it can be granted a zone-GUID upon removal from the locker.  This means an item in a locker doesn't cause any extra strain upon the game environment.  Moreover, a player's client only knows that these items exist (depicts them) when that player opens the locker.  This not only conserves GUID's, and pushes load times for the contents of a locker to the start of the session, but also maintains synchronization as the items are moved.

To summarize:
- should improve player load times
- should allow for a greater availability of GUID's per zone
- each player can cart around an obscene number of items in his locker without impact

___Features___
__`SpecificNumberSource`__
Unlike `MaxNumberSource` (formerly, `LimitedNumberSource`), which works by accepting a maximum number and allocating all of the numbers counting up from zero, this source of GUID's accepts any group of numbers as GUID's to assign.  It feels like a `NumberPool` in that it utilizes a subset of numbers but, being a `NumberSource`, it possesses not only underlying architecture for key-object association but it also only requires that much architecture and not any more.

__`LocallyRegisteredInventory`__
A specialization of a `GridInventory` for the purposes of implementing the aforementioned `SpecificNumberSource` as context.  Equipment inserted are registered to the number source, and are unregistered upon removal.  Only used for avatar lockers currently.

__`LocalLockerItem`__
The object used for zone-to-locker unique identifier number referencing.

___Caveats___
The GUID range for the locker referencing is arbitrary and hardcoded for the moment, to be impvoed upon during the more general registration and globally unique number system overhaul update.  The range is 40150 inclusive to 40450 exclusive, just above the fake projectile objects.

___Addenda___
1.  Player locker content is not persisted?
2.  But why would you want to save 150 grenades into your locker?


src/main/scala/net/psforever/login/WorldSession.scala
```diff
@@ -504,6 +505,215 @@ object WorldSession {
...
+    * @see `GridInventory.CheckCollisionsVar`
+    * @see `GUIDTask.RegisterEquipment`
+    * @see `GUIDTask.UnregisterEquipment`
+    * @see `IdentifiableEntity.Invalidate`
+    * @see `LockerContainer`
+    * @see `Service`
+    * @see `Task`
+    * @see `TaskResolver`
+    * @see `TaskResolver.GiveTask`
+    * @see `Zone.AvatarEvents`
```
jgillich, line 566: Are these generated by your IDE? Seems quite excessive to me.

Fate-JH: By hand.  I ask my autoformatter to do as little as possible since I inevitably fight with it and, being "auto-", it will generally win through subversion.  The most I expect of it is to indent properly.

I come from a school of thought that goes, "If you use it, you document it."  If you wanted to boil it down, you could probably trim it to "@see Zone.AvatarEvents" to indicate that this function uses the avatar event pipeline and "@see TaskResolver" to indicate that we use the task resolution system.  For now, I'm just compiling the list of entires that will be refined during a "do much later" meta-task that is going back and correcting all the comments for formal API generation.

PR #588, opened 2020-09-25
_

Fate-JH:
A `MatchError` is when no patterns included in a `match { ... }` statement would properly parse the input.  Matching patterns can often feature guards, boolean logic that decides whether the given pattern is a closer match to what is wanted than another pattern of similar or same syntax.  Most of those could easily be written into a single case with an `if ... else ... `; but, there is nothing inherently wrong with using guards as long as one is careful not to assume that `None` can be the last ditch pattern in an `Option` match anymore.

The pattern `_` catches everything.


Fate-JH:
I can actually create a ticket with the title of "_" ...

PR #619, opened 2020-10-22
Powered

Fate-JH:
For major facilities, power originates from a generator receiving nanites from the NTU resource silo.  Amenities belonging to a facility that receive power behave differently than when without power.  The facility itself is different when with power than when without power.  For facilities without generators or NTU resource silos, power and nanites manifest from thin air.

There's basically a lot of testing that needs being done but the limit of characters I may simultaneously have in one private instance at a time and not have that instance degenerate is three.  Four is like a PowerPoint presentation.

___`Features`___
__`PoweredAmenityControl`__
A new superclass for `Amenity` control agenices that exhibit different operation depending on whether they had been informed they are receiving power.  The control agencies that inherit this class have their `receive` functionality finalized by it and must append their operations onto either of two paths - `poweredStateLogic` and `unpoweredStateLogic`.  Additionally, callbacks for the transition between powered states may be invoked.

__`BuildingActor`__
"Power" is an abstract concept that derives almost entirely from the application of the message `BuildingActor.PowerOn()` and `BuildingActor.PowerOff()` and an independent variable that keeps track of which state the facility is in.  Although any given entity can be assigned a power state, the facility is the only object through which power is naturally transmitted.  Messages about power states changing will occasionally be sent back from an amenity and administered by an amenity's owning facility, e.g., what a facility does for its generator.  Right now, only the generator has given this responsibility back to its facility; but, the generator is the only one that seems to require this responsibility in any major way at the moment.  (Projection: this will change when the main terminal and viruses are implemented.)

__`GeneratorControl`__
The generator is a unique entity that serves as a confused junction between the instigation of power and the supply of NTU.  In terms of NTU, the generator needs NTU for its auto-repair operations and that is also what determines its "powered" status.  For that reason, it is not a `PoweredAmenityControl`-type agency, though it accommodates the same sort of state logic.  It must also accommodate the state logic for NTU provision (`SuppliedWithNtu` and `NtuDepleted`) without negatively affecting existing auto-repair implementation.  Whether or not the generator is "not destroyed" also plays a part in the decision.  The generator is a lifecycle-heavy component that has a variety of changes it can undergo - online, offline, normal, critical, destroyed, etc. - and those states affect other aspects of the facility in which it is installed.  Notably, whether any other amenity in the facility is accessible or whether the facility is spawnable even if the spawn tubes are operational otherwise.  For these, it sends status reports back to its owning facility and lets the facility dispatch messages.  Previously, before "power", it handled all of the message dispatch by itself.

__`DoorControl`__
The relationship between opening doors and checking against IFF locks has been moved from a more external site (`SessionActor`) to an internal location (this class, obviously).  Moving the door opening logic like so allows it to be under the auspicies of whether or not the door is powered.  If not powered, the door will just open freely.  If the door is powered, the existence of an IFF lock will be probed, same as before.  In the end, the door should behave the same way as before, and should also behave correctly if the facility has lost power.  Additionally, it is no longer necessary to call back to the player who wanted to open the door to open the door, if it ever was.  (It never was.  The original purpose for doing this was to give the person who wanted to open the door an "initiative bonus" that allowed the door to open before anyone else who subscribed for the event.  Implementation is moving away from the "initiative bonus" angle.)

__`TerminalControl` and `ProximityTerminalControl`__
When power is lost, no transactions with the terminals occur.  Standard terminals do not resolve their submitted requests and terminals that operate based on proximity do not affect local targets.  It should still be possible to open terminals, if the "use" prompt for a visual terminal UI becomes available, but nothing can be done from it.  Opening terminals is still overtly controlled from the client-facing session actor rather than through terminal control agency so the only thing that deters the UI are messages dispatched about the status of the owning facility in general.

__`SpawnTubeControl`__
Spawn tubes are considered offline when they are unpowered, but that does not grant them any other specific benefits or penalties.  They behave mostly the same.  The owning facility for a spawn tube will send a message that disables or enables the spawnability of that facility regardless of the state of its installed tubes on power change.

__`PainboxControl`__
The painbox will not cause damage if the power is off.  It will only reactivate if the SOI has players.  (A generator can auto-repair itself as long as there is NTU available from the silo.)  Subsequent startup messages will never cause the painbox boundaries to be re-defined subsequent times.

__`ImplantTerminalMechControl`__
When power is lost, any mounted player is forcibly dismounted and can not re-mount.

__`FacilityTurretControl`__
When power is lost, any mounted player is forcibly dismounted and can not re-mount.

___`Caveats`___
1.  When the generator is destroyed while being supplied with NTU, it will explode in 10s and kill everyone within 14m.  It will not actually report being "destroyed" until such a time when it explodes.  If the generator is destroyed while under the assumption that NTU is unavailable, it will not explode.  Likewise, if the generator is queued to explode but NTU becomes scarce during the prep time, it will not explode.  In these cases, it will switch quickly to characteristics of being destroyed without any further fanfare.  This may not be Gemini Live behavior.
2.  ~~The capitol force dome and~~ linked benefits are not currently influenced by facility power state.  Capture mechanics and power are untuned and may interfere with each other.

___`Addenda`___
1.  The NTU command (`!ntu`) has been enhanced this was necessary for testing purposes and the change is useful enough to leave.  Originally just a means of jump-starting a zone's facilities with a range of NTU, the command can now target specific facilities by name, and can set an exact amount of NTU.  The syntax is `!ntu [facility] [value]` where value can either call out a facility by name ("dagon") or refer to the current facility ("curr[ent]"), if any and if that facility has a silo.  Additionally, the NTU value can be set from empty to full in intervals of 10% (0-10) or to an exact number (>10).  The original functionality of setting NTU within a certain range of values can still be employed by not listing a value; and, same as before, designating no target facility sets all facilities in the zone to a new NTU value.  Though setting facility NTU to a value between 0 and 10 is impossible, it's not really a problem?
2.  In the future, the opening door should try `ask`ing its corresponding IFF lock whether or not that door may open for someone.
3.  For some reason, `/zone` and `/warp` commands cause players that release after dying to be ejected from the world.  These ejection messages can stack up to three times per login.  This is a dire but preexisting issue?
4.  Maybe we can fix Gemini Live.  Have we tried turning it off and on again?

### DIFF
--- a/src/main/scala/net/psforever/actors/session/ChatActor.scala
+++ b/src/main/scala/net/psforever/actors/session/ChatActor.scala
@@ -7,7 +7,7 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors, StashBuffer}
 import net.psforever.actors.zone.BuildingActor
 import net.psforever.objects.avatar.{BattleRank, Certification, CommandRank, Cosmetic}
 import net.psforever.objects.serverobject.pad.{VehicleSpawnControl, VehicleSpawnPad}
-import net.psforever.objects.{Default, GlobalDefinitions, Player, Session}
+import net.psforever.objects.{Default, Player, Session}
 import net.psforever.objects.serverobject.resourcesilo.ResourceSilo
 import net.psforever.objects.serverobject.structures.Building
 import net.psforever.objects.serverobject.turret.{FacilityTurret, TurretUpgrade, WeaponTurrets}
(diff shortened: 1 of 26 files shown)


Fate-JH:
Power state and capitol force dome state should cooperate now.

PR #637, opened 2020-11-25
Yet Another Corpse Fix

Fate-JH:
It's that time again!

___Features___
Corpses!

___Caveats___
Corpses ...

___Addenda___
1.  `LoginActor` did not seem to be getting the correct IP information due to a loss of meaning for the numeric `sessionId` values.  They were replaced by the randomized string `connectionId`.  Despite this, everything seemed to be working.  Since it was part of a database call, however, I made certain the correct information was being proliferated.  Additionally, `sessionId` was repurposed into a counter for the number of login attempts or the number of session instances - separate numbers, now.  At first, I thought this might have something to do with the crashes upon release, the issue that I intended to fix; now, I don't know why this wasn't a crash for some other reason.
2.  Corpses?


Fate-JH:
Okay, coverage is abysmal; but coverage was always abysmal; and, everything else worked.

PR #652, opened 2020-12-23
Auto-Repair Tuning

Fate-JH:
NTU draining due to amenity auto-repair should be significantly reduced in three measures:
1.  global configuration value that modifies drain
 Check `application.conf` and `AmenityAutoRepair`.  Value is currently set to 0.5x.
2.  increased repair values
 Information from the "Structure Armor" page on the wiki (http://wiki.psforever.net/wiki/Structure_Armor) was used to calculate the non-ADB mending values that needed to be applied each time.  For almost all amenities, this value increased, also resulting in fewer calls to the NTU Silo.
3.  modified task queueing for the auto-repair operation
 Originally, auto-repair cycle was a queue of: Delay, Request (tx), repeat.  Responses (rx) were processed whenever they arrived.  The new order is: Delay, Request (tx), Wait, Response (rx), repeat, explicitly.  This doesn't actually change drain but, I would think, the perception of drain rate.

___Addenda___
~~1. The integration tests for auto-repair were moved back into the non-parallel testing group (which was also proven to work correctly).  The changes to the auto-repair operation above means the `StashBuffer` for `BuildingActor` will probably not overflow as easily as before, hence these tests should always complete now.~~
2. The `ResourceSilo` display was not behaving as expected.  One should note that filling the silo to 10 with an ANT on Gemini Live would result in the silo reporting as 9 immediately thereafter if even one thing in the base is not fully repaired.  ~~While the code can not be this efficient, it's easy to see that only a full silo should always be seen as a 10 and even a small drain should be seen as a 9.~~  To cover for this, 1 includes reporting for all NTU levels below 0.2x, doubling the length of time one can hang at the precipice.


Fate-JH:
After some experimentation, the integration tests and all other tests in the `server` folder are never run with `sbt test`.  They, in fact, can not be run except individually under the current circumstances; and, individually, their reporting is broken, as they never display as passed or failed and do not report about exceptions.

PR #676, opened 2021-01-23
Capture terminal logic refactor

Mazo:
The whole thing was one giant spaghetti mess, so I've finally taken a crack at making the logic a bit easier to follow and less spread out across a lot of loosely related files. There should be practically zero functional change in this PR, excluding the cavern map fix. If anything behaves differently during testing let me know.

`HackCaptureActor` is now responsible for the majority of the logic.

Things of note:
1) A new `CaptureConsoleAware` trait and `CaptureConsoleAwareBehavior` was added so that a `Building` can find it's `Amenity` objects that are interested in knowing when the CC state changes, for example wall turrets and implant terminals to kick any currently mounted players. I had intended this to just be a trait on the `Actor` but then filtering the list of `Amenity` using `isInstanceOf` does not work, as the actor is hidden behind an `ActorRef`. If there's a way to do this cleanly I'm happy to do so.

2) There's now the beginning of a `PlanetsideAttributeEnum` so that we can stop using magic numbers for the attribute type in `PlanetsideAttributeMessage`

3) A few files were moved around so they're grouped together, just files related to capture terminals and implant terminals currently.


src/main/scala/net/psforever/actors/zone/BuildingActor.scala
```diff
@@ -201,6 +205,22 @@ class BuildingActor(
         }
         Behaviors.same
 
+      case AmenityStateChange(terminal: CaptureTerminal, data) =>
+        // Notify amenities that listen for CC hack state changes, e.g. wall turrets to dismount seated players
+        building.Amenities.filter(x => x.isInstanceOf[CaptureTerminalAware]).foreach(amenity => {
+          amenity.Actor ! CaptureTerminalAwareBehavior.TerminalStatusChanged(terminal, data.get.asInstanceOf[Boolean])
+        })
```
Fate-JH, line 212: Please unwrap the optional value in `data` using a `match` statement.

Fate-JH (changes requested):
My only concern is the commented request in `BuildingActor`.  If everything else works, who am I to complain?

src/main/scala/net/psforever/actors/zone/BuildingActor.scala
Mazo: Amended

PR #678, opened 2021-01-27
LLUs. 🤯

Mazo:
https://www.youtube.com/watch?v=_J6-3l3hCm0

Surprise!

This has been a long time coming. This work was originally planned to be done a few years ago after capture terminals were functional, but it quickly spiralled into a rabbit hole of needing lots of complicated supporting functionality. What was originally just "implement LLUs" ended up with the following (not an exhaustive list either!):

- A c# reader for .ubr files to pull out model information and their nested entities
- A basic .mpo reader to pull out facility details
- Reverse engineering the entire GUID assignment system with some help from @NickPSF 
- An exporter for all map objects (AmenityExtractor)
- A crude map object to scala code generator (MapGenerator), that was eventually replaced by @jgillich to read directly from the JSON files

After refactoring the capture terminal logic **(note: merge #676 first as this builds on top of it),** it made it much easier to see where the pieces of the LLU logic would fit in, as such there is now a `CaptureFlagManger` that sits under each `LocalService` for a zone, and handles basically everything LLU related, with messages passing primarily between `HackCaptureActor` and `SessionActor`

@Fate-JH This is my first attempt at turning the decompiled functions from the client into a working packet encoder/decoder for `CaptureFlagUpdateMessage`. Comparing with `HotSpotUpdateMessage` was a huge help to see the pattern of how to handle it, and I think I've got it correct but if you could pay extra attention to that it would be appreciated

I'm 100% certain there will be edge cases in this I've not accounted for, however it should be *mostly* functional.

Known issues:
- Dropping a llu while mid air will leave it hovering. Nothing we can do about this right now.
- Technically you can sit on top of a warpgate and shoot people inside it without the bullets disappearing, as you can't enter the warpgate bubble fully but when above you are slightly within it but not weapon locked. This is highly unlikely to happen in a realistic scenario.
- You can drive a vehicle whilst holding a LLU if you drop the LLU, queue a vehicle at a terminal then pick the LLU back up before getting seated in the vehicle
- Taking a LLU into water doesn't do anything


ScrawnyRonnie:
Just to keep things in one place so they don't get lost in Discord chat I'll put some other findings here:
- If a player carrying the LLU enters a vehicle as a passenger and the vehicle is destroyed, this causes the LLU to become invisible to the defending empire.
- Hacked LLU base does not show the hack timer on the map when double clicked to see base info. It just says "Control Room: Hacked"

src/main/scala/net/psforever/objects/definition/converter/CaptureFlagConverter.scala
```diff
@@ -0,0 +1,25 @@
...
+      CaptureFlagData(
+        new PlacementData(obj.Position, obj.Orientation, obj.Velocity),
+        obj.Faction,
+        obj.Owner.asInstanceOf[Building].GUID.guid,
+        obj.Target.GUID.guid,
+        millisecondsRemaining
+      )
+    )
+  }
+}
```
Fate-JH, line 25: Please add a trailing newline.

src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlag.scala
```diff
@@ -0,0 +1,59 @@
...
+    val obj = CaptureFlag(GlobalDefinitions.capture_flag)
+    obj.Position = pos
+    obj.Orientation = ori
+    obj.Target = target
+    obj.Owner = owner
+    obj.Faction = faction
+
+    obj
+  }
+}
```
Fate-JH, line 59: Please add a trailing newline.

src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagDefinition.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import net.psforever.objects.serverobject.structures.AmenityDefinition
+
+/**
+  * The definition for any 'CaptureFlag'
+  * Object Id 157
+  */
+class CaptureFlagDefinition extends AmenityDefinition(157) {
+  Name = "capture_flag"
+}
```
Fate-JH, line 11: Please add a trailing newline.

src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagSocket.scala
```diff
@@ -0,0 +1,32 @@
...
+  def Constructor(pos: Vector3)(id: Int, context: ActorContext) : CaptureFlagSocket = {
+    Constructor(GlobalDefinitions.llm_socket, pos)(id, context)
+  }
+
+  def Constructor(tdef: CaptureFlagSocketDefinition, pos: Vector3)(id: Int, context: ActorContext) : CaptureFlagSocket = {
+    val obj = CaptureFlagSocket(tdef)
+    obj.Position = pos
+    obj
+  }
+}
```
Fate-JH, line 32: Please add a trailing newline.

src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagSocketDefinition.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import net.psforever.objects.serverobject.structures.AmenityDefinition
+
+/**
+  * The definition for any 'CaptureFlagSocket'
+  * Object Id 450
+  */
+class CaptureFlagSocketDefinition extends AmenityDefinition(450) {
+  Name = "llm_socket"
+}
```
Fate-JH, line 11: Please add a trailing newline.

src/main/scala/net/psforever/services/local/support/CaptureFlagManager.scala
```diff
@@ -0,0 +1,282 @@
...
+      TrackFlag(flag)
+
+      taskResolver ! CallBackForTask(
+        TaskResolver.GiveTask(GUIDTask.RegisterObjectTask(flag)(socket.Zone.GUID).task),
+        socket.Zone.LocalEvents,
+        LocalServiceMessage(
+          socket.Zone.id,
+          LocalAction.LluSpawned(PlanetSideGUID(-1), flag)
+        )
+      )
```
Fate-JH, line 114: Do LLU's really not have fixed GUID's?  That's crazy.  They're as much an amenity of their facility as their socket is.

src/main/scala/net/psforever/services/local/support/CaptureFlagManager.scala
```diff
@@ -0,0 +1,282 @@
...
+      case PlanetSideEmpire.VS => "VanuSovereigncy" // Yes, this is wrong. It is like that in packet captures.
+    }
+  }
+}
+
+object CaptureFlagLostReasonEnum extends Enumeration {
+  type CaptureFlagLostReasonEnum = Value
+
+  val Resecured, TimedOut = Value
+}
```
Fate-JH, line 282: Please add a trailing newline.

Fate-JH (changes requested):
Additionally, I would merge the capture terminal PR before this one, and then re-evaluate the changes, as this LLU branch was obviously built on top of the capture terminal branch, for obvious reasons.

src/main/scala/net/psforever/services/local/support/CaptureFlagManager.scala
Mazo: Nope, at this point we've accounted for pretty much all fixed GUIDs in every zone in the amenity extractor script.

Fate-JH (approved):

Fate-JH:
All that's left is to clarify the two issues @ScrawnyRonnie mentioned.  We can do stuff with LLU's in water later since that's going to be a bit more involved.

Mazo:
LLU will now be dropped before deleting the player avatar that is seated in the vehicle. @ScrawnyRonnie 

As a bonus, I've wired up the generator state to the lattice benefits, and fixed an edge case where facilities may not update their lattice benefits properly

PR #683, opened 2021-02-02
Fix vehicle inventory unloading on mounting

Mazo:
Yay, race conditions.

### DIFF
--- a/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
+++ b/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
@@ -119,10 +119,7 @@ object MiddlewareActor {
     packet.isInstanceOf[CharacterInfoMessage]
   }
 
-  /**
-    * `KeepAliveMessage` packets are bundled by themselves.
-    * They're special.
-    */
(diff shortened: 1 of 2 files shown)


ScrawnyRonnie:
Testing seemed to have just 1 minor bug. When I had either 1 ammo box or the maximum amount that would fit in the trunk, it caused the trunk's ammo to show as 1. Not sure if the amount of ammo per box matters for this. In my case the maximum I could fit was 1500 ammo. Firing once and reloading made it show the correct amount. Did the same thing for solo and passenger/gunner vehicles, but again a minor thing because the trunk never showed empty as it had before and fixes itself on reload.

nwuollett:
@Mazo What's the status on this? Need more testing?

NickPSF:
@Mazo What's the status on this? Need more testing?

Mazo:
@NickPSF Not sure if the issue @ScrawnyRonnie mentioned is related or a completely different issue. I didn't come across it when I was testing, might be worth merging and raising as a separate issue if it continues?

NickPSF:
@Mazo Seems fien to me!

PR #694, opened 2021-02-10
Test fixes / tweaks

Mazo:
This should fix some of the flaky tests and speed up some of the water related ones.

There's still a failure with `VehicleControlInteractWithWaterTest` however, and I have no idea why. It runs perfectly fine when ran on it's own, but in a batch of all tests will fail even with greatly exaggerated timeouts for the expected message. Even running it in a smaller batch seems to work

@Fate-JH If you have any ideas on the above let me know.

### DIFF
--- a/src/test/scala/objects/PlayerControlTest.scala
+++ b/src/test/scala/objects/PlayerControlTest.scala
@@ -757,6 +757,7 @@ class PlayerControlDeathSeatedTest extends ActorTest {
 class PlayerControlInteractWithWaterTest extends ActorTest {
   val player1 =
     Player(Avatar(0, "TestCharacter1", PlanetSideEmpire.TR, CharacterGender.Male, 0, CharacterVoice.Mute)) //guid=1
+  player1.Definition.UnderwaterLifespan(suffocation = 3000, recovery = 3000) // Override the default 60s down / 10s recovery so the tests don't take forever to run
   val avatarProbe = TestProbe()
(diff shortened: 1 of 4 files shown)


src/test/scala/objects/PlayerControlTest.scala
```diff
@@ -757,6 +757,7 @@ class PlayerControlDeathSeatedTest extends ActorTest {
 class PlayerControlInteractWithWaterTest extends ActorTest {
   val player1 =
     Player(Avatar(0, "TestCharacter1", PlanetSideEmpire.TR, CharacterGender.Male, 0, CharacterVoice.Mute)) //guid=1
+  player1.Definition.UnderwaterLifespan(suffocation = 3000, recovery = 3000) // Override the default 60s down / 10s recovery so the tests don't take forever to run
```
Fate-JH, line 760: `player1.Definition.UnderwaterLifespan(...)`
As with my far far ealiertests involving a singleton classes is evidenced, changing a single source class for one test changes it for every test.  All tests involving a drowning player will endure this 3000ms / 3000ms time limit.  Given the reliability of actor tests, depending on the order they are implemented, it's possible that a test whose constituent actors may receive less attention than hoped may be affected by these updates times.  In other words, it'll drown faster too.
`GlobalDefinitions` can be like a singleton to all tests in that respect.

NickPSF:
This one still in flux?

PR #749, opened 2021-04-06
Cargo vehicle Mounting and Gating Fix

Fate-JH:
Cargo vehicle mounting and vehicle gating with cargo vehicles should work properly again.  Please do double check the gating of the passengers of a cargo vehicle for me.

Don't think of it as fixing something that was broken, think of it as accommodating something that should have never worked in the first place.


Fate-JH:
Once-cargo vehicles should now be restored as proper cargo vehicles if they fail to escape the loading bay region of the cargo bay within a generous, sufficient amount of time.
If the vehicle is deconstructed during loading, unloading, or while loading or unloading, it's actually one of the error messages that catches the issue in the act.  Since that causes no problems other than a logged anomaly, it should be fine to leave it like that.

PR #751, opened 2021-04-07
Flaky Facility Faction Fix

Fate-JH:
Here's the story so far:
When the server starts up, all of the zones start up.  Upon each zone, every building is created.  Buildings also have a separate flag that indicates whether a supply of NTU is available, `true` by default since only major facilities have resource silos to vary the power state (normally).  When a facility's amenities start, a resource silo may get constructed also will also gain a control agency.  That control agency is eventually passed a "start-up" message that initializes the silo to its proper functionality and is supposed to return a message to the building control agency indicating the NTU status upon receipt.  Resource silos start with no NTU in stock.  It's an override and code that runs after normal zone setup that distributes an artificial NTU value (max).  While all this is going on, somewhere else in the code, the last-known faction affiliation of the building is being polled from the database and will be assigned to the building as long as it is powered and believes it has a supply of NTU at the time.  **As long as it believes it has a supply of NTU at the time**.

SO.  The problem we are looking at goes something like this: the facility has completed being set up and the resource silo is sent a "start-up" message; the database is queried about the base's faction affiliation; and, the NTU of the facility is artificially set.  The silo reports to the facility that it has no NTU first, so the facility goes into no NTU mode.  The database reports the facility's faction affiliation second, but it fails to gets applied because the facility has no NTU.  The silo has its NTU artificially set third, so the facility goes into "has NTU" mode.  This is a very specific order of things that should go right going wrong, and that results in a facility previously captured by one faction at the end of one run of the server suddenly being neutral at the start of the next run of the server.

There's a long-winded way to solve it and make certain everything is sequenced properly with the faction of the database call being applied last but I completely eschew that noise and just set the start NTU to silos to max.

It works, doesn't it?

### DIFF
--- a/src/main/scala/net/psforever/actors/session/SessionActor.scala
+++ b/src/main/scala/net/psforever/actors/session/SessionActor.scala
@@ -6823,7 +6823,7 @@ class SessionActor(middlewareActor: typed.ActorRef[MiddlewareActor.Command], con
         //silo capacity
         sendResponse(PlanetsideAttributeMessage(amenityId, 45, silo.CapacitorDisplay))
         //warning lights
-        sendResponse(PlanetsideAttributeMessage(silo.Owner.GUID, 47, if (silo.LowNtuWarningOn) 1 else 0))
+        sendResponse(PlanetsideAttributeMessage(silo.Owner.GUID, 47, silo.LowNtuWarningOn))
         if (silo.NtuCapacitor == 0) {
           sendResponse(PlanetsideAttributeMessage(silo.Owner.GUID, 48, 1))
         }
--- a/src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagSocketControl.scala
+++ b/src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagSocketControl.scala
@@ -1,0 +1,20 @@
+// Copyright (c) 2020 PSForever
+package net.psforever.objects.serverobject.llu
+
+import akka.actor.Actor
+import net.psforever.objects.serverobject.affinity.FactionAffinityBehavior
+
+/**
+  * An `Actor` that handles messages being dispatched to a lattice logic unit (LLU) socket.
+  * Actually does nothing lol
+  * @param socket the socket entity being governed
+  */
+class CaptureFlagSocketControl(socket: CaptureFlagSocket)
+  extends Actor
+  with FactionAffinityBehavior.Check {
+  def FactionObject      = socket
+
+  def receive: Receive = {
+    case _ => ;
+  }
+}
--- a/src/main/scala/net/psforever/actors/session/ChatActor.scala
+++ b/src/main/scala/net/psforever/actors/session/ChatActor.scala
@@ -355,6 +355,8 @@ class ChatActor(
                           val soi2 = building.Definition.SOIRadius * building.Definition.SOIRadius
                           Vector3.DistanceSquared(building.Position, position) < soi2
                         }
+                    case Some(all) if all.toLowerCase.startsWith("all") =>
+                      session.zone.Buildings.values
                     case Some(x) =>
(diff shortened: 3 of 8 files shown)


Fate-JH:
According to logging, there are five instances where the initial message from a silo to its facility does not get received.  In short, "Message [net.psforever.actors.zone.BuildingActor$SuppliedWithNtu] to Actor[akka://PsLogin/user/system-default-actor#-1390451820] was not delivered. [1] dead letters encountered" and this is a consistent ten facilities.  I wish to explore why which instances generate this problem and rectify them before this PR is closed.

src/main/scala/net/psforever/actors/session/SessionActor.scala
```diff
@@ -6823,7 +6823,7 @@ class SessionActor(middlewareActor: typed.ActorRef[MiddlewareActor.Command], con
         //silo capacity
         sendResponse(PlanetsideAttributeMessage(amenityId, 45, silo.CapacitorDisplay))
         //warning lights
-        sendResponse(PlanetsideAttributeMessage(silo.Owner.GUID, 47, if (silo.LowNtuWarningOn) 1 else 0))
+        sendResponse(PlanetsideAttributeMessage(silo.Owner.GUID, 47, silo.LowNtuWarningOn))
```
Mazo, line 6826: Is sending a boolean going to work for `PSAM`?

Fate-JH:
At first, my assumption was that some facilities might have not been ready to accept the initial setup messages of the silos but that turned out to be inacurate.  What was really happening was the very reason I had constructed `Default .Actor` in the first place!  LLU sockets, called `CaptureFlagSocket`s on our side, were never assigned a control agency and, though they have no need to know about power or NTU, all amenities in a facility are to receive those messages in case they are powered.  In fact, all objects of type `PlanetSideServerObject` should be expected to have a (working?) control agency.  Supplying that resolved the log messages.

src/main/scala/net/psforever/actors/session/SessionActor.scala
Fate-JH: Check out the `implicit` `WorldSession.boolToInt` function.

src/main/scala/net/psforever/objects/serverobject/llu/CaptureFlagSocketControl.scala
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2020 PSForever
+package net.psforever.objects.serverobject.llu
+
+import akka.actor.Actor
+import net.psforever.objects.serverobject.affinity.FactionAffinityBehavior
+
+/**
+  * An `Actor` that handles messages being dispatched to a lattice logic unit (LLU) socket.
+  * Actually does nothing lol
```
Mazo, line 9: > All amenities in a facility are to receive those messages in case they are powered. In fact, all objects of type `PlanetSideServerObject` should be expected to have a (working?) control agency

It's probably worth adding something along those lines as a comment to this, so in the future we know why it exists but does nothing

Alternatively, can we just skip over `PlanetSideServerObject` that have a null `Actor`?

Fate-JH: Technically, it has the `Default.Actor` that bounces messages sent to it into `deadLeaves`, not `null`.  If it were `null`, that part of the server would explode whenever a message was dispatched to it.  The purpose of `Default.Actor` is to avoid that explosion and to indicate that something's actor was not started correctly.

In the documentation for `PlanetSideServerObject`:
"An object layered on top of the standard game object class that maintains an internal `ActorRef`.
"A measure of synchronization can be managed using this `Actor` as a 'controlling agent'."
I can add a comment that "this actor only exists to comply with the contract for `PlanetSideServerObject` entities" but that seems more excessive that the informal "lol".

PR #756, opened 2021-04-12
Quick fixes

Mazo:
Fixes a potential LLU crash
Also fixes Sobek (and other dropship center) CC door lock


jgillich:
LGTM. I was going for a different solution because we're storing the connections in a `door -> lock` map when it should be a `lock -> door` map, but this is probably the only case where it matters so this fix is good enough.

jgillich (approved):

jgillich:
Not sure what is going on with Github actions so :shipit:

PR #768, opened 2021-04-17
Leave Squad Service by Character Id

Fate-JH:
The original error message indicated a failure due to a character id that could easily be resolved into a `Long` value, and that would have become a `ClassCastException` anyway (`String` to `Long`).  My guess is that the exception might have been raised by a bad value lookup and, therefore, I have expanded a number of those lookups along the paths that would have lead to the exception message into `get`/`match` statements, for safety.  Logging of the stack trace for such an error was also included, as only the console would have known what went wrong before.  Debug-level logging has actually been raised to debug-level logging.


Fate-JH:
Fixed the filters so the log should log again.
Sorry, guys.  Sometimes I'm an idiot most of the time.

PR #775, opened 2021-04-17
Fix terminal hack clearing

### LINKED ISSUE #759: Terminals - Advanced Hacking - Factions - Enemy terminals can get into a state where they appear hacked to the player, but cannot be used
Enemy terminals can get into a state where they appear hacked to the player, but cannot be used

jgillich:
Fixes #759


jgillich:
Tested and it works so I'm merging, please yell at me if this is wrong :smile:

PR #784, opened 2021-04-21
Fixes From 20210421 Logs

Fate-JH:
- chat will explode if the data in `CMT_DESTROY` can not be cast to a number
  - why can text data be included in `CMT_DESTROY`?
- loadouts will contain entities like "Telepad" and "BoomerTrigger" but will not reconstruct "Telepad" or "BoomerTrigger" under the current server protocol
  - nothing wrong with writing it to the database as long as you remember that it's there


Fate-JH:
- sending messages to actors before they have been assigned
  - it's actually ill-advised to initialize `ActorRef`s to `noSender`; I apologize for having done this myself

Fate-JH:
- `SocketActor` would try to kill an actor that was never listed in the first place
   - it is best practice to check that a value exists before using it

Fate-JH:
- password field can be empty (`None`)
  - should we alow that to bounce off the database, or stop it before trying?
- squad elements being forced to draw before poreprly setup
  - probably caused by failure to `SquadService` correctly; pay attention to deaths during squad join?

PR #797, opened 2021-04-26
Load Vehicle

Fate-JH:
Originally, the purpose was to diagnose a chat message that indicated vehicles failing to be added to the same zone multiple times, indicating overactivity.  The solution for that was rather simple, though at first I pursued it in the course of vehicle loading messages between the client-facing session and the zone vehicle service.  In the course of those changes, the `VehicleSpawnPad` construction process now deals in introducing the vehicle into the zone's vehicle management queue directly and uses the responses from that process to determine when to issue client loading packets.  The real solution to the original problem was merely not re-submitting the vehicle to the zone management queue during avatar re-loads.

Included is some logging changes, including a case that doesn't complain about writing over unresolved projectiles if the overwrite happens beyond that old projectile's normal lifespan.  This case may need mild expansion in the future.


Fate-JH:
I think I've resolved the issue of ANT's depositing NTU when not deployed by thoroughly confusing them.

PR #802, opened 2021-04-30
Vehicle Spawn Pad QoL

Fate-JH:
The vehicle spawn pad had been, for some time, missing some minor features, mainly proper messages and events.  For example, the maximum distance the would-be-driver can flee from the side of the vehicle spawn terminal was never being enforced.  As long as sufficient delay existed in between the current order being fulfilled and one's own order, a player could go all the way across the continent and be brought back to the spawn pad by the will of the nanites when their time comes.  Now the distance is only about 35m.  If the player should stray onto the spawn pad during a vehicle's creation, where they should have died and an informative message be displayed, no death would occur.  Now, near instant death.  Player conditions such as what key items they are carrying and whether they are in transit, things that previously were localized to the server-facing session, now travel with the player's server-side entity.

Ordering a vehicle proceeds much like normal, with the exception of the messaging that is caused by the process, and the inclusion of the ability to exchange vehicles orders so long as the player are already in the vehicle order queue.  The second order overwrites the first and the player does not lose their place in the queue.

Ways to test this PR involve ordering vehicles and everyone but the person whose order is currently being handled doing something that would sabotage the order - fly away, try to zone away, succumb, change armors, mount the seat of a different vehicle, destroy the vehicle order terminal, etc..  Consequently, some people must also tread onto the spawn pad itself during vehicle creation to die a few times.  (Ritual sacrifice will help this PR succeed.)

___Features___
__`SpecialCarry`__
All of those special key items that will never be able to grace an inventory icon.  Only the lattice logic unit is the only holdable thing at the moment but messages exist for the monolith units and the vanu modules.  No word on the rabbit ball.

__`VehicleSpawnReason`__
For when you die for being on the vehicle spawn pad.

__`SpecialEmp`__
Previously, the generation of a server-side electromagnetic pulse was confined to speial functions and definitions on the `Zone` object.  Now, they are mostly located here.  Remember that this currently exists to allow projectiles with jammering properties to affect boomer explosives through walls, to balance out the fact that boomer explosives can inadvertantly bleed damage through nearby perpendicular walls.

Additionally:
__`VehicleSpawnControlRailJack`__
The kill box of the vehicle spawn pad is executed as if it were a "server-side explosion", but using `VehicleSpawnReason` as a justification.  The region of explosion is a cuboid 1m into the trench and at least as wide and long enough to cover the entirety of the pad, and as high as to clear a few meters above it.  As with other explosions, all damageable targets that find their way into the blast region are affected.

__`Zone`__
"causeEmp" and "cause Explosion" have been replaced with a generic "serverSideDamage" function that accepts a wider variety of  customization in order to deal a valid form of damage to (specific) targets.

___Caveats___
1. `Cuboid` was an attempt at resolving the polyhedron shape of the vehicle spawn pad for use in reconstructing the profile for the server-side damage.  It eventually was proven unnecessary when an easier solution existed and its functionality is currently untested.  In the future, with refinement, it may be useful for representing damage profiles of more boxy features, ones not as suitable for the cylinders that are currently utilized by most entities, the `apc` vehicles for example.  It will not be leveraged for now.
2. Players will die from being on the spawn pad during vehicle creation even when in Sanctuary, even when a Spectator or a GM.
3. Although almost all messages are now based on the client's string literals (`@`), the message that informs that one's vehicle is blocking the spawn pad from starting its next order is still custom.

___Addenda___
1. A fix for a significant stickiness side-effect of a previous "players can not exit or bail just-spawned vehicle too early" fix that caused unrelated players to get stuck inside everything.
2. A fix for a minor issue with vehicles drowning with weird seating arrangements.
3. A fix for a minor issue with the system for interacting with capture consoles that would fail to reach a default case.

### DIFF
--- a/server/src/test/scala/actor/objects/VehicleSpawnPadTest.scala
+++ b/server/src/test/scala/actor/objects/VehicleSpawnPadTest.scala
@@ -13,6 +13,7 @@ import net.psforever.services.vehicle.{VehicleAction, VehicleServiceMessage}
 import akka.actor.typed.scaladsl.adapter._
 import net.psforever.actors.zone.ZoneActor
 import net.psforever.objects.avatar.Avatar
+import net.psforever.objects.serverobject.terminals.Terminal
 
 import scala.concurrent.duration._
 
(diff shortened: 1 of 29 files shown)


Fate-JH:
Out of draft, into review stage.
Point out all my flaws.

Fate-JH:
Active orders can be cancelled by user activity up to the point where the ordered vehicle is spawned and attached to the spawn pad lifting platform.  After that, it exists.  Deal with it.

PR #831, opened 2021-05-22
No Second-class Facilities

Fate-JH:
Except for bunkers.

Minor over-logging where a facility without an LLU socket was described as "not a facility".


Mazo:
👍

PR #840, opened 2021-05-28
Deployable Behaviors

Fate-JH:
There are three big changes to be addressed.  

Firstly, the process of constructing, deconstructing, or adjusting ownership of deployables has been shifted away from `SessionActor`, where it was formally handled almost entirely and instead pushed onto the deployable itself, with inquiry and feedback requested directly of its owning player, if any.  The deployable, in this sense, has been that much more independent from a player's perspective (client session).  The process itself has not changed much, save where it must request help from the owning player to check against combat engineering limits, and how it no longer has to juggle a zonal timer to manage its deconstruction.

Secondly, there is no more distinction between `SimpleDeployable` and `ComplexDeployable` between different types of deployables.  To be quick about it, `ComplexDeployables`'s were very independent and handled a lot of their complicated internal operations.  `SimpleDeeployables`'s require a lot of hand-holding just to accomplish basic needs but also didn't have too many needs.  With the change, there are just "`Deployables`" now, which are based on `ComplexDeployables`, as expected.  Admittedly, most deployables were of the `Complex`-variety and the sole representative of a `Simple`-type was the `TelepadDeployable`.  (Originally originally, there were a number of `SimpleDeployable` representatives including the TRAP, but a previous update converted most of them.  `TelepadDeployable` objects were kept as-is during the aforementioned update because they are involved in a more complex teleport process.)  The consequences of this was a disjoint damage resolution branch that had to calculate and route through `LocalService` then apply itself back in `SA`.  Additionally, because `SimpleDeployables` and `ComplexDeployables` had different levels of autonomy and only the latter was granted a control agency, they had to be split up whenever common behavior was intended to be induced.

Third, the aforementioned Router-Telepad relationship which allowed for teleportation has also been mostly relocated from `SA`'s influence and has become a messaging exchange between a Router vehicle and a `TelepadDeployable` deployable.  There is no longer a third party governing the teleport setup; and, without the need for an independent timer to manage deconstruction of the deployable, the process has been mostly simplified down to one that occurs between the `TelepadDeployable` and the `InternalTelepad`.  Teleporting between the two is still conducted directly in `SessionActor` since it maintains a brief variable check and message dispatch.  No need for expensive activity.  Animations related to teleportation are also local to `SessionActor`.

___Features___
__`DeployableBehavior`__
The primary driver for deployable construction and deconstruction operations.  A reorganization / dumping ground for the procedures that used to cycle in and out of (`World`)`SessionActor` in the process of dealing with deployables.  Some aspects of deployable management remain rooted to `SA`, mainly those that can be found at the end of service, but much of those were unavoidable.  Management of deployable ownership is also important to the lifecycle enforced by `DeployableBehavior`.

___Caveats___
All caveats related to the original deployable PR and any updates should still be valid.  This PR was intended to be a pure refactor, not a feature change.

___Addenda___
1. According to the game logs, at various times, entity registration would consume and produce numbers whose registration data could not be recovered.  After a bit of testing, two sources were narrowed.  The first is overly-enthusiastic medkit consumption, handled elsewhere.  The second is the management of `BoomerTrigger` entities, paired with `Boomers`.  Due to the various ways that `Boomers` could be removed from the game world, no more than three seperate checks for the clean-up of their respective `BoomerTrigger` existed.  It was entirely possible for more than one of these clean-up cases to be execute, resulting in the same entity having two ongoing but racing unregistration tasks.  One would succeed; the other would fail; occasionally, both failed because the. "incorrect" one completed first.  Destruction of the `BoomerTrigger` is now managed by the `Boomer` in cases where the `Boomer` explodes, separated from the mine and managed by the session when being eliminated by the player, and this streamline eliminates the multiple executions.
2. Reports of undamageable `TelepadDeployable` entities were being submitted.  Now that this sort of entity has damage control similar to most other entities that can be damaged, this issue should be resolved, hopefully.
3. In theory, the code is devised such that deployables can be constructed without an owner.  This would come with the expected caveats, such as `BoomerDeployable` entities not coming with companion `BoomerTrigger` entities; but, the act of creating an ownerless deployable is not employed in the current code, so the full extent of the functionality of this mode is untested.
4.  I am of two minds regarding the switch from a zone-level deconstruction support service and a per-deployable deconstruction cycle.  On one hand, separating it unto each entity allows for an easier management (it makes the calls easier and more readable).  On the other hand, separating it unto each entity means that each entity has its own timer taking up some memory.  The only thing that can be said wityh any confidence is that this matches how vehicles decay and are deconstructed.


Fate-JH:
Did not even realize the router override was baked into this PR.  We probably should isolate the overrides files so they only travel with prod server build, unless otherwise intentionally included.

PR #866, opened 2021-06-13
Fix publishing docker containers on push (tagged with sha for all pushes, master is tagged with sha, master, latest)

Mazo:


Fate-JH:
The error for this PR's test "**Publish Docker Image / docker (pull_request)**" occurs when trying to "**Build container image**": 

"Error: buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = unexpected status: 403 Forbidden".

Would this be a showstopper, for some feature?

Mazo:
Yeah I need to work out what's going on there. I mentioned it briefly in our developers channel

PR #871, opened 2021-06-16
Separate Vehicle Controls

Fate-JH:
All vehicles now initialize a control agency that is better tailored to the specific operation of that said vehicle.  That said, most vehicles still use the generic agency and the specialized ones are only special because of that one degree of difference that makes the vehicle unique.  As there were (should be) next to no changes for any vehicles that were already functional, this effect emphasizes how many duties the default control agency performs.

This all seems underwhelming right now but it will become much more important when working with battleframe robotics (I hope).

___Features___
__`AmsControl`__, __`AntControl`__, __`RouterControl`__
The specialized control agencies for the advanced mobile spawn, the advanced nanite transport, and the Router, respectively.

__`DeployingVehicleControl`__
Asides from being a parent class of the former three specialized control agencies, the specialized control agency for the Switchblade and the Flail.

__`CargoCarrierVehicle`__
The specialized control agency for the Galaxy and for the Lodestar.

__`ApcControl`__
The specialized control agency for the armored personnel carrier vehicles - the Juggernaut, the Vindicator, and the Leviathan.  Possesses a charging capacitor that, when filled, allows the driver to cause the vehicle to emit a 15m electromagnetic pulse.

___Caveats___
1. ~~The electromagnetic pulse of the APC's is purely a radial calaculation.  It will pass through both walls and floors.  The dome calculation for explosives does exist but contains a slight mistake that would make it not work as well as expected for game entities as massive as the APC's.  The calculation oversight will be fixed in a different PR.  For now, just~~ humble yourself in the realization that letting enemy vehicles cavort about the courtyard and the perimeter of your facility is a poor idea anyway.
2. Charging time for the APC capacitor is based on numbers found in the ADB, which seem to coincide with approximate timestamps and other numbers found in the first time event captures for the APC's.  Arguments for changing the charge interval will be accepted but must be accompanied by disputing video evidence.

### DIFF
--- a/server/src/test/scala/actor/objects/AutoRepairIntegrationTest.scala
+++ b/server/src/test/scala/actor/objects/AutoRepairIntegrationTest.scala
@@ -14,7 +14,6 @@ import net.psforever.objects.serverobject.deploy.Deployment
 import net.psforever.objects.serverobject.resourcesilo.{ResourceSilo, ResourceSiloControl}
 import net.psforever.objects.serverobject.structures.{AutoRepairStats, Building, StructureType}
 import net.psforever.objects.serverobject.terminals.{OrderTerminalDefinition, Terminal, TerminalControl}
-import net.psforever.objects.vehicles.VehicleControl
 import net.psforever.objects.vital.Vitality
 import net.psforever.objects.vital.base.DamageResolution
 import net.psforever.objects.vital.damage.DamageProfile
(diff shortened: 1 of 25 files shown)


Fate-JH:
The calculation in question has been modifed and the modification reports promise.  I also ran physical tests to ensure that the previous conditions were still supported properly.  The APC EMP should now have the same dome damage range as the mines.  Still travels through walls, but not through the ground under its wheels.

Okay, get the garbage truck on its side and it'll probably still affect targets through the floor until it rights itself ...

PR #932, opened 2021-09-09
Collisions

Fate-JH:
I've struggled with how to write this introduction and realized later that I'm writing more about the pain about writing the code to simulate pain from collision rather than the code that simulates pain from collision.  I suppose there's nothing but pain in going forward anyway ...

A collision represents the interaction between one's character and the environment - which includes amenities - and one target directly interacting with another target - which excludes amenities.  The only consequence of this limitation is that the game doesn't think crashing into facility turrets and vehicle terminals can cause them damage, and that's fine by me.  Valid targets that can interact in collision are therefore limited to players, vehicles, and deployables.  The environment deals damage but is never considered a target so it has no need for being represented.  Taking damage from collision includes: moving into targets too quickly, having other targets moving into your position too quickly, and by quickly changing one's altitude far too quickly.  That final distinction requires that bailing from vehicles - for both passengers and for cargo vehicles - properly be respected and protected.

Entity on entity collision is, in fact, incredibly difficult to read from the packets because the entity taking collision damage will only take damage "out of the blue".  Video evidence can be found, but is inconclusive.  No explicit connection is made between the one who inflicted the damage - the only who actually receives the damage packet - and the other entity that just receives some damage.  Only a small subset of the packets discovered are useful because an immense number of variables about the encounter should probably need to be known to measure what kind of response to give.  At the moment, going full vanilla in these updates like I really want to do is going to be far too time consuming for the scope of the feature and may not be possible in some cases.

Go play in traffic and tell me how it goes.

___Features___
__`VehicleSubStateMessage`__
Undriven vehicles can now be pushed around and will respect their proper positioning on different clients.  While this may solve the "pilot bailing vehicle hangs in midair" problem in some fashion, our workaround for the cluttered airspace is too secure to allow for it at the moment.  A different update will play with the possibilities.

__`GenericCollisionMsg`__
The primary mechanism by which the client communicates what it considers a collision that causes damage.  The fields were reorganized so that the fields of the same entities that partook in the collision are adjacent to one another, rather than the natural staggering of the fields.  The first entity is always "you".

__`CollisionData`, `CollisionZData`, `CollisionXYData`__
Classes that maintain the information regarding collision values.  Data is formatted in pairs - a value used to indicate an intensity, and a value used to indicate the amount of damage to inflict.  Originally, these are multiple separated fields with the expected name mapping.  For "Z", the intensity is height; and, so, a measurement of vertical translation of the entity is taken live.  For "XY", the intensity is percentage of maximum vehicle throttle; and, so, the maximum forward velocity of the vehicle on flat terrain is noted in the `VehicleDefinition` of that kind of vehicle.

__`Collision[With]Reason`, `CollisionDamage[With]Modifiers`, `CollisionDamageModifierFunctions`__
Standard "damage to vital statistics" classes to support integration into the existing damage resolution workflow.  The `DamageProperties` for collision doesn't discuss damage fields because that's based on some other calculation; but, it does enforce a nonvanilla calculation - collision being considered a form of "direct" damage, that it gets resisted very slightly by the exo-suit resistances utilized for projectile damage.

__`MountedEntity`, `CargoBehavior`*, `CarrierBehavior`__
See Addenda 3.  These classes are the deconstructed facets of the original `CargoBehavior` assigned to provide targeted functionality for the specific object types.

___Caveats___
1) The more I try to make sense of collision damage math based on what little is possible to trace in the packets compared to what I read in the ADB, the less sense it makes.  The fields in the ADB really don't lend themselves to making sense of the math implied.  This is all a big guess to introduce ther basics of the mechanic.
2) Near-maximum throttle roadkill damage against infantry would have been rather overpowered were it not for an artificial modifier.  The ADB contains a field that is the maximum damage possible against Infantry, defined per vehicle.  The estimation I make finds the percentage of the current velocity against maximum forward throttle for that kind of vehicle.  Many vehicles do more damage than players have health at full throttle, however, and that's far too much damage done based on memory of how vanilla worked.  This artifical multiplier that the damage has been applied.
3) The tactical response area protection (TRAP) deployable has a damage multiplier against targets colliding with it defined in the ADB that would easily reach absurd levels, even if only the smaller vehicles would be destroyed instantly.  The multiplier has been set lower here just to make it seem fair.
4) Small deployables like sensors and the small turrets (and maybe the telepads) should always be destroyed when collided by a vehicle _at any speed_.  This is client-enforced behavior and I must respect it because I do not want to break immersion.  Currently, the same math that performs normal damage calculations also enforces the destruction of these deployables and I am not fully confident it won't leave the ghost of a deployable in the game world - other players will continue to see the deployable but it will vanish for the player. Usually, if the vehicle that crashed through it immediately backs up, it will crash into an "invisible deployable" again and that properly destroys it.
5) It's probably possible to bail out of a vehicle from a height that does not cause fall damage on landing.  Despite that, the flag will still be set and remain set after landing.  The conditions for cancelling a bail status requires either increasing in altitude or utilizing the status due to a collision event.  Under the stated circumsyances, it is probably possible to conserve bail protection and then hold onto it to survive a greater drop later.  Off the top of my head, I can not conceive of any situations under which it would be useful (or abused) right now.
6) I attempted to implement a knockback effect from collisions at first but gave up on it for now.  Vanilla PlanetSide did have something of knockback interactivity during collisions but, without knowing how to make vehicles move from the perspective of the driver without completely reconstructing them and breaking the flow of the game, let alone how to safely push targets without having to place them somewhere explicitly, risking placing those targets in out of bounds areas, the matter of pushing targets based on collision was tabled as "unsafe".

___Addenda___
1) I originally attempted to solve my issues using "proper momentum calculations" and that went terrible.  To cut to the quick, there are ADB fields that suggest the use of "force" and "mass" as modifiers somehow.  All vehicles have a mass field and, as mentioned, a maximum velocity.  These clues started me on a path to using this "velocity" to get to "momentum" and from that "force" and somehow apply that to damage values.  The results were either terribly underappreciated or consistently overpowering and I decided to table my attempts to stabilize the math after a certain amount of effort wasted.  Remember: I have no clue how any of this is supposed to work, really.
2) Fixed an issue with `DamageWithPositionMessage` packet attempting to clamp the wrong value between 0 and 255 causing the packet to not encode correctly.
3) Attempted to fix an issue resolving blockmap positioning when the target should not have a blockmap entry to reference due to changing map dimensions.  The problem demonstration was: deconstructing from a warp gate on right side any the standard continent (removed from blockmap), receiving an update packet like `PlayerStateMessageUpstream` (re-added to the blockmap), and reconstructing at one's sanctuary (failure to resolve the blockmap for an index that is out of bounds).
4) Fixed cargo carrier vehicles such as the `lodestar` and the Galaxy `dropship`.  Before the vehicle control agency separation update, all vehicles contained the ability to be ferried as cargo vehicles and ferry cargo vehicles if that vehicle also had cargo holds.  After the split, these responsibilities were erroneously assigned only to vehicles that had cargo holds and that made it impossible for anything but carrier vehicles to be cargo vehicles, equally impossible.  Further dissemination of logic was necessary.
5) Fixed an issue where the construction items did not resolve their initial construction mode correctly.  The problem demonstration was: using an FDU (`advanced_ace`) while only possessing fortification engineering would never let the player attempt to build anything but a one manned field turret (the first fire mode) that the player would need assault engineering to build.  The only appropriate deployable in this case should be the tactical resonance area protection (the second fire mode).
6) When I was doing the math to try and guess when a player is falling or jumping, I had first sampled from the velocity field to determine whether or not the player was accelerating on that axis.  This method proved a useless determination as the velocity field actually fluctuates in such a way that, for some updates, a relatively increased velocity will give way to a slower velocity then go back to a increased velocity.  Even when free falling at what the game considers terminal velocity, the field will report variation in between each individual update.  I speculate this is due to slight imperfection in timing of the packet updates and the client casually sorting that out and, without some way to properly predict the variance, I settled on a different method.  In any case, there's not enough reporting from the client to determine velocity changes or acceleration with any sort of granularity at the server level.
7) AT THE SAME TIME, I discovered that the velocity field underreports itself by a factor of average 3.6 times.  I don't know why the factor is a sub-4 number and it's _not even an average of 4_ since the game otherwise does desparately try to maintain the illusion of a four packet per second pace.
8) The average land speed of a wild Auraxian scrub is 27 KPH.

P.S. - I am certain someone is going to be excited about the possibility of launching vehicles, and that someone is someone who is not me.  The thing is: you can already launch vehicles in the client.  It all depends on the difference in mass and the inelasticity of the interaction.  For example, position a Deliverer and drive into it from the side with a vehicle of somewhat less mass, a Harasser, for example.  Eventually, some kind of launch should respond.  It may or may not be controllable.  It may or may not be something that can be adjusted or replicated in all the situations that people would wish it work.  Knowing that the client is enough to generate the physics disaster that allows otherwise ground vehicles to take to flight, and with this developer's poor reception to momentum calculations, I wash my hands of the matter.


Fate-JH:
> [error] Error while emitting net/psforever/objects/GlobalDefinitions$
> [error] Method too large: net/psforever/objects/GlobalDefinitions$.init_vehicles ()V
Lines 5560 to 7189.
It's that time again.

ScrawnyRonnie:
After some testing:
- Driver is not credited with the kill. It shows the vehicle as the killer.
- Damage received seems high. Rexo was killed by AMS and Reaver both at full speed. While realistic, full Rexo wouldn't die to these vehicles on Gemini.
- Reaver (and other light aircrafts I would guess) basically blows up on impact at high speed. It would do a bit of damage to you before, but I don't remember dying if I bumped something at high speeds from full armor.
- Bailing as a pilot from an aircraft took damage (4 health remaining) if the bail animation happens. If bailing so the animation does not play (by angling the aircraft rather than being level with the ground) no fall damage was taken.

From observation, impact damage at lower speeds "feels" right on. The collision damage increase based on the increase in speed (whether it be the moving vehicle or your character's fall speed) may just need to be lowered.

Dethdeath:
- Infantry fall damage seems too high, jumping off of a tower kills you. Surging down the stairs in a stairwell can take away 50 health just from one flight of stairs.
- Collision damage between vehicles seems off, a good example of this is when you bump into a parked mosquito with a lightning, after like 8 collisions between these 2 vehicles, the lightning is dead and the mosquito is just below half health.
- Getting into the cargo hold of a galaxy took away 25% of my ANTs health.
- An ANT that fell out of a galaxy because it was deconstructed in the air nearly lost all of its health. While this is intended behavior, the damage seems excessive. I feel like it was maybe half of this in the past?

Fate-JH:
I updated the damage value calculations, especially for squishies, and the kill tracking.  Give it a try.

ScrawnyRonnie:
image
- NC is the AMS driver. The AMS was at full armor before the collision. I drove down a hill to get to a higher than maximum speed (while driving on level ground) and the collision caused the AMS to blow up. Interestingly, both the driver and softie were credited with kills and rewarded xp.
- Low speed, ~40 kph Reaver still killed Rexo from full.
- Dropping off a tower took just over 50% health. Looking at footage from Gemimi, dropping off a tower didn't take any health. It only took 10 armor from a Rexo.


Fixed by latest changes:

- Bailing from an aircraft (both with and without the bailing animation) no longer taking damage.
- Driver is credited with kill. If possible, in the chat, the vehicle's icon instead of the skull would be ideal.
- Reaver was not taking as much damage when bumping into a tree or building at high speed. It was taking ~ 1/5 to 1/4 which seem about right.

Fate-JH:
A few things regarding your first test with the AMS:
1. It doesn't matter how fast the vehicle is moving over its maximum velocity because damage is, generally, counted in tiers of speeds where 1+ is always the same maximum damage.  Additionally, against Infantry, the damage is always capped to the maximum damage against Infantry as is recorded in the ADB for that vehicle type.
2. I commented about this in Discord already, but the client does not give me sufficient information regarding whether the damage target has had a long fall or if the damage target crashed into a wall.  Frequently, it will sum two separate calculations, and one of those calculations might be incorrect for the given situation. The game is always keeping track of how far the controlled entity has fallen or risen and applies that value during one of the collision calculations.  Going downhill and crashing into something at the base would count as a "fall".
4. It appears the order of events follows: the AMS, going downhill, went downhill so much that it took damage, applying both head-on collision and fall collision.  When it struck the player at the base of the hill, that subsequent collision was enough to kill the player.  When it struck the player, the feedback from the collision was enough to destroy the vehicle, and the player was awarded the kill by virtue of being involved in the impact.  If the player had survived the collision, the player would have died in the vehicle's explosion anyway.

Fate-JH:
Based on previous feedback, I did a thing again.

Fate-JH:
I did yet another thing.  I can't replicate @Dethdeath 's Lightning descent properly but I did remove the primary number that was causing the calculation of fall damage in some instances fo collision.  Additionally, I tried to resolve @KingFeraligatr 's complaint about activation of mines.

PR #958, opened 2021-10-25
Stamina Kit Fixed, Maybe

Fate-JH:
I think I did this originally because I didn't want to settle on sending the confirmation message to clean up the stamina kit independently of sending the message to change the amount of stamina, then I forgot about it when tying things up for the pull request.


Fate-JH:
Okay, try it now.

PR #962, opened 2021-10-31
Relog Fix

Fate-JH:
The entities being stale is often a problem.
This time, stale entities were the solution.

___Addenda___
I think player relogging causes implant timers to reset.  I mean, I think it always did, and that's not the problem I set out to fix, but it's worth paying attention to this disclaimer before you complain.


Fate-JH:
The things you can do on a phone these days lol
Happy halloween!

PR #982, opened 2021-12-25
Lump of Coal

Fate-JH:
Here come BFR's, here come BFR's, from within the enemy's ranks
Salty tryhards just mad `cause they couldn't ROFLstomp with tanks
Klaxons are flaring, turrets are broken, more important issues to right
So abandon your armor and just redeploy, if the BFR's come to the fight

Here come BFR's, here come BFR's, from across the lattice lane
You've no skill if you're countered by the pilot of a scouting plane
The shuffling creaking and clanking of an intimidating piece of scrap
So unsling your rifles when you bail out, if the BFR's try to attack

For years, the imposing pentagonal structure in the courtyard of major facilities had lain dormant and useless, save for cover.  For a time in the game's lifespan, these sheds were not even part of the facility courtyard.  Many a map can be found online where its existence is neglected and there is only empty space.  For most of the game's life however, the shed did something.  When the game was moved to free-to-play, the caverns and all of the additional features granted by the caverns were blocked, and it stopped working again.  Finally restored, a feature that has not been witnessed in years finally has been implemented.  Battleframe robotics.

All factions have access to two variants of the battleframe that are similar.  The bipedal frame of the body is the same for each, though between the different units the VS battleframes are humanoid, the NC battleframes are chicken-walker-style, and the TR battleframes are somewhere between the two.  By virtue of having legs, all battleframes have the ability to crouch or kneel.  Battleframes all have a shield generator for a self-charging shield which is a powerful defensive asset for the larger than normal ground vehicle and also has a visual representation that periodically swirls around the unit when active.  The pilot has simultaneous access to two removable and exchangeable equipment slots on the "arms" of the body-like vehicle, a departure from most vehicles where the pilot may or may not have a weapon and all weapons that are mounted to a given seat are fixed.   Additionally, the weapons attached to these arm weapon mounts, if attached, can be deactivated as well, one shooting independent of the other.  A trunk is accessible from the vehicle's flank.  The primary variant is usually called a "gunner" variant because it possesses a similarly removable and exchangeable bulky third weapon mounted above the "head" cockpit and a second seat associated with that weapon is accessible from the rear of the vehicle when it is kneeling.  The other variant is called a "flight" variant because it possesses a jet attachment (pod) instead of a third weapon and second seat.  It operates on a recharging capacitor.  Rather than literally flying, though, it's more like the VS MAX jumpjets with much better horizontal maneuverability in exchange for shield activity.  The flight variant also possesses much better ground speed than the gunner variant.

The VS battleframe is called the Aphelion and its flight variant is called the Eclipse.  The TR battleframe is called the Colossus and its flight variant is called the Invader.  The NC battleframe is called the Peregrine (it's a type of falcon) and its flight variant is called the Eagle.

Lodestar vehicles and facility rearm repair silos have special terminals for interacting with battleframe units (though, graphically, all terminals share the same interface design with restricted tab access).  The terminal attached to the aforementioned spawn shed allows for spawning of the standard battleframe vehicles in addition to up to five previously-saved battleframe loadouts.  The rearm repair terminal allows for weapon customization of any of the two or three weapon mounts and loading of ammunition into the trunk, as well as access to the battleframe loadouts affecting the existing Vehicle as if it were Infantry.  The Lodestar terminal is similar to the rearm repair terminal, though obviously attached to a mobile supply unit rather than as a fixed environmental element.  One major departure from typical inventory-space operations is that most things can not be dropped from a battleframe trunk or when acquired from a battleframe terminal.  Most equipment that would be dropped normally (when held instead of stowed where the interface is disposed) is destroyed instead.  To properly move ammunition boxes, for example, one should manually open the trunk of the battleframe and move equipment in it into one's own backpack and then drop the ammunition box from the backpack onto the ground.  This does not work for battleframe weaponry as the trunk does not allow access to the weapon mounts.  The Lodestar also allows for the battleframe vehicle to mount it and be transported as cargo the same as it does for any other non-flight vehicle (see Caveat/Visual 4 and 5).

Gunner-variant battleframes possess weaponry that causes persistent environmental damage.  The Aphelion's plasma rocket pods and the Peregrine's particle cannon leave dangerous fields called radiation clouds whose damage is fixed to a location for a short duration and cause periodic damage for the duration that a player lingers in it.  The plasma mortar cloud is like being bathed in a plasma grenade.  The particle cannon eats away at the health of mounted targets, a departure from other forms of projectile-based damage.

Stepping back, in every cavern, there is a central structure called the Core Complex within which is a beam of light called the Core Beam that dominates the otherwise hollow interior of the building.  Jumping into the Core Beam generally results in death unless certain personal accomplishments within the caverns are met.  In that case, jumping into the Core Beam results in a special certification crucifixion that allows one to pull battleframe robotic vehicles from the aforementioned battleframe sheds, if that facility is connected to a valid on-continent cavern warp gate or facility connected to a cavern warp gate whose next connected facility is also faction affiliated.  This act of being imprinted is a very important permission-granting that allows access to the battleframe robotics.  Without it, the terminals on the sheds are inert.  As caverns are still mostly unimplemented in this revival of the game, all players have been granted free imprinting; and, with all facilities being granted cavern connection benefits due to the aforementioned issues with the caverns, all players from all affiliated facilities can deploy the battleframe units of that particular faction with the appropriate certifications unlocked.

___Features___
__`AvatarActor`__
Modifications to handle battleframe loadouts.  No database modifications or additional support elements are necessary for the new loadouts, save for the clarification of an `Enumeration` field that wasn't defined beforehand (`LoadoutType`).

__`ZoneProjectileActor`__
Dedicated zone support class to handle any remote (control) projectile whose property `ExistsOnRemoteClients` is true.  Previously, the code and messaging pathways that supported these entities only existed in `SessionActor` and utilized the event system as a cooperative method of determining projectile depiction.  Now includes two tiers of remote projectiles - the classic missiles and the new radiation clouds.  Radiation clouds are localized area of effect projectile damage proxies completely controlled by the `ZPA` - from creation to deconstruction - that exist for a set amount of time and are also added to the zone's blockmap structure for their duration in the game world.  Classic remote projectiles are eliminated by client interaction.  Remote projectiles are logged on the `Zone` entity and can be depicted at zone loading in this manner.

__`RadiationCloudConverter`__
Composes `ObjectCreateMessage` packet data for projectiles that are classified as "radiation clouds".

__`VehicleSubsystem`__
Vehicle features that can be influenced by game packets.  They're similar in concept to `Utility` elements but have no globally unique identifier number and may be attached to abstract vehicular concepts as opposed to physical or real vehicle elements.  Most subsystems just have on/off conditions, at least for now.  This feature isn't fully developed yet but some subsystems are being utilized to represent certain important battleframe conditions such as the flight pod, shield generator, and arm weapon functionality.

__`ArmorSiphonBehavior`__
Mixin that supports the battleframe armor siphon weapon and allows it to rope into the standard projectile damage system.  Support for both the `SiphonOwner` (the battleframe) and the siphon's `Target` (all vehicles) is included separately.  Also included is the damage modifier that allows the armor siphon's damage resolution path to apply repairs to the aforementioned `SiphonOwer` entity equal to the damage caused to the aforementioned `Target` entity.

__`BfrTransferBehavior`, `NtuSiphon`__
Mixin that supports the battleframe NTU siphon and serves the same purpose as `AntTransferBehavior` for advanced nanite transfer vehicles.  The `NtuSiphon` class is just a wrapper that supports the `NtuContainer` interface mapped to the fields of a `Tool` entity, allowing the arm weapons to be treated with standard nanite transfer interactions.

__`VehicleCapacitance`__
Mixin created by refactoring code from `ApcControl` for accouting of the vehicle capacitor.  While still useful for the APC-type vehicles, it features additional layers of timing controls for flight capacitor control when applied to the flight variant battleframe vehicles.

__`MountableInventory`__
Refactoring of the `VehicleData` seating transcoders so as to share with `BattleframeRoboticsData`.  Still contains extended functionality for representing all players who may be seated in the vehicle but the appropriate converter should still limit the scope of seated players to just the driver or pilot for the foreseeable future due to client dissatisfaction with representing more than one seat.

__`ServerObjectControl`__
A super class for control agency actor classes that implements the specific logic for changes to an entity via `PlanetsideAttributeMessage` packets, `GenericObjectActionMessage` packets, and `GenericActionMessage` packets.  For the moment, only battleframe units implement their arm weapon activity using this method (see below).  The primary benefit will be moving code out from `SessionActor` and into respective objects.

__`BfrControl`, `BfrFlightControl`__
Control agencies for battleframe vehicles.  Since only battleframe vehicles implement siphon weaponry, part of the algorithm necessary for making siphons work is directly implemented here rather than the aforementioned mixins.  Additionally, the unique delayed explosion of the battleframe units is supported, as is flight pod management for the flight variant battleframes.

__`AvatarAwardMessage`, `FrameVehicleStateMessage`, `GenericObjectActionAtPositionMessage`__
These packets were developed in the course of developing battleframe vehicles. `AAM` was analyzed during the process of probing the imprinting process but is actually useless for the imprinting process.  `FVSM` serves the same purpose as `VehicleStateMessage` for battleframe vehicles and is very important for this update.  It possesses slightly different fields that reflect the operational differences between battleframes and normal vehicles. `GOAPM` is necessary for triggering the NTU siphon's EMP feature (and not much else at the moment).

___Caveats___
__General__
1. I have implemented battleframe robotics.  This was the first mistake.
2. As mentioned, all players start as imprinted for free.  This will not be the case when the caverns are better maintained later.  For reference, a player needs a prerequitie number of enemy kills in the caverns as well as a prerequisite number of enemy facility captures in the caverns and the imprinting field, when set to "not imprinted", reports for both.
3. Battleframes vehicles would normally have various subsystems that fail upon taking certain thresholds of damage and, when damaged, will have various affects applied to them.  Some damage caused the trunk to lose equipment; some damage caused degraded performance from the weapons; some damage disabled the ability of flight variants to fly or the radar to detect hostiles.  Such nuances were introduced originally in an effort to make battleframe robotics more "balanced".  The operation of these damageable subsystems have not been implemented in the current build so do not concern that they can not be triggered.  They will also not be found in any quality of life pass related to this PR.
4. Equipment dropped from the battleframe, when it does happen, first attempts to properly delete the item from the game environment and then immediately attempts to drop the item (onto the ground).  It is not possible to react quickly enough to suppress the packets from the client so various mechanisms have been implemented to suppress fake warning messages.  Item deletion is always the outcome in our case.  BFR weaponry can not be dropped on the ground (and properly depicted), in any case.
5. Jammering a battleframe vehicle may not produce expected results.  This is partically due to the subsystem rails not being represented in whole.  A lot more analysis and work needs to go into this feature.

__Visual__
1. The smaller explosions along the battleframe vehicle's torso before the final destructive explosion are not always visible to all players due to unknown reasons.  Significant distance, facing direction, and torso visibility is probably a factor.
2. The dismount animation for battle frame vehicles has a tendency to break, resulting in the player driver emerging through an unopened driver cockpit.  All dismounts following that first animation failure will also result in animation failure until some form of animation refresh such as a zoning process.  It will probably break again after that anyway.  The conditions and reasons for the failure are uncertain and this may actually be a long-standing client issue that was never addressed.
3. Battleframe vehicle have a tendency to glide-walk across the ground when observed from a third person perspective.  This is a known visual bug introduced when the developers of days gone attempted to adjust the speed of the battleframes but neglected to properly update the animation speeds for the walk cycle.  To a player sat in one of the vehicle's seats, the walk cycle should look fine.
4. When a battleframe vehicle is mounted as cargo in the Lodestar, it does not anchor itself properly for third person perspective.  The battleframe will instead remain facing the direction in which it mounted the other vehicle rather than assume a side-facing mount position, will bob in its cargo bay a bit as if loose, and will spin to stay oriented in the original mounting direction despite how the Lodestar moves.
5. When a battleframe vehicle is mounted as cargo in the Lodestar, it should officially be mounted sideways into the cargo bay, much like how the Router is to be mounted sideways, facing the flanks of the shoebox carrier as opposed to the pilot's front.  Strafing foot jets on the vehicle would allow for this maneuver easily.  In certain views, the vehicle will be depicted as naturally mounted sideways as well.  Nothing stops the battleframe vehicle, however, from walking straight into the cargo bay of the vehicle and properly mounting, snapping the vehicle into its sideways oientation upon completion.  Since it fits just fine, the requirement seems weird. This property is shared with the Router, but the Router is oblong in such a way that it requires sideways mounting.
6. Normally, the battleframe vehicle has two arms that can independently be fired of the other and at least one of the arms must be enabled at all times.  In the case of NTU siphons or armor siphons, only one arm may be enabled at a time.  The state of the arm is indicated visually by the disabled equipment slot icon becoming faded.  Upon vehicle gating or any other sort of warp transport, the battleframe will forget its disabled arms and both will always display as active.  Despite my best attempts to have the vehicle in the correct state upon client readiness, the packets that affect the behavior and UI elements will always arrive too early in the loading process and thus the state of the arm weapon mounts that they are trying to set will not take.  To correct this issue, upon weapons fire, a check will confirm that the arms that have been detected are allowed to discharge and, if they are not allowed, then the packets that disable a weapon mount arm will be dispatched by the server and will properly update the state of the weapon and UI.  During this initial fire cycle any arm weapon that is incorrectly set will always fire a phantom projectile with no loss of animation.  The exact moment the client visually becomes ready can not be known so this is unavoidable.
7. The NTU siphon has a secondary electromagnetic pulse attack when charged but the exact effect triggered for the EMP is unknown.  The replacement is a similar EMP caused by the APC vehicles - the Juggernaut, the Vindicator, and the Leviathan.
8. Primary weapon interaction animations for the NTU siphon and the armor siphon are unknown.  Despositing NTU into a friendly resource silo will result in the resource silo's particle effect animation activating; but, no other valid interaction will be accompanied by fancy particle effects.  One must watch the subtle weapon animation of the siphon carefully.
9. Immediately upon being spawned, the view of the game world outside of the spawn shed may be culled for a brief moment.  Vanishment of the world is most certainly caused by a blindespot in the game's underlying portal system.  During some failed vehicle spawns in the process of development, looking at the interior of the BFR shed while its door is open causes the world to be visible, while looking out of the shed doors under the same circumstances will cause that game world to be invisible.  This is merely a very sutble bug and it corrects itself almost immediately sine the battleframe is made to walk out of the shed.
10. The equipment slot weapon UI for the siphon ammunition is weird.  It displays as '1 / 0" or "100 / 0".  Don't worry about it as long as the first field works.
11. The NTU siphon will sometimes/always start with one phantom ammunition.  As noted, this ammunition doesn't actually exist and is a product of the object creation packet.  Since it doesn't actually exist, it is ignored when the siphon starts charging and, since nothing happens at one charge, no harm should come from the discrepancy.

__Addenda__
1. Corrected an issue with the EMP from the the APC-type vehicles causing fatal damage rather than simple electromagnetic jammering.
2. Corrected an issue with remote controlled projectile global uique identifier registration pools, especially on Cyssor and Ishundar.
3. Corrected an issue with the message `Transferbehavior.Stopping` not being received correctly by certain NTU systems that are built upon typed actors.
4. Corrected an issue where the vehicle spawn times for some vehicles had an incorrect cooldown.  None of the vehicles where the cooldown was intentionally set differently for balance, i.e., the Galaxy Gunship, were affected.
5. Corrected an issue where an initialization value for facilities that is used to activate vehicle gate shields was not being appropriated to the correct facilities.  Note that the value is still defaulted; the change was to the targets.  (Another incorrect assumption about how imprinting works.)
6. Corrected an issue where stamina regeneration did not start upon periodic implant drain.
7. Yes, I typed "ROFLstomp" up there.  When's the last time you ever heard someone use that one?
8. Merry Christmas!


jgillich:
Awesome work! Just played with it, pulled Eagle and it worked fine. But pulling a Peregrine seems to kill the session:

```
2021-12-25 15:04:57,945  INFO n.p.actors.session.SessionActor - ILoveBFRs will load in zone z10 at position Vector3(3397.0,5799.0,48.0) in 0 seconds
2021-12-25 15:04:57,947  INFO n.p.actors.session.SessionActor - ILoveBFRs has left zone home1
2021-12-25 15:04:57,947  INFO n.p.actors.session.SessionActor - ILoveBFRs has spawned into z10
2021-12-25 15:05:19,443  INFO n.p.actors.session.SessionActor - ILoveBFRs is accessing a bfr_terminal
2021-12-25 15:05:25,855  INFO n.p.actors.session.SessionActor - ILoveBFRs mounts the peregrine_gunner in the driver seat
2021-12-25 15:05:30,770 ERROR akka.actor.SupervisorStrategy - None.get
java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:627)
	at scala.None$.get(Option.scala:626)
	at net.psforever.actors.session.SessionActor.$anonfun$FindEnabledWeaponsToHandleWeaponFireAccountability$1(SessionActor.scala:9303)
	at net.psforever.actors.session.SessionActor.$anonfun$FindEnabledWeaponsToHandleWeaponFireAccountability$1$adapted(SessionActor.scala:9293)
	at scala.collection.immutable.Set$Set2.filterImpl(Set.scala:212)
	at scala.collection.immutable.Set$Set2.filterImpl(Set.scala:183)
	at scala.collection.StrictOptimizedIterableOps.filter(StrictOptimizedIterableOps.scala:217)
	at scala.collection.StrictOptimizedIterableOps.filter$(StrictOptimizedIterableOps.scala:217)
	at scala.collection.immutable.Set$Set2.filter(Set.scala:183)
	at net.psforever.actors.session.SessionActor.FindEnabledWeaponsToHandleWeaponFireAccountability(SessionActor.scala:9293)
	at net.psforever.actors.session.SessionActor.HandleWeaponFireAccountability(SessionActor.scala:9244)
	at net.psforever.actors.session.SessionActor.HandleWeaponFire(SessionActor.scala:9153)
	at net.psforever.actors.session.SessionActor.handleGamePkt(SessionActor.scala:5369)
	at net.psforever.actors.session.SessionActor$$anonfun$receive$1.applyOrElse(SessionActor.scala:442)
	at akka.actor.Actor.aroundReceive(Actor.scala:537)
	at akka.actor.Actor.aroundReceive$(Actor.scala:535)
	at net.psforever.actors.session.SessionActor.akka$actor$MDCContextAware$$super$aroundReceive(SessionActor.scala:164)
	at akka.actor.MDCContextAware.aroundReceive(MDCContextAware.scala:25)
	at akka.actor.MDCContextAware.aroundReceive$(MDCContextAware.scala:14)
	at net.psforever.actors.session.SessionActor.aroundReceive(SessionActor.scala:164)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:580)
	at akka.actor.ActorCell.invoke(ActorCell.scala:548)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:270)
	at akka.dispatch.Mailbox.run(Mailbox.scala:231)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:243)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

jgillich:
After zoning with a BFR you can't move until you exit and re-enter (applies to both WGs and /zone)

Fate-JH:
> After zoning with a BFR you can't move until you exit and re-enter (applies to both WGs and /zone)

You need to uncrouch.  The only control for crouching as far as I discovered is the `FVSM` packet and that doesn't go back into the client that should be pushing them onto the server.  The default keybinding for crouch control for the BFR's is `C`.

Mazo:
Bug: Starfire secondary weapon can't be reloaded in some cases

Repo steps:
1. Pull an eclipse
2. Have a starfire on both arms
3. Use the "clear all" button on the rearm silo
4. Grab two more starfires and some ammo, fire then attempt to reload. Only the left arm will reload correctly.

Interestingly, this doesn't seem to affect the PPA or laser. The TR BFR weapons seem unaffected. (NC not tested)

Mazo:
Bug: Shields remain active when exiting a BFR

Repro:
1. Pull bfr
2. Exit BFR
3. Observe shields are still active. Pull another vehicle, attempt to damage the BFR and the shield bar does not change, and does not take damage once shields are depleted

Mazo:
Bug: Weapons can be added to the trunk causing weird inventory behaviour, plus client crashes

Repro:
1. Pull BFR
2. Clear all at loadout terminal
3. Keep double clicking on the armor siphon to buy one for each arm, after that they will start to overflow into the trunk

Mazo:
Bug: You can bail out of flight variants

Repro:
1. Pull flight variant
2. Hold space to fly
3. ALT+G to bail

I don't remember ever being able to do this in live, and you take fall damage when doing it which would suggest it's not expected behaviour.

Mazo:
Bug: You can pull a gunner variant while on cooldown

Repro:
1. Pull a BFR
2. Jump out, go to pull another BFR. The flight variant correctly blocks purchase, but another gunner variant can be purchased.

Mazo:
Bug: NC Anti Tank Machinegun has too much ammo (150 vs 55)

Repro:
1. Pull any NC bfr
2. Ammo counter will read 150. Weapon cannot be reloaded until ammo drops below the usual level of 55 according to the wiki.

http://wiki.psforever.net/wiki/Anti-Tank_Machine_Gun

Mazo:
Bug: Radiation clouds don't seem to work at all, this seems to be affecting all weapons that should create radaiation, such as the VS rocket pods, the NC particle cannon and possibly even the radiator handheld weapon (assuming this worked beforehand).

The starfire also doesn't seem to have aggravated damage when shooting at a wall turret, but this may be expected (does it only do damage over time to aircraft)

Mazo:
Bug: Gunner weapons can be dropped on the floor

Repro:
1. Pull BFR
2. Place new gunner weapon on cursor at rearm terminal
3. Click on an ammo box in the trunk while the weapon is on the cursor
4. The weapon then gets dropped on the floor automatically

Fate-JH:
> Bug: Shields remain active when exiting a BFR

Easily handled.
> I don't remember ever being able to [bail out of flight variants] in live, and you take fall damage when doing it which would suggest it's not expected behaviour.

The fall damage is because the `GlobalDefinitions` entry doesn't mark the seat as bailable thus the occupant never gains bail protection upon dismounting; but, if the client allows you to bail, then bailing must be allowed.  The server does/can not force working bail status on the client.
> Bug: You can pull a gunner variant while on cooldown

This is a weird UI thing where both timers update for one of the BFR types in a packet but, internally, only that specific variant has a server timer blocking it.  Since we already can timer block all MAX variants when a single kind is selected, a solution should be feasible.
> Bug: NC Anti Tank Machinegun has too much ammo (150 vs 55)

Copy-paste error, or maybe I was looking at the wrong value in the ADB?
> The starfire also doesn't seem to have aggravated damage when shooting at a wall turret

Aggravation from a starfire projectile only affects aircraft.

Fate-JH:
The aforementioned issues should be corrected.

Fate-JH:
Updated:
Battleframe subsystems are enhanced thus that jammering the vehicle can impair some functionality (movement speeds) and will completely disable others (shields and flight systems) for the duration of the effect.

Fate-JH:
Updated comments:
___Features___
__`VehicleSubsystem`__
Vehicle features that can be influenced by game packets. They're similar in concept to `Utility` elements but have no globally unique identifier number and may be attached to abstract vehicular concepts as opposed to physical or real vehicle elements. Some subsystems represent certain important battleframe conditions such as the flight pod, shield generator, and weapon functionality.  One specific subsystem is dedicated to toggling use of the battleframe arm weapons.  A number of subsystems have been provided are used to represent behaviors for the battleframe vehicle's jammed state.  The UI for the subsystems caddies the UI for vehicle health, and includes icons that indicate the state of the subsystem.  The icons will change color (green, yellow, red) and the specific problems that plague the system will be listed in full text next to it.

__`AvatarAwardMessage`__, __`FrameVehicleStateMessage`__, __`GenericObjectActionAtPositionMessage`__, __`ComponentDamageMessage`__
These packets were developed in the course of developing battleframe vehicles. `AAM` was analyzed during the process of probing the imprinting process but is actually useless for the imprinting process. `FVSM` serves the same purpose as `VehicleStateMessage` for battleframe vehicles and is very important for this update. It possesses slightly different fields that reflect the operational differences between battleframes and normal vehicles. `GOAPM` is necessary for triggering the NTU siphon's EMP feature (and not much else at the moment).  `CDM` is used for subsystem control, jamming and weapon toggle.

___Caveats___
__General__
5. Jammering a battleframe vehicle may not produce expected results.  Certain subsystems will report failure when the vehicle is jammed but the wiki was not specific about which ones as it scrambled the categories of damageable statuses of a subsystem and its jammable statuses within one portion of the main battleframe page. Most status failures can not currently be quanitified at the server level in any case such as speed and radar modifications.  The following things currently occur:
- for all weapons:
-- the cone of fire has increased recovery time
-- maximum projectile range decrease
- for the vehicle:
-- transit (travel mode) is disabled
-- movement in all directions recieve a speed reduction
-- proximity radar will fail to show any enemy targets
-- the flight pod will stop working
-- the shield generator will recharge slowly
-- morever, the shield generator will become ineffective
- for the siphon (and other statuses):
-- no changes to the siphons, because I do not have statuses for them on file
- general:
-- the vehicle does not play the alarm klaxons when subsystems suffer jamming failure; the kalxons are being reserved for actual **damage** failure

__Visual__
12.  A previously active energy shield should turn to periodic static upon being jammed.  If the shield is deactivated such as when no seat is occupied and then the vehicle is jammed, the shield may or may not activate upon the vehicle being mounted and a now-activated shield may or may not still have the static overlay that indicates being ineffective.  It will display properly when the jammering condition ends provided some seat is occupied.  Additionally, that status for the shield generator may go missing from the subsystem UI if the driver mounts after being jammed.

Fate-JH:
I WARNED YOU

Issue #977, opened 2021-11-22
Combat log shows "He Mine" instead of High Explosive Mine


Dethdeath:
![HeMine](https://user-images.githubusercontent.com/77748540/142879821-4ff3c7b8-5aff-41f2-b586-636ed3923fd0.png)

PR #983, opened 2022-01-23
fixed typo

Imstupidpleasehelp:
Changed "he_mine" to "high_explosive_mine" and also formatted the page I was on


jgillich:
Thanks for your contribution! Is this meant to address #977? I'm not sure if this does it. Most names are derived from the IDs in the game files and we try not to deviate from them.

Imstupidpleasehelp:
Yes it was meant to address #977, it appears my attempt was unsuccessful

PR #985, opened 2022-02-02
Battleframe Branch Bugfixes

Fate-JH:
Various corrections to functionality related to feedback through Discords's #bug-report channel.

Read the commit messages for specifics.


Fate-JH:
Clarification 1: You can drain a facility with siphons to the point where the UI reports 40% and then keep draining until it can not drain any more without dropping to 30%.  You can not drain to any value that would drop the value below 40%.

PR #987, opened 2022-03-24
Last Infantry Weapons

Fate-JH:
The Scorpion is programmable rocket launcher capable of and entirely dependent upon indirect explosion damage.  In its "primary fire mode", the weapon, when scoped, acquires the range of a target out between fifty-five meters and about two-hundred meters.  The weapon can be switched over to a proper discharge mode where it will shoot a projectile out to that distance in a straight line.  Once it has reached the expected distance, the primary projectile explodes and releases five small bomblets called "little buddies".  The little buddies will descend twenty-five meters in a predictable downwards pattern and also explode; this explosion is the primary method of the weapon's damage dealing, though the initial projectile's explosion will also inflict some damage.  Two projectiles fly straight down on the target's head while the remaining three scatter around the vicinity.  If any of the projectiles encounter an obstacle before traveling far enough to explode, they become inert and no damage is caused.  The Scorpion is not designed to be as hard to use as possible but it certainly is the hardest weapon to use out in the field as a soldier utilizing it must be able to process unreasonably accurate distance calculations.  Only for those of elite skill (and certification), it would only ever be viable in the hands of blatant hackers and trolls who would rather use other cheese anyway.  I seriously only damaged targets with the bomblets twice during two-plus weeks of controlled testing.

The Radiator is weapon of Ancient origin that fires energy projectiles, either ones that arc and bounce off surfaces, or ones that fly straight and stick to surfaces.  Either way, the projectile itself does no damage on surface contact.  Shortly, however, the projectile will produce a "radiation cloud".  All Infantry in range of the cloud will begin suffering continuous damage that one is only safe from by avoiding the spherical cloud entirely.  Operating a mechanized assault exo-suit or occupying a vehicle offer some protection from the radiation but the damage will not be completely eliminated.  The radiation cloud will only exist for a few seconds but one should never dwaddle in its range carelessly for any length of time.  Not only does one need a special certification for use of the Radiator, the facility in which one can acquire a Radiator must provide the benefits of a cavern weapon module.

When at a facility that provides the said benefits of a cavern weapon module, Ancient weaponry (for Infantry and Vehicles) will be recharged naturally without needing to resupply ammunition cells from an equipment terminal.  For Vehicles, the vehicle must be navigated within vicinity of the rearm/repair silo in the facility courtyard.  Mounted weapons are valid targets for recharge.  Infantry only need to be anywhere within the sphere of influence of the benefited friendly facility for the recharge to occur.  Only Ancient weapons in their equipment slots and in their backpacks are considered for charging.

___Features___
__`LittleBuddyProjectileData` and __`LittleBuddyProjectileConverter`____
The little buddy is a remote projectile - must be created on each client that would view it - that is constructed using different parameters than different remote projectiles.  As such, it also converts into its object creation packet differently.  The transcoder is unique in two ways: its length reports a trailing bit that will never be flagged; and, the transcoder must be accessible by both the simplistic object create message packet and the advanced object create detailed message packet.  The converter never bothers to figure out what any of the fields for the transcoder do and it just sorta works anyway.

__`WeaponRechargeTerminalDefinition`__
Similar to `MedicalTerminalDefinition`, but for ammunition.  So, not really like a `MedicalTerminalDefinition`, more like the rival of a `MedicalTerminalDefinition`.

__`DummyExplodingEntity`__
A class intentionally included to get around a design flaw in the prototype for `Zone.serverSideDamage`.  Essentially, where there is no functional entity to inhabit the properties of the explosion, an entity of this class is created, dolled-up, utilized, then discarded.  Consider its existence short-lived.

__`SessionActor`__
A major portion of the support for the little buddy projectiles for the Scorpion and the proximity terminals for mobile weapon rearming (as well as everything else proximity terminals do) is contain in this class for lack of a better place to occupy or lack of a better workflow.  It just sorta works anyway.

___Caveats___
1. The Scorpion's little buddy second-stage projectiles do not know where the ground is and do not report back on their flight properties. The damage caused by them is actually rendered as server side explosive damage after a fixed time duration.  Neither an actual detonation nor a failed detonation due to interacting with other surfaces are comunicated.  The explosion will always occur one second later at whatever is 25m distance from the origin of the those little buddy projectiles.  The consequence of this is that projectiles can technically cause explosive damage through the roof of a building or inside an interior room so long as the second-stage projectiles spawn close enough to the floor above the ceiling of the said chamber. or one of its walls.  The only consideration that can be written for this is to stop the projectiles from causing explosive damage against targets if the point of explosion occurs beneath their foot height.
2. The Scorpion's little buddy second-stage projectiles do not have damage fall-off.  This intentional design decision hopes to make the weapon even slightly more practical.
3. According to the wiki description, the five little buddy bomblets from the Scorpion are deployed in a semi-predictable manner.  Two of them fly straight downwards towards the original target point.  The remaining three bomblets fan out in a somewhat indeterminate directions at indeterminate angles.  The goal is to maximize the chances of target encounter.  In our case, the two downwards projectiles continue to travel downwards towards the original target point, but at very subtle angles away from it,  The remaining three bomblets spread out evenly based on the standing direction of the user at a predictable angle.  Due to bomblet spawn issues inherent to the client and the unrealistic calculations required to use the Scorpion with any sort of finesse, it probably does not matter which method is utilized.
4. Radiator damage is more progressive and extreme than the normal interval for analysis of the zone-level interactionsd allows.  As such, part of the damage is simulated through aggravated effects.  The practical tradeoff is the damage presentation looks correct but its threshold per dose is quantitized per 250ms period.  Exposure to the radiation field for 150ms will incur the same amount of damage as if one was exposed for 250ms; and, 300ms exposure the same as if 500ms exposure; and, so forth.

__Addenda__
1. The Scorpion's internal name is OICW, standing for "objective individual combat weapon".  It refers to a series of progressive developmental service weapons produced via competition between arms producers for the US military.  In reality, very little left the prototyping phase.  The specific weapon on which the Scorpion is based is the XM25 counter defilade target engagement (CDTE) system, an airburst grenade launcher with laser-programmable ammunition.
2. Resolved an issue where radiation clouds were having their damage value negated in two different ways - having the wrong modifier value and having the wrong geometry representation.
4. Resolved an issue where the repair terminal functionality for landing pads was still expecting increments in terms of nanoseconds while the field had been converted to express milliseconds.


Fate-JH:
The numbers on the Scorpion's sub-projectiles are correct,  The weapon is just really really hard to deal damage with and I can not think of anything that can be done to make it easier to work with right now.

PR #1010, opened 2022-10-23
Shortcut to Grenade

Fate-JH:
This first major feature development fueled by development of a function that is not vanilla to the game.

If the player has a grenade - either a fragmentation grenade or a plasma grenade - somewhere on their person and they wish to switch to it quickly, a command is now available that summons the grenade into his hand and brandishes it for throwing.  The player has to throw it manually since the act involves variable strength.  Any weapon that he is holding previously is replaced into the same holster while the hand is down.  The grenade is placed into the **first sidearm holster** and whatever was originally in that holster is either placed into the player's backpack or dropped onto the ground if there isn't enough room.  If the grenade is in the other holster, the hands are merely swapped without modifying the position of any equipment; if already in a holster but no weapons are drawn, that grenade is merely drawn.  Of course, any player donning a mechanized assault exo-suit may not do this.  A maneuvering of equipment allows for the grenade to be drawn very slightly faster than usual if it comes out of the backpack.

Along the base of the player HUD are eight slots that are intended to be a hotbar.  When things are assigned to these slots, they are called "shortcuts".  An additional seven sets of hotbar slots can be accessed using vertical arrows alongside the hotbar area, resulting in sixty-four shortcuts, with eight available at a given moment.  Quick-use access to these hotbar shortcuts is normally assigned the function keys (F1 to F8) but keybindings may modify this.  The game defaults the first hotbar element as a quick access to any medkits in one's backpack, while each implant the player installs will also add an hotbar icon.  By making use of the text chat command `/macro`, the player may create their own text change message that will be automatically spoken within an assigned domain (local, squad, platoon).  This custom message appears in the shape of a word bubble with three letters that will hopefully indicate the nature of the underlying chat message.  Clicking on an open hotbar slot places the message there while clicking on an occupied hotbar slot will cause the shortcuts to swap between cursor and slot and will require the swapped shortcut be placed somewhere else.  Left clicking on an occupied hotbar slot when in cursor mode removes the shortcut but allows it to be moved to another slot; right clicking on a shortcut deletes it entirely and it must be remade.  Exit cursor mode to forget about any yet-to-be-placed shortcuts that are attached to the cursor.  These shortcuts are persisted in the database but must be reloaded to the client every character redraw event (zoning, respawn, reconstruction, etc..).

___Commands___
These chat commands are naturally in the game.
`/macro [acronym] [message]`
The normal method of adding macros to the hotbar.  The process has been described previously.

These chat commands have been added.
`!macro medkit`
Run a sanity test to make certain the medkit shortcut icon is available on the hotbar.

`!macro implants`
Run a sanity test to make certain all of the avatar's installed implants have a shortcut icon available on the hotbar.

`!macro [implant]`
Run a sanity test to make certain this implant has a shortcut icon available on the hotbar if it is installed on the avatar.

`!macro [acronym] [message]`
A backup method of adding macros to the hotbar.  Follows the same process that has been described previously.

`!grenade`
The quick-switch ...-to-grenade command.  Proof of concept not-vanilla feature.  Follows the same process that has been described previously.

___Features___
__`AvatarActor`__
Preexisting class.  Loads shortcuts at the time of avatar login; and, reads and writes them over the course of the avatar's lifetime. Sets an initial login condition that adjusts the time when things like certifications and initial shortcuts are established.

__`ObjectHeldMessage`__
Preexisting class.  Logic pathways for `ObjectHeldMessage` have been moved from `SessionActor` to `PlayerControl` and completely propagates through `AvatarService` to `SessionActor`'s rather than half of the result immediately returning to the sending client.  An additional variable check allows for the player's arm to be held is a certain position (arm up or arm down) until a condition is satified or the check is unset, which is useful for stopping the arm from lowering when wearing a mechanized assault exo-suit, for example.

__`CreateShortcutMessage`__
Preexisting class.  The process of transcoding the shortcut has been simplified to remove quote-unquote unnecessary fields.

__`net.psforever.objects.avatar.Shortcut`__
A storage medium for hotbar shortcuts previously read from the database or added by the client, maintained by the `Avatar` entity.  Restored once per client refresh (respawn, etc..).

__`net.psforever.persistence.Shortcut`__
A transfer medium between the server and the database for hotbar shortcuts.

__`ChatService`__
Since all new commands re completely normal packets that are distinguished only by their "commands" being preceded with a bang character (!), custom functionality is facilitated by `ChatMsg` packets

__`WorldSession`__
The quick grenade action is mainly handled by a function added to this support object.

___Caveats___
1. The player may not quick draw jammer grenades.  Jammer grenades are far too strategic to be quick-draw-able.
2. The player may only quick-draw actual grenades.  As a response to the command, he will not draw any weapon that is a grenade launcher (Thumper) or has a grenade launcher (Punisher) or that utilizes grenade-like projectiles (Rocket Rifle), whether or not the appropriate grenade type is loaded.  The anti-vehicular Terran Republic mechanized assault exo-suit is, of course, right out ...
3. The determination of whether a fragmentation grenade or a plasma grenade gets drawn is based on the order in which the grenade was inserted into the player's inventory.  It won't matter whether that first available grenade is in the top left-most position of the backpack space or the bottom right place, or if a jammer grenade was put into the backpack before any other type of grenade.
4. Implants always want to install their icons into the earliest slots of the hotbar - two through four mainly - but will also always install in the latest hotbar slot before trying to overwrite earlier ones.  A workaround that should stop this from happening has been implemented, though holes in the logic may exist.  Please report these bugs immediately as they may cause the database to break down.
5. The player may prune their shortcuts in anyw ay they want, including removed the default-by-feature shortcuts for medkit and implants.  While these shortcuts as missing, status of and application of these features is still possible through alternate means.  Medkits are visible in and can be used from the inventory window.  Restoring the medkit shortcut requires putting more medkits into one's backpack.  Implants can have their initialization and activation statuses checked and they can be activated from the character information window.  Restoring implant icons requires uninstalling and then reinstalling the implant.  Implants can have their initialization and activation statuses checked and they can be activated from the character information window.  Restoring the medkit shortcut requires putting more medkits into one's backpack.  To make this restoration process easier, however, custom text chat commands have been added.

___Addenda___
1. In case it isn't obvious, the features in this update are intended to be used together in this way: `/macro [acronym] !grenade`.  This is the only sensible way the quick grenade option could be made useful.
2. Discord user tijon better appreciate this.

### DIFF
--- a/server/src/main/resources/db/migration/V007__Shortcuts.sql
+++ b/server/src/main/resources/db/migration/V007__Shortcuts.sql
@@ -1,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS "shortcut" (
+  "avatar_id" INT NOT NULL REFERENCES avatar (id),
+  "slot" SMALLINT NOT NULL,
+  "purpose" SMALLINT NOT NULL,
+  "tile" VARCHAR(20) NOT NULL,
+  "effect1" VARCHAR(3),
+  "effect2" TEXT,
+  UNIQUE(avatar_id, slot)
+);
--- a/server/src/test/scala/actor/service/AvatarServiceTest.scala
+++ b/server/src/test/scala/actor/service/AvatarServiceTest.scala
@@ -237,8 +237,8 @@ class ObjectHeldTest extends ActorTest {
       ServiceManager.boot(system)
       val service = system.actorOf(Props(classOf[AvatarService], Zone.Nowhere), AvatarServiceTest.TestName)
       service ! Service.Join("test")
-      service ! AvatarServiceMessage("test", AvatarAction.ObjectHeld(PlanetSideGUID(10), 1))
-      expectMsg(AvatarServiceResponse("/test/Avatar", PlanetSideGUID(10), AvatarResponse.ObjectHeld(1)))
+      service ! AvatarServiceMessage("test", AvatarAction.ObjectHeld(PlanetSideGUID(10), 1, 2))
+      expectMsg(AvatarServiceResponse("/test/Avatar", PlanetSideGUID(10), AvatarResponse.ObjectHeld(1, 2)))
     }
   }
 }
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -13,7 +13,7 @@ import scala.concurrent.{ExecutionContextExecutor, Future, Promise}
 import scala.util.{Failure, Success}
 import scala.concurrent.duration._
 //
-import net.psforever.objects.avatar.{Friend => AvatarFriend, Ignored => AvatarIgnored, _}
+import net.psforever.objects.avatar.{Friend => AvatarFriend, Ignored => AvatarIgnored, Shortcut => AvatarShortcut, _}
 import net.psforever.objects.definition.converter.CharacterSelectConverter
 import net.psforever.objects.definition._
 import net.psforever.objects.inventory.Container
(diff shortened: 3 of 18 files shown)


tijon1:
amaze

Fate-JH:
Maybe a week with a lot of breaks to account for my job.  I did the actual code for grenades on the first day.

PR #1012, opened 2023-01-07
Relog Fix

Fate-JH:
Introduction: ever notice that all my PR's are formatted exactly the same?  Let me change it up this time.

Problem statement: relogging into the game world before the persistence timer has completely logged a character out but a character dies while it's user is disconnected causes the character to lose some attachment to the game world.  The primary demonstration is that manipulation of items in the inventory is not respected, but other symptoms can be observed easily.  Switching game zones (when you can) corrects the issue.

Problem explanation: a restored location login causes the old zone and the new zone to be the same location during login.  At one point in the original login protocol, the new zone gains an anchoring event subscription long before the normal subscription hooks are made.  All of the meaningful unsubscriptions in the old zone are removed entirely immediately after that.  When the two zones are the same, even this anchoring subscription is lost and is never remade; and, this anchoring subscription goes on to become one of the most important subscriptions for that character and session, at least until the next zone proper transfer occurs and everything is executed normally.

Problem scope: the very specific scope of the issue was logging back into the game and the game respawning you on the same continent in which you logged out last.  The end of persistence does not matter.  The dying in between logout and login was not a necessary step, it merely brought greater attention to the issue.  Additionally, this problem was originally reported after a client crash.  The client crash, too, was merely an event that highlighted the problem and did not contribute to it.

Solution: flip the order of the subscription addition and the subscription removals such that the anchor subscription is always set properly.  Cleanup where necessary.

Addenda:
1. For some reason, I rewrote the order of events in `AvatarActor` because I encountered a different bug when it came to the player avatar not being set when it should have been and was expected to be set.  While this may have just been another symptom, you can't help but admit the final result looks tidy.
2. I rewrote the `ActionResultMessage` transcoder because the inefficiency of the original was bugging me.
3. Special thanks to ScrawnyRonnie for reminding me that the easiest way to crash the client was not to actually crash the client.

### DIFF
--- a/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
+++ b/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
@@ -297,12 +297,20 @@ class MiddlewareActor(
                 send(ServerStart(nonce, serverNonce), None, None)
                 cryptoSetup()
 
-              /** Unknown30 is used to reuse an existing crypto session when switching from login to world
-                * When not handling it, it appears that the client will fall back to using ClientStart
-                * TODO implement this
-                */
               case (Unknown30(nonce), _) =>
+                /*
+                Unknown30 is used to reuse an existing crypto session when switching from login to world
+                When not handling it, it appears that the client will fall back to using ClientStart
+                Do we need to implement this?
+                */
                 connectionClose()
+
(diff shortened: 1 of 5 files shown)


Mazo:
👍

PR #1016, opened 2023-01-27
Code Style Improvements

Fate-JH:
Only cosmetic and performance changes.

See https://trello.com/c/jEJ4t9Qx/372-style-updates-1 for updates and details.


src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1334,107 +1089,115 @@ class SessionData(
...
-        } else {
-          log.warn(s"GenericActionMessage: ${player.Name} can't handle action code 20")
-        }
-      } else if (action == 21) {
-        if (player.ExoSuit == ExoSuitType.MAX) {
-          player.Faction match {
-            case PlanetSideEmpire.NC =>
-              ToggleMaxSpecialState(enable = false)
+      action match {
+        case 15 => //max deployment
```
leobenkel, line 1093: if you have 


```scala

object Foo {
    val abc = 12
    val Bar = 24
}
```

you can then match on the upper case starting `val` : 

```scala
import Foo._

val i : Int = ???

i match {
   case Bar => ???  // valid
   case abc => ??? // not valid
}
```

that way you have a meaning to the number, because it would be hard to know what `15` might mean here.

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1640,92 +1403,392 @@ class SessionData(
...
+      case _ => ()
+    }
+  }
+
+  def handleUseCooldownRenew(definition: BasicDefinition): Unit = {
+    definition match {
+      case _: KitDefinition => kitToBeUsed = None
+      case _ => ;
+    }
+  }
```
leobenkel, line 1508: for the method that are of kind:

```scala
def foo(a: A): B = {
    a match {
          case ...
          case ...
          case ...
   }
}
```

it is possible to change it to:

```scala
def foo: A => B = {
      case ...
      case ...
      case ...
}
```
and that does not change the way it is called `foo(a)` in both case.

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1640,92 +1403,392 @@ class SessionData(
...
+              case Some((v: Vehicle, _)) if v.PassengerInSeat(player).contains(0) =>
+                attemptRecoveryFromNoGuidExceptionAsPlayer(player, nge)
+                SupervisorStrategy.resume
+              case Some((_: Vehicle, _)) =>
+                SupervisorStrategy.resume
+
+              case _ =>
+                //did not discover or resolve the situation
+                writeLogExceptionAndStop(nge)
+            }
```
leobenkel, line 1589: not 100% sure but it might be possible to combine those into a `collect` and that might make it cleaner

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1640,92 +1403,392 @@ class SessionData(
...
+        player.VehicleSeated = None
+        v.Seats(0).unmount(player)
+        player.Position = v.Position
+        player.Orientation = v.Orientation
+        zoning.interstellarFerry = None
+        zoning.interstellarFerryTopLevelGUID = None
+        attemptRecoveryFromNoGuidExceptionAsPlayer(player, e)
+      case None =>
+        writeLogException(e)
+    }
```
leobenkel, line 1621: that can be `collect`

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1640,92 +1403,392 @@ class SessionData(
...
+          writeLogExceptionAndStop(e)
+      }
+    } else {
+      SupervisorStrategy.resume
+    }
+  }
+
+  def attemptRecoveryFromInventoryDisarrayException(inv: GridInventory): Unit = {
+    inv.ElementsInListCollideInGrid() match {
+      case Nil => ;
```
leobenkel, line 1652: `;` to `()`

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -1640,92 +1403,392 @@ class SessionData(
...
+      case list if list.isEmpty => ;
```
leobenkel, line 1653: same here

src/main/scala/net/psforever/actors/session/support/SessionData.scala
```diff
@@ -2324,31 +2320,26 @@ class SessionData(
...
-            None
-        }
-      case None =>
-        None
+  def findLocalVehicle: Option[Vehicle] = {
+    player.VehicleSeated.flatMap { vehicleGuid =>
+      continent.GUID(vehicleGuid) match {
+        case Some(obj: Vehicle) => Some(obj)
+        case _ => None
+      }
```
leobenkel, line 2342: that looks like a `filter`

leobenkel (changes requested):
Amazing job on the refactoring ! It looks so much better ! I added a few nit comments here and there :)

src/main/scala/net/psforever/actors/session/support/SessionData.scala
Fate-JH: I am uncertain if that will improve much.  None of the cases would change; and, I only want one result anyway.

src/main/scala/net/psforever/actors/session/support/SessionData.scala
Fate-JH: Best or worst case, `collect` goes over every entry.  `find` stops on the first positive result.

src/main/scala/net/psforever/actors/session/support/SessionData.scala
Fate-JH: Clarification: I would rather create an `Enumeration` or an `EnumType` eventually.

leobenkel: In scala 2.x , I was strongly discourage using `Enumeration`. It is a lot cleaner to use a `sealed trait` and `case object` . the compiler will be able to detect not exhaustive pattern matching that way

src/main/scala/net/psforever/actors/session/support/SessionData.scala
leobenkel: well it reads weird to have a `chain(...).chain(...).chain(...) match {...}` instead of `chain.chain.chain`

src/main/scala/net/psforever/actors/session/support/SessionData.scala
leobenkel: ok

src/main/scala/net/psforever/actors/session/support/SessionData.scala
Fate-JH: Used `IntEnumType` `case object`s.

src/main/scala/net/psforever/packet/game/GenericActionMessage.scala
```diff
@@ -2,76 +2,69 @@
...
+
+object GenericAction extends IntEnum[GenericAction] {
+  val values: IndexedSeq[GenericAction] = findValues
+
+  final case object ShowMosquitoRadar extends GenericAction(value = 3)
+  final case object HideMosquitoRadar extends GenericAction(value = 4)
+  final case object MissileLock extends GenericAction(value = 7)
+  final case object WaspMissileLock extends GenericAction(value = 8)
+  final case object TRekLock extends GenericAction(value = 9)
+  final case object DropSpecialItem extends GenericAction(value = 10)
```
Mazo, line 19: I think this one should be 11, not 10.

src/main/scala/net/psforever/actors/session/support/SessionData.scala
Fate-JH: I didn't end up using a `filter`.
Actually, I don't even recall where this was in the original.

src/main/scala/net/psforever/packet/game/GenericActionMessage.scala
```diff
@@ -2,76 +2,69 @@
...
+  final case object FacilityCaptureFanfare extends GenericAction(value = 12)
+  final case object NewCharacterBasicTraining extends GenericAction(value = 14)
+  final case object MaxAnchorsExtend extends GenericAction(value = 15)
+  final case object MaxAnchorsRelease extends GenericAction(value = 16)
+  final case object MaxSpecialEffect extends GenericAction(value = 20)
```
Mazo, line 24: The previous comment had information about if the action should be from the client, or from the server to be effective. Is it worth prefixing these with something like `Server`/`Client`?

Fate-JH: I'm not confident if that kind of organization was ever useful.  In the beginning, I just parsed ordered packet data that was just a bunch of numbers, separated only by the direction they were encountered, since they were just numbers, then tried to attach meaning to them through trial and error.  That's how the original list was created,  Now that they have names, I don't know if they need direction designation.
Maybe the ones that come from the client, as the majority of them are still packets you send to the server.

PR #1027, opened 2023-02-16
Players Properly Assigned Blame for Killing Vehicle Occupants

Fate-JH:
Chat is a little snitch.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -6,6 +6,7 @@ import akka.actor.typed.scaladsl.adapter._
 import akka.actor.{ActorContext, ActorRef, Cancellable, typed}
 import akka.pattern.ask
 import akka.util.Timeout
+import net.psforever.objects.serverobject.mount.Seat
 import net.psforever.objects.serverobject.tube.SpawnTube
 import net.psforever.objects.sourcing.{PlayerSource, SourceEntry, VehicleSource}
 import net.psforever.objects.vital.{InGameHistory, ReconstructionActivity, SpawningActivity}
@@ -1254,6 +1255,7 @@ class ZoningOperations(
     log.info(msg)
     log.debug(s"LoadZoneInVehicleAsDriver: $msg")
     val manifest  = vehicle.PrepareGatingManifest()
+    vehicle.Seats.collect { case (index: Int, seat: Seat) if index > 0 => seat.unmount(seat.occupant) }
     val pguid     = player.GUID
     val toChannel = manifest.file
     val topLevel  = interstellarFerryTopLevelGUID.getOrElse(vehicle.GUID)
--- a/src/main/scala/net/psforever/objects/Vehicle.scala
+++ b/src/main/scala/net/psforever/objects/Vehicle.scala
@@ -511,7 +511,6 @@ class Vehicle(private val vehicleDef: VehicleDefinition)
 
   def PrepareGatingManifest(): VehicleManifest = {
     val manifest = VehicleManifest(this)
-    seats.collect { case (index: Int, seat: Seat) if index > 0 => seat.unmount(seat.occupant) }
     vehicleGatingManifest = Some(manifest)
     previousVehicleGatingManifest = None
     manifest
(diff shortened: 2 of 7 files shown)


Fate-JH:
This should work up to the point where we have a mountable non-vehicle with more than one seat, or the seat is a different index.

ScrawnyRonnie:
A few findings so far. In these screenshots the NC character blows up the vehicle. "vehicle" is the driver. "passenger" is the passenger/gunner. **Unrelated to vehicle testing, players do not see the equipment of others. Holstered or equipped things, as well as helmets, are not seen.**

First one it credits the kill on the driver. The passenger survives and is stuck in the vehicle unable to get out. In this one they were both in the vehicle and then zoned somewhere else.
image

Another same test, this time not zoning while in the vehicle (to see if that had anything to do with getting stuck). Same scenario; passenger survives and is stuck.
image

Same kill, passenger's perspective
image

Again from passenger's perspective. Weirdness then happened a minute later while still in the vehicle.
image

ScrawnyRonnie:
Another issue with zoning while in a vehicle if passenger are aboard.
1) Driver and passenger enter a vehicle.
2) Use a warpgate to zone somewhere else.
3) The passenger will get stuck in an endless loop of zoning to that place every ~20 seconds or so. It will result in the passenger appearing to re-enter the vehicle over and over again.
image
4) Attempting to zone somewhere else again will crash the driver's client. The passenger will be left standing in the warpgate they were attempting to zone from. The vehicle will actually zone to the new location, but without the passenger.

ScrawnyRonnie:
Just a side-by-side making sure the equipment/helmet invisible thing didn't only happen in a vehicle.
image

Fate-JH:
Check if either of these issues existed before this PR.  Create a new ticket linking to this one, if so.

ScrawnyRonnie:
Tested on our "main" (51000) server and none of these issues are present.
- Passenger dies properly when the vehicle is destroyed and can respawn.
- Zoning causes no issues with the driver or passenger.
- Equipment players are wearing is visible to others.

Fate-JH:
Okay, the process has been amended.  Please test vehicle gating.
Also, please provide more context - reproduction - to players not being able to see each others's equipments if it persists. Did not internally test cosmetics - hats and such, but armaments seem fine (now).

ScrawnyRonnie:
I was going mad because of apparent inconsistent results and come to realize these additional changes may not have been needed to begin with. First test I see the same issue would be happening with a gunner getting stuck in that zoning loop, then it would stop happening the next time and all passengers/gunners would be fine. I made batches of characters multiple times and went through this cycle of it happening and then not. I think I pinpointed the whole cause of the issue though. Here goes...

On character creation you are BR 1. To cert up some vehicles (I know you start with some, but wanted to transport multiple characters) I would use the /setbr command. Before doing this, the helmet of the player I did this on was visible. After using the command the helmet would still be visible, but only until zoning or after a crash and logging back in it would then be invisible. During this time where the helmet wasn't showing properly based on the player's helmet toggle the zoning issue would happen, **but after using the /helmet command to appear correctly to other players it stopped happening and zoning was fine for all in the vehicle**

Could use another person to test this out to see if my thought process is right here. If it is the setbr command causing the issue with the helmet/etc then we _shouldn't_ have issues if this is merged.

I haven't gotten to blowing up vehicles yet and testing what happens there. I was stuck on this issue for far too long already.

ScrawnyRonnie:
Pretty convinced now it is the use of /setbr that causes the zoning issue, but once using /helmet everything seems fine with that.

I did test destroying a vehicle too. Both occupants died and it showed them in the feed. Still credits the killer just one kill as far as bep/cep - that is another old issue that I created. Would like to get some people on to do more testing.

Fate-JH:
It's probably just a default flag issue.  When a player reaches the BR where cosmetics can be applied, only then do the cosmetic flags start to affect object create packets.  I'll transition that into a different issue/ticket.
I'd like to close this one and update the server with it before the weekend fights start in proper.

Fate-JH:
I also included a fix for wall turret hotspots not showing up.  The map function for timing worked by recognizing short-hand object types and the type of object that is a wall turret changed.
I also added a generator hotspot, because it's that important.

PR #1033, opened 2023-02-20
Neutral Hotspots

Fate-JH:
Reduces combat. (It feels that way if you look at the map.)

### DIFF
--- a/src/main/scala/net/psforever/objects/zones/HotSpotInfo.scala
+++ b/src/main/scala/net/psforever/objects/zones/HotSpotInfo.scala
@@ -2,7 +2,6 @@
 package net.psforever.objects.zones
 
 import net.psforever.types.{PlanetSideEmpire, Vector3}
-
 import scala.concurrent.duration._
 
 /**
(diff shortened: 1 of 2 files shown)


ScrawnyRonnie:
Hotspots appeared and disappeared for the involved empires. Did these tests:
- Neutral base turret and gen for the empire that damaged them
- Enemy base turret and gen for the owning empire and attacking empire
- Shooting an enemy character all good as well
The fireworks were fun while they lasted

Dethdeath:
Tested this too for good measure, everything seems fine.

ScrawnyRonnie:
For some reason the hotspot issue is back. Shooting a neutral turret doesn't make a hotspot again until you shoot an enemy owned one, then both and any additionally created ones are stuck forever.

Dethdeath:
Can confirm, we did various tests last night. I shot all the wall turrets at 2 bases on Ceryshen (not neutral) and only 2 of these hotspots remained after like 5 minutes, they never went away after that. So it can be a pretty random result, which is why we thought it was fixed initially.

PR #1034, opened 2023-02-22
You Can (Not) Hold

Fate-JH:
I was hustling to resecure the CC
I crashed to the desktop and couldn't log back in without crashing again
"Wait for a bit, your med app is to blame"
Man, I don't want my persistence to end

I threw it on the ground
I'm just a granny hacker, I don't have time for this
To think I spent points on this certification
Man! Gotta pump this garbage into my own arm

### DIFF
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -1446,9 +1446,8 @@ class AvatarActor(
                     name,
                     cooldown.toSeconds,
                     item match {
-                      case t: ToolDefinition => GlobalDefinitions.isMaxArms(t)
-                      case _: VehicleDefinition => true
-                      case _ => false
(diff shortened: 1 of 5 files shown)


ScrawnyRonnie:
Butterfingers indeed! Really can't hold onto those things. Both slots and a full inventory of them spilled to the ground. Fortunately I was able to pick them all back up and smuggle them to my destination when I logged back in.

ScrawnyRonnie:
I did also try gracefully logging off and back in - med app still equipped
"crashing" and waiting long enough for persistence to end - med app still equipped
It does only drop while persisting so no crashes for me. It is much less of an inconvenience to pick it back up than to have to wait so I'd call it a win.

Dethdeath:
Does what it says on the tin. Maybe in the future we can get something better than it being dropped.
This is 100 times better than a crash though.

PR #1040, opened 2023-02-28
No Uniform, No Helmet, No Service

Fate-JH:
For the longest time, the player avatar has been able to set cosmetic items at battle rank 24, but player characters haven't been able to depict cosmetics below battle rank 25.  This is tied to the third and final uniform upgrade.  Mixing uniforms and cosmetics in the wrong pattern causes player characters to be seen without their firearms.  For a while now, too, there's been three known uniform upgrades from the normal one and two unknown uniform upgrades that are mirrors of other nearby upgrades.  While one of these previously-unknown upgrades continues to remain unknown, the matter of what to do about cosmetic appearance at battle rank 24 has been handled.

Thanks to Scrawny and Untouchable for discovering this issue with the latter's dress.

Please break fewer things.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -806,7 +806,13 @@ object AvatarActor {
     out.future
   }
 
-  def toAvatar(avatar: persistence.Avatar): Avatar =
+  def toAvatar(avatar: persistence.Avatar): Avatar = {
+    val bep = avatar.bep
+    val convertedCosmetics = if (BattleRank.showCosmetics(bep)) {
+      avatar.cosmetics.map(c => Cosmetic.valuesFromObjectCreateValue(c))
+    } else {
+      None
+    }
     Avatar(
       avatar.id,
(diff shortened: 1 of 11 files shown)


ScrawnyRonnie:
Simple test in sanc worked.
image
Weapons and helmet were also visible before changing the cosmetic appearance options.

Would any further scenarios need to be tested? Zoning in vehicles with BR 24 or > passengers/drivers?

Dethdeath:
Both
Saw a double selection pop up right after being promoted to BR 24, went away after actually picking something. 

A character appeared to be wielding no weapons after zoning, this was fixed when they swapped armor at a terminal.

Dethdeath:
As Scrawny mentioned, the rexo client freeze #1045 still occurs on this build, but using the /helmet command prevents it from happening.
Does additional code need to be added to prevent the freeze?
Maybe force use the /helmet command when players first reach BR24?

Fate-JH:
Still have to go back over this PR and integrate the new information.

Fate-JH:
Should be better.

PR #1044, opened 2023-03-03
Scorpion Death Message

Fate-JH:
Corrects incorrectly attributed blame for death by Scorpion little buddy sub-projectiles, thus that the player who fired the weapon is blamed in the death message and not the sub-projectiles.

Maybe.

I don't know.

You couldn't pay me to stand out in the Ceryshean tundra and fire the weapon on the remote possibility that it hits something.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionData.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionData.scala
@@ -2434,6 +2434,7 @@ class SessionData(
       case obj: PlayerSource => obj.Seated
       case _                 => false
     }
+    sendResponse(ChatMsg(ChatMessageType.CMT_OPEN, wideContents=false, "", s"Killer method: $method", note=None))
     new DestroyDisplayMessage(
       killer.Name,
       killer.CharId,
--- a/src/main/scala/net/psforever/actors/session/support/SessionData.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionData.scala
@@ -2434,7 +2434,6 @@ class SessionData(
       case obj: PlayerSource => obj.Seated
       case _                 => false
     }
-    sendResponse(ChatMsg(ChatMessageType.CMT_OPEN, wideContents=false, "", s"Killer method: $method", note=None))
     new DestroyDisplayMessage(
       killer.Name,
       killer.CharId,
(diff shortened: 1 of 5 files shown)


Fate-JH:
I want to copy something from my original PR for the Scorpion, emphasis on the last sentence:

"The Scorpion is not designed to be as hard to use as possible but it certainly is the hardest weapon to use out in the field as a soldier utilizing it must be able to process unreasonably accurate distance calculations. Only for those of elite skill (and certification), it would only ever be viable in the hands of blatant hackers and trolls who would rather use other cheese anyway. **I seriously only damaged targets with the bomblets twice during two-plus weeks of controlled testing**."

ScrawnyRonnie:
Managed to get a few insta-kill shots in (characters were full or slight damage to armor), which was nice and difficult to do. This was done by aiming higher and how I was used to using this weapon on Gemini. For those kills it showed a skull instead of the weapon. Also just one kill credit - each time all died from the same shot.
image
Aiming lower and it chipping away at their health it showed the weapon. Also did /suicide while the last shot was enroute to test that.
image
So the thing that does enough damage to kill someone from full is a different part of the projectile? The initial kill and screenshot was that insta-kill part of the blast/explosion that showed the buddy message. Looking back I notice it had a skull there too.
image

Fate-JH:
I don't know why the different results between OHKO and multiple shots but I think I can fix the issue with a quick change.

Dethdeath:
squad
Got this upon relog earlier.

Fate-JH:
Give it a try now.

Fate-JH:
> squad Got this upon relog earlier.

Well, do you accept?  Don't keep lsmsltu hanging.

ScrawnyRonnie:
One hitter quitter still a skull. Maybe it should just stay a skull due to how epic it is to kill someone like this instantly.
image

Fate-JH:
Since you've been so adept at handling this weapon, Scrawny, can I ask you to give this another shot?
For now, when you kill a player, a message in chat will report a number.  Please murder someone in one shot with the Scorpion, then murder them again but with multiple Scorpion shots, and report the numbers that appear in chat next to the message "Killer method:".  That should shed some light on what the skull and crossbones is supposed to be.

ScrawnyRonnie:
image

ScrawnyRonnie:
Killed by avatar?

Fate-JH:
I didn't want to explicitly create a new `DamageReason` (what it says on the tin) for the little buddy projectiles but it became necessary due to a level of indirection that is different from standard `ExplodingEntityReason` causes.

I also tried it out myself to check that OHKO damage message was sufficiently formatted.  I never want to see this weapon again.

PR #1048, opened 2023-03-07
Login Inconsistencies

Fate-JH:
When logging back into a character whose persistence had previously ended and the character did not die:
- armor will not default to full or to 50
- all weapons in the character's hands and in their backpack will have the same ammunition available as when that character logged out
- all ammunition boxes in the character's backpack will have the same quantity available as when that character logged out
- all weapons and ammunition boxes in the character's locker will have the same ammunition available as when that character stowed them

Todo:
- ...

### DIFF
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -257,19 +257,26 @@ object AvatarActor {
   }
 
   /**
-    * Transform from encoded inventory data as a CLOB - character large object - into individual items.
-    * Install those items into positions in a target container
-    * in the same positions in which they were previously recorded.<br>
-    * <br>
-    * There is no guarantee that the structure of the retained container data encoded in the CLOB
-    * will fit the current dimensions of the container.
-    * No tests are performed.
-    * A partial decompression of the CLOB may occur.
-    * @param container the container in which to place the pieces of equipment produced from the CLOB
-    * @param clob the inventory data in string form
-    * @param log a reference to a logging context
-    */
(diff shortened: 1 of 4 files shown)


Dethdeath:
Seems to be working fine.

Fate-JH:
Modifying lockers to also perserve ammunition for weapons and ammo boxes between storage should be important because it places emphasis on looting.  Enemy equipment is ignored in loadout selection and this should be seen as eliminating other ways to freely print foreign ammo such as plasma cells and the various AV missiles.

ScrawnyRonnie:
Also did a quick test and it worked.

For the locker todo, maybe this is assumed in what is to be implemented with that, but locker items aren't preserved between logins. It looks like this was addressed at one point (https://github.com/psforever/PSF-LoginServer/pull/1005). I can't say when they stopped being persistent because I don't use them.

Dethdeath:
I use my lockers pretty often, I think they're fine?

ScrawnyRonnie:
I'm not sure what Deth did to get this VIP access to the locker persistence database, but my characters don't retain locker items between sessions. His do (minus the new test char he made) so we're curious why that is. 

Our characters on this test server have empty lockers at the next login. I just tried again after the recent updates to this PR.

Fate-JH:
Lockers should work again now.  If it doesn't for an existing character (which means that character is probably out of commission permanently) try creating a new one to test with.

Dethdeath:
Made a new character on 2 accounts, locker goes empty between sessions each time.

Fate-JH:
It's a good thing I asked you to try with new characters.

ScrawnyRonnie:
Houston, we have lockers.

Tested a new char and a char that had been created earlier on this test server. Both retained their locker items between sessions.

Dethdeath:
Yep, they work fine here as well.

PR #1050, opened 2023-03-14
Code Style Improvements 2

Fate-JH:
For previous, see: [PR 1018](https://github.com/psforever/PSF-LoginServer/pull/1016#issue-1559218004)

Only cosmetic and performance changes to the indicated files, with two exception: extending the range of voice macros, and a helper function for a future expansion.

Do test a few things, mainly things with vehicles, including the HART and droppods.  I was extremely uncomfortable when breaking up the `CanMount` message like that.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/ChatActor.scala
+++ b/src/main/scala/net/psforever/actors/session/ChatActor.scala
@@ -5,13 +5,6 @@ import akka.actor.typed.{ActorRef, Behavior, PostStop, SupervisorStrategy}
 import akka.actor.typed.receptionist.Receptionist
 import akka.actor.typed.scaladsl.{ActorContext, Behaviors, StashBuffer}
 import akka.actor.typed.scaladsl.adapter._
-import net.psforever.objects.avatar.{Shortcut => AvatarShortcut}
-import net.psforever.objects.definition.ImplantDefinition
-import net.psforever.packet.game.{CreateShortcutMessage, Shortcut}
-import net.psforever.packet.game.objectcreate.DrawnSlot
-import net.psforever.types.ChatMessageType.{CMT_GMOPEN, UNK_227}
-import net.psforever.types.{ExperienceType, ImplantType}
-
 import scala.collection.mutable
 import scala.concurrent.ExecutionContextExecutor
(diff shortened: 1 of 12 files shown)


ScrawnyRonnie:
Voice macros are picked up by other chars up to ~40m or 50m. Had different results depending on the character being in a vehicle or not.

HART seems fine. Used it a few times.

Zoning in vehicles is not working. The passengers and/or gunners are left in the warpgate and never zone. The pilot traverses just fine, but they are alone in the vehicle on the other side. 

With the stuck passengers I was able to exit the vehicle or use /zone to force them out. Returning to the warpgate they could see the same vehicle multiple multiple times, once for every time it zoned with them in it.
image
After zoning to another cont and going back, all of those phantom vehicles were gone.

Fate-JH:
Someone else mentioned that vehicle zoning was broken before this PR - in the ways that you mentioned, I believe - so I need to look into that issue on a larger scale anyway.

All I did was extend the range to 40m from 25m.  Between this PR and that other PR with where I fixed case sensitivity, I can confidently say that being seated in a vehicle does not go into our considerations for whom to dispatch the `ChatMsg` packets for voice macros.  I am not certain about your testing but maybe the distinction is the kind of seat (closed / open)?

Dethdeath:
Didn't see anything else that's broken.

ScrawnyRonnie:
The issue I mentioned with vehicle passengers not zoning does not happen on the main server. I double checked and tested again to confirm and was able to zone with multiple passengers on 51000. Yes, there are some issues there with vehicles and zoning, specifically with cargo vehicles, but this is new behavior not exhibited there. As you said vehicle zoning was already on our radar, but I think has done something unexpected.

Fate-JH:
Understood, but I don't think I can only half fix something.
I can hold off on updating `SessionGalaxyHandlers` for this update as that class includes essential passenger gating routines.  If I take that out, will the build match 51000 behavior, I wonder.

Fate-JH:
Changes to `SessionGalaxyHandlers` rolled back, but also retained for ease of reference.

ScrawnyRonnie:
All good with passenger/gunner zoning.

PR #1051, opened 2023-03-14
/tell someone

Fate-JH:
Soon, we'll have advanced heuristic analystics in place and players won't even need to know how to spell each others's names properly to use any of the commands.

I admit to causing only part of this issue.


Fate-JH:
Hey, grandpa, we call that "predictive text".

PR #1052, opened 2023-03-17
Reduced Upstream/Downstream Load

Fate-JH:
Each of twenty avatars in the same zone would receive twenty upstream packets (`PlayerStateMessageUpstream`) from their own clients to the server.   The server must redirect the information from the `PSUM` packets from the avatars to the other nineteen players, and the one that arrives back at on one's own doorstep is ignored on arrival.  Each of the twenty recipients of nineteen messages will, at worst, bundle all of this data into nineteen downstream packets (`PlayerStateMessage`) and dispatch these packets to their clients.  Every individual clients's 250 milliseconds, another upstream to downstream cycle will be initiated.  Unbundled, just slightly under four hundred packets are expected to fly around the network at any given moment for this scenario.  Other packets other than `PSMU` and `PSM` will be dispersed in the network, moving between client and server and server and client, in a similar fashion.

In address of the postulation that the constant flow of upstream packet traffic is causing a frustrating bottle neck of downstream packet traffic for the client, and that this congestion is an inevitable strain that will cause **disconnect within fifty minutes of operation in some form of client panic**, I have introduced the following attempted solution.

While it is customizable from Config fields, the basic setup right now follows three rules:
1) As a target player reports activity that is moved farther from an avatar, that avatar is allowed to skip one or several client updates for that player depending on a mapping of distance to tolerable delay.
2) As the number of players within the draw distance increases beyond a minimum value, for each additional groups of players, a number of meters are added onto the distance calculations for the above time delay.  This means that the delay will become greater in a significantly populated area even if the distance between players does not change.
3) When a target player has moved or been positioned outside of the draw distance for an avatar, or has been manipulated in some way that disrupts its visibility to that avatar, that player will be moved far away from any region of interest in the game and will not send additional updates until moving back or being moved back into a drawable distance and entering a drawable state.  The states affected only involve spectator status at the moment, not cloaking or radar.

This staggering of packets shall be known as an upstream-to-downstream ratio.

With this change, there needs to be a clarified rule about visibility: an avatar can see another player at 1:1 upstream-to-downstream ratio regardless of distance if:
1) either player is holding a Sniping weapon (`heavy_sniper_rifle` or `bolt_driver`);
2) either player, on foot, is holding a weapon that causes damage with its projectile, is crouched, and also has an implant called Range Magnifier active; or,
3) either player is seated at the gunner position of some vehicle and also has an implant called Range Magnifier active

Exposure to active radar is not factored into this previous determination due to simulation limits.

This draft pull request is for testing, not for merging.  That said, changes to the configuration fields will require a new commit as modifying these parameters for the test server is not easily feasible.


Dethdeath:
Did a test with 18 clients, still got the 50 minute DC. The characters were all standing near each other on Sanc. Perhaps there is a change with larger distances now, further testing should reveal if that's the case.

Fate-JH:
I would say that mean the test is a failure since I had conditions that would disable packet-sending even if you had a bunch of people stand close together, with few to none moving; but, I see I made an oopsie with the code.  Any player that isn't moving and is sending the exact same packet data every turn should just be passed up on sending out an update because the data is the same ... or it would be, if the ever-changing `seq_time` wasn't one of the fields in the message being saved.  I need to blank that.

Fate-JH:
I needed to test it on my end to make certain it was working correctly but I adjusted the parameter thus that the server won't spam `PSM`s from other players that are not moving, even if they are close.  The feature which would make this rule most visible should be the latency counter when hovering over another player.  The millisecond counter should go up and into the red on any stationary target, I believe.

ScrawnyRonnie:
50 minute standing still test passed. Was over an hour with no issues.

Then did a test with all chars running against the same corner of a base wall. This did make the packets appear in the capture. It did cause the issue, but at varying lengths of time. They started to DC after a little over an hour, but each time it was spaced out and not even all chars had DC'd yet after about 1:20 into this test. Progress for sure.

ScrawnyRonnie:
I should rephrase that a little cause both tests lasting over an hour makes it sound like the first one didn't have enough time and may have still had the issue. These were back to back in the same base without logging out between tests. I just took note of the time when I started moving the characters, so in total it was over two hours of 19 characters in the same area, only seeing the issue after 50+ minutes of movement.

Fate-JH:
Respawning players will treat other players in the same zone as if those other players were beyond the infantry draw distance and will thus receive no update packets from them until completely respawned.  This hopefully means that the cycle of death and respawning will contribute to staving off off the 50 minute issue.

ScrawnyRonnie:
I did another test using a router. The characters that I did /suicide on also DC'd. I did both 5 second and 15 second spawn tests. For the 15 second ones, I did /suicide on 6 characters, 1x on one, 2x on another, and so on up to 6x on the last. The upstream packets did stop during that time on the map selecting a respawn location. While those packets stopped, others did not. For example, nothing is happening at the top, then players started going through the pad.
image
Packets were also still sent while other actions, such as shooting or reloading, were going on. 

So either it is more than just the psmu packets or the time it takes to respawn is not enough time to prevent it from happening. Additional testing can be done to narrow it down:
- All players just running in place with no other actions (so other packets aren't sent during respawn) and /suicide then respawning in normal time on a few, and sitting on the map for a minute on a few others to see if a certain "cooldown" duration is needed
- Same/similar test as I did today, but also just sitting on the respawn map for longer to rule out duration again, to clear the other packets as potential suspects

Fate-JH:
For this test, the only thing that I am suppressing are `PSM` packets and the main thing I am doing to "hide" players who are too far from the avatar as a viewer, or while the avatar's own character is dead and released, is move the others to the far reaches of the zone for that avatar's client.  There's no other state trickery involved - the other players need to be valid at all times so it's a simple matter of just making them pop back into visible range with the packets.

Fate-JH:
Also, it a bit disaapointing this did not affect the 50 minute issue as intended.

ScrawnyRonnie:
I did another same test, but this time I had every character respawn. I waited 20 seconds on the map, then the countdown was an additional 10, so with a 30 second break they still DC'd. I suppose that is better because if it just needed a longer period of time, then another solution would be needed. Are we now looking at those other packets that would also send while just standing still before this PR? The SlottedMetaPacket and MultiPacketEx possibly because those were still being received with these latest moving tests, but were no longer received during the standing still for 50 test that passed after your changes that suppress packets received from still avatars.

Fate-JH:
> Are we now looking at those other packets that would also send while just standing still before this PR? The SlottedMetaPacket and MultiPacketEx possibly because those were still being received with these latest moving tests, but were no longer received during the standing still for 50 test that passed after your changes that suppress packets received from still avatars.

Not much we can do there.  `SlottedMetaPacket` and the `MultiPacketEx` are just the normal vehicle of transit for (most) packets originating from the server, and from the client.  They're just wrappers.  In this case, the `PlanetsideAttributeMessage` packets and the `GenericObjectActionMessage` packets are within those wrappers and were unfurled when they arrived on the client.  The only reason they show up in the logger at all is because it attaches just after those wrapper packets pop off an internal queue and the resulting decomposed packets are pushed back into the same queue for later processing.  Due to the way the networking protocol for the server was rewritten, I would have to completely rewrite it again just to send individual unwrapped packets to the client.

As to what those `PAM` and `GOAM` packets are doing: I guess they originate from router teleporter behavior.  The `PAM` is something and the `GOAM`s are the effects applied on either end of the teleportation.  To be honest, things like this could be suppressed and character state would be preserved when unseen.  I just don't think I can flag something this specific as "nonessential" easily.

ScrawnyRonnie:
Maybe this will help. I tested this again, but on some characters I left the area and didn't return. I had 19 characters total. After about 40 minutes I left the area on 3 of them and stayed far away for the remainder of the test. All 3 of them also DC'd along with the bunch at Zotz. The max range for receiving player state messages no longer seems to be respected on 61007. Player locations circled.
image

ScrawnyRonnie:
> I needed to test it on my end to make certain it was working correctly but I adjusted the parameter thus that the server won't spam `PSM`s from other players that are not moving, even if they are close. The feature which would make this rule most visible should be the latency counter when hovering over another player. The millisecond counter should go up and into the red on any stationary target, I believe.
> 
> This is the test that I had hoped be performed.

I tried something slightly different with this. The 'not sending updates if not moving' would cause an issue with someone coming into range or after respawning not seeing still players as seen in the videos. If that change was just intended for testing purposes then ignore me.

https://user-images.githubusercontent.com/73139382/227731827-9943c225-2c53-46f2-b740-d03c6f97ca9a.mp4
https://user-images.githubusercontent.com/73139382/227732153-788d7ff3-57f4-4c3c-b13c-ecbf4da8c83e.mp4

PR #1054, opened 2023-03-26
50 minute disconnect fix/workaround

jgillich:
And a few misc changes


.devcontainer/Dockerfile
Fate-JH, line 1: Please terminate this file with a blank line.

PR #1055, opened 2023-03-27
Last Ditch Bail

Fate-JH:
If your friends all bailed out of a Galaxy, would you do it, too?


Dethdeath:
Works fine, tested regular passenger, gunner, MAX and vehicle. Only thing is that the vehicle in the cargo hold takes significant fall damage, as if it doesn't receive the vehicle bail status?

PR #1059, opened 2023-04-15
fix docker build

jgillich:


GeekOfWires:
Just wanted to toss an FYI that I couldn't get the docker image to build on my Mac w/o making the change

PR #1061, opened 2023-04-17
Working BANK

Fate-JH:
I see how this works.
Or didn't work.
How it pretended to work.
Why did it ever work?
I mean, how did it work after it stopped working?
What even is working?


ScrawnyRonnie:
Approved. Tested the ways that caused it to break and it now works. Thanks for the quick fix.

Issue #1053, opened 2023-03-24
Instant Action with no Hotspots


ScrawnyRonnie:
Instant Action will work even if there are no hotspots. Here is the current behavior:

This only applies to and happens with **home conts**; with neutral conts it does not happen.

- If your character is on any home cont and your empire owns a base or tower, IA will send (keep) you there. This happens even if the below conditions are also met.
- If you are on a neutral cont or in sanc, an enemy player is on a home cont, and your empire owns a base or tower on that cont, IA will send you to the cont the enemy player is on.
- Same as above, but a friendly is on a home cont.

ScrawnyRonnie:
After Sunday's fight and the server being restarted to apply the changes for this issue, it is still happening. I did discover how to make IA send you to or keep you on a cont. It still requires a player to currently be on that cont, but the other prerequisite is for a facility or tower to change ownership. 

For example I went to Esamir with both an NC and a TR character. NC owned most all bases and towers; TR owned none. IA did not work and showed the message "no hotspots..." on both chars. I then hacked a tower with my TR character (actually hacked, didn't use the /capturebase command). After the tower flipped to TR, IA then worked on both chars and sent the NC to the base right near the tower.

Did the same thing on Desolation to confirm and had the same results.

I then went to Nexus and tried using /capturebase instead, but it did not work. I then hacked a tower normally and IA kept us there, so there is something different between the command and actually hacking a facility that causes IA to do this.

ScrawnyRonnie:
This may or may not help narrow it down, but it also only happens with either hacking a tower or a hack & hold type of base. Capturing an LLU did not cause it to happen, so it isn't strictly ownership of a base/tower changing that makes IA do this.

PR #1065, opened 2023-04-20
Instant Action, Just Add ...

Fate-JH:
I changed the algorithm for determining where to instant action a player, based on feedback that the original determination could be wrong after a while.  I'd like to think I didn't change anything, but I think I did.  If I check the original code, there was a big issue with adding up combat heat that had nothing to do with allies and spawning on that point regardless of viable allies in combat situations.  To be honest with you, I've not tested the new code because my tests would be quaint and simplistic, even if effective at all.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -67,8 +67,13 @@ database {
 
 # Enable non-standard game properties
 game {
-  # Allow instant action to AMS
-  instant-action-ams = no
+  instant-action = {
+    # Allow instant action to direct a player to a hotspot-local friendly AMS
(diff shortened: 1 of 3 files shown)


ScrawnyRonnie:
I did a quick test to make sure instant action was still functioning correctly with hotspots, which it is. I couldn't get IA to happen once the hotspots were gone. This one may just need to go live to see if it happens again because we're not exactly sure how to induce the issue in https://github.com/psforever/PSF-LoginServer/issues/1053. If it doesn't happen after the Sunday fight (seems to come up on conts after a fight happens) I'll close the issue.

PR #1069, opened 2023-04-25
Deconstruct in Privacy

Fate-JH:
When deconstructing at any spawn tube - in a facility or at an advanced mobile spawn - the player will be concealed thus that they can be allowed to select their destination and complete the passage without interference by external forces.

This utilizes the changes in "Reduced Upstream/Downstream Load" (a1cf6c27018109c7909dbb8ca591a2e56fad4e91) to stop any player that attempts to deconstruct from being visible to other players.

___Addenda___
1. The calculation for determining whether or not the `PlayerStateMessage` packet will be sent - the aforementioned "Reduced Upstream/Downstream Load" - has been simplified out of necessity to eliminate some issues discovered during deconstruction testing.  Play around with it a bit, though I think walking through the simplicity of the remaining checks should speak for themselves.
2. Various allowances have been included in the off-chance that the deconstruction process is cancelled before making a destination selection.  If the player regains control and the deconstruction will not happen, they shall not be permanently hidden from other players.  These triggers include basic movement, drawing weaponry, and implant activation.  As should be expected, this does mean that attempting to deconstruct will lower one's drawn weaponry (the process already messed with active implants).
3. I don't even remember what `death_by` is supposed to have been used for in the first place.  I set it to -1 when the player is to be kicked, I set to 1 when the player is to be protected from all damage, and it's normally 0.
4. I am not going to force the spawn tube door closed during the deconstruction process.  It's not worth it.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -163,7 +163,7 @@ class SessionAvatarHandlers(
           case term: Terminal with ProximityUnit => sessionData.terminals.StopUsingProximityUnit(term)
         }
         if (sessionData.zoning.zoningStatus == Zoning.Status.Deconstructing) {
-          sessionData.zoning.zoningStatus = Zoning.Status.None
+          sessionData.stopDeconstructing()
         }
 
       case AvatarResponse.ObjectHeld(slot, _)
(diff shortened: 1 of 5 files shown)


Dethdeath:
I recorded this earlier, before the latest changes were made:

https://user-images.githubusercontent.com/77748540/234180056-eb626387-5f80-4136-85c1-7b56cc2377f2.mp4

In the latest build the first method still works, the boomer method no longer works, but if you reveal the player first by walking into them you can still blow them up afterwards.

Fate-JH:
Player no bump.

Dethdeath:
Very good, couldn't find a way to kill the deconstructed player, which is how it should be.

PR #1071, opened 2023-04-26
Unintentional Explosive Ordnance Disposal Suit

Fate-JH:
*shakes fist* This guy ... !
boom


ScrawnyRonnie:
Seems good and had no instances where the player did not take damage or die from mines or boomers.

Dethdeath:
Good, also had nothing unexpected happen.

PR #1072, opened 2023-04-27
Vehicle Gating with Passengers

Fate-JH:
When a ferrying vehicle passes through a warp gate, its passengers will be able to disembark safely on the destination side by remembering that they actually count as the "passengers" of the vehicle.  Additionally, the ferrying vehicle will not start assuming that it is being ferried inside itself, thus locking its passengers into their seats.

Addenda:
1. Increased the frequency that the `PlayerStateMessage` retention array is cleared (to "between lives" from "between zoning") in an effort to eliminate instances of stationary players being invisible to each other.
2. Expanded various instances that might counter a `MatchError` that could arise and threaten stability.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/ChatActor.scala
+++ b/src/main/scala/net/psforever/actors/session/ChatActor.scala
@@ -717,6 +717,7 @@ class ChatActor(
                   sessionActor ! SessionActor.SendResponse(
                     message.copy(messageType = UNK_229, contents = "@CMT_ZONE_usage")
                   )
+                case _ => ()
               }
 
             case (CMT_WARP, _, contents) if gmCommandAllowed =>
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -5,6 +5,7 @@ import akka.actor.typed.scaladsl.adapter._
 import akka.actor.{ActorContext, typed}
 import net.psforever.services.Service
 
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 //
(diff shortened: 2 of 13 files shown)


Dethdeath:
The passenger desync problem is fixed.
The invisible player issue also seems fixed from what I can tell.

ScrawnyRonnie:
I had visual issues with passengers after gating. The aircraft appeared to keep rising up on their screen, then reset to proper location and do it again. Only happened with passengers, not gunners. It stopped after a minute or so. Would like to test again later tonight before closing this.

Fate-JH:
> The aircraft appeared to keep rising up on their screen

Wait, rising up, not falling?

ScrawnyRonnie:
It actually happened before I even used a warpgate, so gating isn't part of this. It did it in a Galaxy as well. Seems to happen (and not every time for whatever reason) after pressing 'E' to raise the aircraft up.
https://user-images.githubusercontent.com/73139382/235004704-f465312b-2e05-4ea7-9325-4bff354c7027.mp4

Earlier and just now I had one instance where a passenger ends up falling underground on another part of the map, not in the vehicle. They are desynced when it happens.

Fate-JH:
I've been darting around in a Galaxy, a pilot and a passenger, attempting to encounter the same fate as Scrawny on my local build. I can only manage two player character tests for now. Haven successfully mounted, dismounted and bailed, risen and descending, and vehicle gated without issue for some time now, I find myself confused at the behavior of your Phantasm and the mysterious gravity that keeps it on the ground.  (I use L.Shift and L.Ctrl for raise and descend, respectively.)  The two passenger perspectives that can not cancel the vertical velocity of the vehicle - suggesting their `VSM` packets have stopped updating properly shortly after the Phantasm began rising off the ground, maybe - is strange as this is not something I changed, accelerated, or withheld as I have with `PlayerStateMessage`.  This issue rather consistently affects the passenger seats on flying vehicles? do you have any other reproduction steps or are the players who would be pilot and passengers interesting in any way?

The desync is just quirky.  Don't feel I can do much with that anecdote.

ScrawnyRonnie:
Can probably conclude it is just something odd on my end with multiple clients. I instead just ran one client on my PC and the other on a laptop and did not experience the rising aircraft issue.

I also did a few tests with the idle characters and they are now visible.

PR #1088, opened 2023-05-19
Misc Fixes Again

Fate-JH:
1. Battleframe robotics vehicles should restore control to the driver after being spawned.
2. Dying while in a vehicle no longer has the chance of spawning you in whatever destination is currently closest to the world origin.  You should be sent to the closest spawn that matches your selection.


Dethdeath:
1. Fixed.
2. Fixed.

PR #1090, opened 2023-05-20
Both Hands on the Wheel

Fate-JH:
Apparently I need to remind people that they can not perform driver-by shootings in this game with a piece.
Holster your junk away before getting a vehicle.
Oi.


Dethdeath:
This seems to resolve the issue.

Issue #1085, opened 2023-05-11
Vehicle wrecks appear on the minimap and have full health


Dethdeath:
![wreck](https://github.com/psforever/PSF-LoginServer/assets/77748540/9ab5d3c9-8c0f-43f3-99f7-7b54d1932c6c)

https://github.com/psforever/PSF-LoginServer/assets/77748540/dee7f05c-0938-4169-93cd-8cc90122a2d5

It appears when you pull a new vehicle and it seems like the vehicle's previous owner has a higher chance of seeing the wreck like this, others don't always see its healthbar. For some reason this stuff doesn't show while on foot.

Fate-JH:
Has there been any change in this behavior?

Dethdeath:
Still the same.

ScrawnyRonnie:
I think this has to do with a vehicle becoming ownerless. With vehicles that are not destroyed, as soon as the owner leaves the zone or becomes the owner of another vehicle, the previous vehicle looks like it is heavily damaged (sparking/lights flashing), but the health bar shows full. If it is damaged, the health bar doesn't go down.
![image](https://github.com/psforever/PSF-LoginServer/assets/73139382/67f8ad0c-f85e-47b4-a077-4beef190be99)
The player on the right was the owner and pulled a new vehicle.
The player on the left side saw it in this state as well, but after leaving the zone and returning that fixed the model/health bar.

PR #1091, opened 2023-05-21
May 20th Event Fixes (maybe)

Fate-JH:
There was a significant `MatchError` caused by the introduction of an `AutoPilot` state to indicate a spawning vehicle; that was 
resolved.

Attempted to restore Kanon telemetry but undoing the changes @jgillich enacted.  I have no clue if that's all that is needed.  There were plugin issues that may or may not be related to `sbt-kanela-runner`.

Attempted to resolve health reporting issues with vehicles after noticing that they tend to break when switching to a new vehicle for ownership.  It's been previously demonstrated that this breaks the model of the vehicle that was previously owned by setting it to its nearly destroyed condition.  This condition also causes the health bar state of the vehicle to break and it will always display as being full health.  By eliminating the previous packet that makes it "unowned" the visual state of the vehicle is no longer set to its trashed state but the health reporting is still unresponsive.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-import xerial.sbt.pack.PackPlugin._
+//import xerial.sbt.pack.PackPlugin._
 
 lazy val psforeverSettings = Seq(
   organization := "net.psforever",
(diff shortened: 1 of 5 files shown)


Dethdeath:
Is #1085 also related to the vehicle health issue?

Fate-JH:
> Is #1085 also related to the vehicle health issue?

It's possible but I didn't get a chance to test before this was merged.  Vehicles must be wreckage when their health is 0.  Similarly, players can only be dead bodies when their health is 0.

PR #1096, opened 2023-05-26
Upgrade to scala 2.13.10

WtzLAS:
idk why but it just works :(
Tested with Zulu 19


wolf109909:
I confirmed it working on my pc so weird

PR #1105, opened 2023-06-02
Reduce Load More

Fate-JH:
After the success of #1052, I have built upon the foundation established in reducing `PlayerStateMessage` packets to reduce other unnecessary outbound packets based on the visibility of a target player.  All of these changes affect weaponry.

___Features___
__`ChangeFireStateMessage_Start`__
Normally causes the ghost projectiles that can be seen being shot from other players's weapons, as if in auto-fire.  Result: suppressed if the player is not visible.  Only Infantry weapons are affected, nothing else.

__`ChangeFireStateMessage_Stop`__
Following a targetted `CFSM_Start` packet reception, normally causes the ghost projectiles that can be seen being shot from other players's weapons to stop firing.  Result: suppressed if the player is not visible, but will also be dispatched if the other player, when last seen, was still reported as shooting.

__`ReloadMessage`__
Normally causes a player's weapon to perform the animation that indicates ammunition reloading.  This animation is actually just an illusion as another player's weapon never actually has an empty magazine - the shooting animation is just skipped and a no-fire sound effect is played.  Result: suppressed if the player is not visible.

__`ChangeAmmoMessage`__
Normally causes a player's weapon to switch between valid ammunition types by removing the old magazine and inserting a new one.  The packet itself, however, only displays the animation.  The rest of the process is performed by `ObjectDetachMessage` packets and `ObjectCreateMessage` packets.  As a consquence, the actual `CAM` packet is suppressed when the other player is not visible; but, the remainder of the `ODM` and `OCM` operations must be performed to maintain the integrity of state of the hidden player model.

__`WeaponDryFireMessage`__
Normally causes the sound effect that plays when a weapon does not discharge a projectile during (attempted) weapon discharge. Result: suppressed if the player is not visible.

__Conclusion__
"Ghost weapon discharge" is a phenomenon of the illusory nature of other players on an avatar's client.  The projectiles that these other players shoot from their weapons are insubstantial, unaimed, and unnumbered (in terms of global unique identifier) and are merely existant to give the impression that the weapon is being discharged at all.  The tracers do not offer help with dodging the real projectiles anyway.  Suppressing `CFSM_Start` and `CFSM_Stop` is significant because they preceed and follow every instance of so-called ghost discharge.  These packets are extremely common in any firefight.  `RM` has just a slightly reduced frequency but is also common in any firefight.  `CAM` is much less common than that and `WDFM` is even less common by comparison.

__Caveats___
The `ChangeFireModeMessage` packet can not benefit from the same improvements of visibility-based suppression as is implemented for the above packets and their handling code paths.  The only action this packet performs is necessary to maintain the integrity of state of the hidden player model.  It actually switches the fire mode, which often requires changing the ammunition slot being utilized by the weapon model, and can not be skipped.

___Addenda___
1. Only Infantry weapons operations are to be affected by the following changes, as aforementioned.  To ensure other forms of weapons's discharge is unaffected, a secondary pathway is available for any form of mounted weaponry.  This is available on the `VehicleEvents` event bus and is actually a duplication of the original message handling on the `AvatarEvents` event bus.  In reality, both Infantry weaponry and vehicle weaponry originally utilized the same messages, even though there was a distinction between players and vehicles for event bus otherwise.
2. Included is a fix for the Reaver causing an unreasonable amount of damage to Infantry during roadkill events.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -3,6 +3,7 @@ package net.psforever.actors.session.support
 
 import akka.actor.typed.scaladsl.adapter._
 import akka.actor.{ActorContext, typed}
+import net.psforever.packet.game.objectcreate.ConstructorData
 import net.psforever.services.Service
 
 import scala.collection.mutable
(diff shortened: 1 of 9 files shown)


ScrawnyRonnie:
Reaver collision damage is good.

Not sure what else to test as this looks like something only noticeable by looking at packets. That is if "suppressed if the player is not visible" means you are outside of the range to receive PSMs for that player?

Issue #1084, opened 2023-05-11
East spawn room door broken at Drakulu, Searhus


Dethdeath:
We came across this on the test server, but it's also present on our regular server. It won't open for anyone or anything.

ScrawnyRonnie:
This lobby door at Rehua/Searhus will not open on both servers too (from either side of the door)
![image](https://github.com/psforever/PSF-LoginServer/assets/73139382/19f9035f-0b05-49f5-b27c-87a389de43a4)

Fate-JH:
Doors that won't open are annoying to debug.

If it's a door with an IFF lock, being on the same horizontal as another door with an IFF lock can cause a major screwup with the associations.  Searhus seems to be awashed in annoying doors.  If that other door belongs to a tower, that means every door on at least one side might be wrong.

There's also the recent early proximity check for whether or not the door should open that might expose doors not in correct position in our files.

Dethdeath:
Yeah it does seem that all the doors with ~3700 as the first coordinate can be affected. I found some more broken doors, made a list of coordinates of all the ones found so far:
3770, 2688, 141 (east spawn room door at Drakulu)
3737, 2164, 53 (back spawn room door at Rehua)
3736, 2188, 60 (Rehua lobby door that Scrawny mentioned)
3791, 5568, 251 (Karihi air term door)
3770, 5540, 229 (Karihi tunnel door from spawns to BD area)
3735, 5501, 229 (Karihi door that goes from vbay to BD area)

Dethdeath:
StrikerZ mentioned the Gwydion air term door on Forseral in #bug-report.
There's another door at Gwydion that is bugged as well, but since there are no other broken doors on Forseral that I'm currently aware of it's probably not worth making a new issue for that continent. Anyway, these are the coordinates:
5528, 3839, 76 (Gwydion air term door)
5528, 3843, 56 (Gwydion bottom staircase door)

PR #1113, opened 2023-07-03
Fixed doors on Forseral - Gwydion

Resaec:
This PR fixes the 2 broken doors on Foreral in Gwydion

- stairway door to aircraft spawn room
- lower stairway, northern door

Fixes doors mentioned by @Dethdeath in #1084 

No wrongly associated doors on Forseral left.


Fate-JH:
I don't blame the test server for having some trouble.  I don't think the PSFCI knows what to do when GItHub asks it to wait for Maintainer approval.

Issue #1109, opened 2023-06-24
AMS & Router undeploys when owner leaves the continent


Dethdeath:
When the owner of an AMS or Router leaves the continent, the vehicle undeploys. To other players that are still on the continent it will continue to look as if it's deployed, even though it's not. When these players leave the continent and come back, they will witness the vehicle in an undeployed state. 

With the AMS this has consequences that are worse than the Router, because players can still select the AMS as a spawnpoint on the map, but are sent to Sanctuary if they attempt to spawn at it:

https://github.com/psforever/PSF-LoginServer/assets/77748540/fa90f1e9-4225-4cde-b3a0-d8e4983bb4d3

When the AMS owner leaves the continent, the other players can briefly see the AMS uncloak, but after that it looks cloaked again:

https://github.com/psforever/PSF-LoginServer/assets/77748540/98b79c6c-a03b-4c41-b7b8-b2507325c128

The Router appears to be fully functional, but the deployed pad doesn't work, because in reality this Router isn't even deployed anymore:

https://github.com/psforever/PSF-LoginServer/assets/77748540/88f2021f-c461-43b4-97cf-566daf1171a5

PR #1115, opened 2023-07-07
Fix AMS, ANT, Router and BFR siphon undeploying when owner leaves zone

Resaec:
Fixes issue #1109 where AMS, ANT, Router and BFR siphons undeploy when the owner leaves the zone.

AMS and Router were tested and are working. I'm not 100% sure that the BFR part works.
The AMS looses its cloaking for 1-2 seconds, but I could not find the cause. Maybe the owner change causes the client to disable cloak for a second.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -1443,14 +1443,10 @@ class ZoningOperations(
       continent.GUID(player.avatar.vehicle) match {
         case Some(vehicle: Vehicle) if vehicle.Actor != Default.Actor =>
 
-          // allow AMS, router, ANT and BFR siphon to remain deployed when owner leaves the zone
-          // TODO: do we need to check for left/right siphon?
+          // allow AMS, ANT and Router to remain deployed when owner leaves the zone
           vehicle.Definition match {
-            case GlobalDefinitions.ams | GlobalDefinitions.ant | GlobalDefinitions.router |
-                 GlobalDefinitions.aphelion_ntu_siphon | //GlobalDefinitions.aphelion_ntu_siphon_left | GlobalDefinitions.aphelion_ntu_siphon_right |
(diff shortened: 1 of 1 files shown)


src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
```diff
@@ -1442,8 +1442,22 @@ class ZoningOperations(
     if (player.avatar.vehicle.nonEmpty && player.VehicleSeated != player.avatar.vehicle) {
       continent.GUID(player.avatar.vehicle) match {
         case Some(vehicle: Vehicle) if vehicle.Actor != Default.Actor =>
-          sessionData.vehicles.TotalDriverVehicleControl(vehicle)
+
+          // allow AMS, router, ANT and BFR siphon to remain deployed when owner leaves the zone
+          // TODO: do we need to check for left/right siphon?
+          vehicle.Definition match {
+            case GlobalDefinitions.ams | GlobalDefinitions.ant | GlobalDefinitions.router |
+                 GlobalDefinitions.aphelion_ntu_siphon | //GlobalDefinitions.aphelion_ntu_siphon_left | GlobalDefinitions.aphelion_ntu_siphon_right |
```
Fate-JH, line 1450: `GlobalDefinitions.aphelion_ntu_siphon` is the arm weapon and should never resolve to the vehicle.

Perform `sessionData.shooting.findContainedWeapon` and then look for the ammo type `Ammo.ntu_siphon_ammo` on any output tool (`tool.AmmoType`).  It's the second set of values in the output pair.

Additionally, or alternately, please switch to `ConditionalDriverVehicleControl`.  I wish to avoid losing driver control of any vehicle if it is moved by the server (as it may) into a weird position.

Fate-JH (changes requested):

src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
Resaec: BFR siphon does not function like a deployment and is not continous. It requires the player to press the fire button and as such does not need to be added to this logic.
I removed them from the check.

Fate-JH (approved):

PR #1116, opened 2023-07-09
No Work

Fate-JH:
Normally, Middleware outbound packet operations will run at least twenty times a second in the hopes of finding packets to bundle and dispatch.  Each packet arriving to Middleware is also an instance when Middleware is called.  Each user connected to the server has an instance running.  The rate at which the self-calls occur can cause self-calls in the middle of an already executing self-call.  And stopping the service produces a run-off of three self-calls that get dropped because the service is stopping.  At 50ms per check, one user produces 20+ calls a second.  Ten users produce 200+ calls a second.  Twenty users produce 400+ calls.  Fifty users produce 1000+ calls.  This is a significant rate producing a significant amount of effort, much of which might be skippable as there may be no work to be done at all.

We can simplify this.  The primary mechanism is managed by a timer and a timer is just something something that runs after a certain amount of time has passed.  Additionally, we only need it to run when work exists and until work has been completed; and, the arrival of work can also be considered work existing.  To that end, when a packet arrives, a variety of things can happen.  The first packet is always dispatched immediately, but plans for no more work (unless it had to be split).  If the second packet arrives before the normal delay time, it is queued and the timer is started as an emergency.  The arriving packets will keep queueing until one arrives after the normal delay has passed.  When this happens, Middleware will immediately start bundling and sending a group of packets.  It might be a pre-made bundle or a newly wound bundle.  The time of dispatch is remembered and the emergency timer is restarted if there is more work to do (either another packet bundle or more unbundled packets).  If a packet does not arrive after the normal delay time ends but before the emergency timer has gone off, the emergency timer itself will trigger the next dispatch process.  Additionally, if there is no (more) work, there will be no planned attempt at resolving work that doesn't exist.

This cuts down on the effort Middleware has to expend.  When a player is alone, the packet rate should mostly be controlled by the packet rate and by the emergency timer equally.  As the number of players increases and the server has to dispatch the other players's packets for state management, the packet bundling and outbound rate should become much more closely aligned with that incoming packet rate and the emergency timer becomes merely an afterthought to the process.  Pointless work effort should remain low.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -220,7 +220,8 @@ game {
     # while additional entries with insufficient time spacing may result in no change in behavior.
     delays = [350, 600, 800]
   }
-  
+
+  # Don't ask.
   doors-can-be-opened-by-med-app-from-this-distance = 5.05
 }
 
@@ -234,7 +235,12 @@ anti-cheat {
 network {
   middleware {
     # How often between executions of the outbound bundling process
-    packet-bundling-delay = 50 milliseconds
+    # Used for outbound packet arrival triggers
+    packet-bundling-delay = 40 milliseconds
+
+    # How often between executions of the outbound bundling process
+    # Affects the base value on the timer
+    packet-bundling-delay-multiplier = 1.25
 
     # Pause inbound packet transmission towards the network if the sequence number is out of order
     # Packets are put aside until the sequence is restored, or this timeout passes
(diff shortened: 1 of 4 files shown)


jgillich:
>  At 50ms, one user produces 200+ calls a second

Actually it's 20 :smile:

Fate-JH:
Okay, okay, I'll fix the numbers. ...

PR #1117, opened 2023-07-10
Optimisations on MD5MAC and updated Scala version to 2.13.10

Resaec:
Adopted ListBuffer to ArrayBuffer switch that also fixes the problem with the new Scala version from #1097


jgillich:
Thanks!

PR #1118, opened 2023-07-10
Old Mac

Fate-JH:
"`MD5-MAC is a ancient MAC algorithm from the 90s that nobody uses anymore.`"
-- people who use MD5-MAC in 2023

After @Resaec gave us a profiling of the application with some losers just idling in sanctuary, propping open doors and holding up the buildings, a point of potential improvement was identified with the MD5-MAC algorithm.  Cryptographic information that is never changed from the start of the client session is retained and reused, rather than the whole container object being recreated to restore original cryptographic state.  The main trade-off is a certain amount of mathing, array modifying, and hashing during object creation, exchanged with equally repetitious `ArrayBuffer[Byte]` operations.

Sadly, those players will never improve.  They'll forever be losers.

### DIFF
--- a/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
+++ b/src/main/scala/net/psforever/actors/net/MiddlewareActor.scala
@@ -108,7 +108,10 @@ object MiddlewareActor {
   final case class Close() extends Command
 
   /** ... */
-  private case class ProcessQueue() extends Command
+  private case object ProcessQueue extends Command
+
+  /** ... */
+  private case object ProcessMissingSubslots extends Command
 
   /** Log inbound packets that are yet to be in proper order by sequence number */
   private case class InReorderEntry(packet: PlanetSidePacket, sequence: Int, time: Long)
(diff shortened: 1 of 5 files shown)


Fate-JH:
Please perform similar profiling again.  I might have not improved anything by doing this.

jgillich:
This isn't thread safe, and doFinal does at least partially reset the state already, it just wasn't complete IIRC

Fate-JH:
The crypto entity that governs the `Md5Mac` process is itself governed by the flow of `MiddlewareActor`'s outbound processes.  Each client, furthermore, has its own crypto entity.  It should never step over itself under normal operations.  It should also never step over itself during tests because, though there is some parallelization in that case, there is virtually never any cryptographic state to babysit.

I say that but, after looking at Middleware, there is a single point of compromise - in the timer set up by `askForMissingSubslots`.  The function `send` might be called whenever the timer executes rather than only during normal packet processing and that involves the crypto entity.  The function is synchonized to the request queue entity but that doesn't make it any safer.  It's an easy solution - callback by self message, or a dual synchronize.

PR #1123, opened 2023-07-31
Add support for launcher login via tokens and file verification

Resaec:
Added code to `LoginActor` to handle client authentication via login token.
Added code to LoginActor to generate the password used by the launcher to authenticate with the API.
Changed code in LoginActor to replace deprecated bcrypt functions.
Changed code in `Account` to add the field `password`, `token` and `tokenCreated`.
Added database `migration V009` containing table changes on `account`, new tables `launcher` and `filehash` and a trigger/function combo to update the tokenCreated column.


Fate-JH:
Is the replacement `bcryptBound` a suitable replacement for the deprecated `bcrypt`?  @jgillich didn't we hold back on updating something because it would pull in a `bcrypt` algorithm that wouldn't work in all the cases where we expect?

jgillich:
The only difference in `bycryptBounded` appears to be that it also performs a input length check. Should be fine, the algorithm is the same (scala-bcrypt is just a small wrapper around jBCrypt anyway) :shipit:

PR #1125, opened 2023-08-08
Return Relog

Fate-JH:
It was probably stale cryptographic information because the old session didn't stop correctly.  Without that, I do not see why it wouldn't work.


Fate-JH:
Okay so there are actually a lot of good reasons why it wouldn't work.

But I still do not see why it wouldn't work.

Resaec:
Tested working. Fixes the ability to re-log after leaving the world server and exessive packet spam.

PR #1127, opened 2023-10-07
PostgreSGL 10 Standards

Fate-JH:
An attempt to resolve the issues with the database being developed on a PG14 version but needing to be compatible with PG10.
I can barely read the standards guide.  People need to stop making web pages like this, or just copy the Java API.  Seriously suffering from so much information in between each section you can barely predict how the final form of the code comes together.

Also, why is it called "PG10"?  The "G" isn't even capitalized in the full name.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -134,7 +134,7 @@ lazy val QuietTest = config("quiet") extend Test
 
 lazy val docs = (project in file("docs"))
   .settings(psforeverSettings: _*)
-  .enablePlugins(ScalaUnidocPlugin)
+  //.enablePlugins(ScalaUnidocPlugin)
   .settings(
     name := "psforever"
   )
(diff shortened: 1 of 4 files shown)


Fate-JH:
By the way, it should be fine modifying this migration because the database stops before being able to implement it anyway.

Fate-JH:
Okay, so it hits the first trigger and goes kaput.  Fantastic.
*opens standard guide web page; browser panics*

Fate-JH:
Theory, it's the "OR REPLACE" part that is the issue.
Let's see how far I can get by removing that.

Fate-JH:
Okay, progress, the test server database liked `V008`, but now it's bumped into `V009`.  Let's try to fix that.

Fate-JH:
Also, unidoc will remain off.

PR #1135, opened 2023-10-23
Scrawny's Playground

ScrawnyRonnie:
Seeing if this fixes two equipment terminals in Supai. They're already broken so how much worse could I make it?


ScrawnyRonnie:
The terminal in the Redoubt spawn room is now working. Tested it with the empire that owns the base and hacking the term with another empire.

The terminal in the ground building works now, but it had to be hacked first. Will try something else with that one next.

ScrawnyRonnie:
Equipment terms in Redoubt bases are functional now. Terminals throughout the smaller buildings still need fixing.

PR #1136, opened 2023-10-23
Hanhau Doors

Resaec:
Fixed 7 door GUIDs on c2


lazer1982:
what cave is C2 i forgot?

Issue #1038, opened 2023-02-25
Lodestar - Lodestar doesn't repair damaged vehicles near it


Dethdeath:
The lodestar doesn't repair damaged vehicles near it even though it should. You can get as much ammo as you want from it, but no repairs.
Tested on Sanctuary and Solsar.

Fate-JH:
This is actually an issue I know or, at least, part of one.  At least one of the flanking repair terminals of the Lodestar doesn't seem to be positioned correctly on the model.  Or maybe it doesn't spawn correctly.  It's honestly been a long time since I diagnosed the issue and I forget if I ever found a golden approach where the "hiding" terminal can be engaged without the other one also triggering.  There's also not much I can do about it.

If neither terminal is activating - check both sides - then I'll run through the logic for proximity terminals.

ScrawnyRonnie:
Fixed in #1137 and can be closed.

PR #1137, opened 2023-10-28
Misc Cavern Fixes/Additions and Reduce Fall Damage

### LINKED ISSUE #1134: Infantry fall damage
Requesting reduced fall damage for infantry to more closely match the original. I've been trying to replicate falling seen in old video footage, but it now results in much higher fall damage than back then. I have a few video examples of dropping down off a gun tower from right near the turret. In the old clips no health was lost and 10 or less armor was. Current damage taken can be upwards of 30 health and 10 armor. It seems to be a complex algorithm because the results can be very wide. Turning and/or changing directional arrow while falling or at the last second seems to help reduce the damage taken, but this wasn't necessary back then. I could probably find more clips of falling to back up my claims, but everyone that has commented on fall damage has agreed that it is much higher than it was, just like rent.

Two possible ways to change the calculation from an ignorant player?
Adjust the velocity (fall speed) required to take higher amounts of damage

Keep the calculation the same, but then divide it by 3 or 4 and take that much damage instead

ScrawnyRonnie:
IA to caves works now
Redoubt spawnroom terms should all work
Reduced fall damage in a very elementary way, but it worked. So should close #1134


ScrawnyRonnie:
Logged in to 61005 to test these things and spot checked some of the terms.

ScrawnyRonnie:
Summarization of what this PR does:

Misc:
- Reduced fall damage in a very elementary way, but it worked. So should close https://github.com/psforever/PSF-LoginServer/issues/1134

Caves:
- IA to caves works
- Redoubt spawnroom terms should all work in all caves
- All [crystals](https://www.psforever.net/items/Crystal/) except the damage one are working.

ScrawnyRonnie:
I think I'm done this time. Since an issue was found with crystals I went ahead and added the other fixes I had ready to this one.

**Lodestar**
- Lodestar repairs vehicles now (addresses https://github.com/psforever/PSF-LoginServer/issues/1038). This was just a matter of tinkering. I don't know why making the radius larger worked. I tried a UseRadius of 30 and it didn't work. 40 did so I left it there.
**Last is first**
-  CommonRepairConditions should cease repairs to a vehicle that is near a repair silo, air pad, or vehicle crystal. I noticed that this would only delay repairs the first time the vehicle was repaired. It was made more clear to me when I increased the delay to 15 seconds. After that time elapsed, it would no longer interrupt the repairs. In the case of findLast, last = first occurrence for that vehicle. Changing it to 'find' fixed this and repair interrupts every time.
**More crystal stuff**
- Deth found some energy crystals that weren't working and we also noticed certain health crystals weren't. As far as I could tell by staring at the maps json files for too long, it seemed to be caused by how the local object GUIDs were being assigned. I just made it + 2000 so those would have no overlap with other objects on the maps and that seems to have fixed all of them.

Dethdeath:
Tested lodestar repairs too, looks good.

ScrawnyRonnie:
Recharge crystals and SOIs with equipment module benefit will only recharge ancient tech weapons now.

Dethdeath:
https://github.com/psforever/PSF-LoginServer/assets/77748540/ead1bb7d-14bd-43f9-aa5a-9a55bb2e0877

For some reason swapping weapons like this bypasses the check.

ScrawnyRonnie:
Nice 😂 That's just gonna have to ship. It's like it keeps track of how much ammo it added to a holster and if swapped the new item also gets that much ammo added.

src/main/scala/net/psforever/objects/equipment/EffectTarget.scala
```diff
@@ -65,7 +85,7 @@ object EffectTarget {
 
     private def CommonRepairConditions(v: Vehicle): Boolean = {
       v.Health > 0 && v.Health < v.MaxHealth &&
-        (v.History.findLast { entry => entry.isInstanceOf[DamagingActivity] } match {
+        (v.History.find { entry => entry.isInstanceOf[DamagingActivity] } match {
```
Fate-JH, line 88: Since history is organized EARLIEST -> LATEST, I'm concerned this is always only going to find the first damaging event, rather than comparing against the latest damaging event.

ScrawnyRonnie: It was just a "worth a shot" change that had the desired outcome. Each time I damaged the vehicle the repairing stopped. Before it would only delay repairs the first time it was damaged and if the repair attempt was started within 5 seconds of that. After 5 seconds it would keep repairing while taking damage.

Fate-JH: I see now.  I began switching around the order of events to be OLD + NEW instead of NEW + OLD, that is, normally appending actions to the history rather than prepending actions.  I didn't completely switch it around in all places before my latest major feature PR apparently.  In the update, however, I desire a OLD + NEW organization to history and exp evaluations were written with that in mind.

ScrawnyRonnie: I'll switch it back and know it'll work as expected again after yours is also eventually merged. I don't know the full effect of some of these things I change. I just try it, log in and see if I get what I was hoping for and hope it doesn't cause issues elsewhere 😂

Fate-JH (approved):

PR #1138, opened 2023-10-31
Blood Money

Fate-JH:
Experience is an Auraxium staple crop, the bread and butter of a healthy diet for your average PlanetSide troglodyte.  Experience points determine battle rank.  Experience points determine certification points.  Experience points determine wardrobe and cosmetics.  Experience points determine dopamine levels.  The blood and guts of your fellow soldier?  We've replaced that with experience points too!  Like a candy at a party, one must take a wack at the dangling dummies to get the sweet insides and feast.  Like a festival in the manner of childish monsters, and some monsters who should have stopped their festivities years ago while they were still young, one must venture to the enemy's strongholds and do the hard work of demanding from them their assets.  I've run out of metaphor.

Prior to this update, players would get 1000 battle experience points and 1000 command experience points per infantry kill and, typically, once per manned vehicle kill.  The real expereience system obviously didn't work that way in vanilla PlanetSide and this was makeshift until more responsible calculations were available.  We still don't have those, but we do have irresponsible ones!  The flat amounts were, of course, never correct.  The command experience deposit was obviously also not correct, but it was also even less correct because the context was wrong entirely.  Wrong mechanism - players should only receive that under certain conditions.  Other common incorrect behavior include being awarded for one kill per manned vehicle regardless of how many passengers as riding in the vehicle at the time of its destruction and only one person being awarded for a kill regardless of how many people tried to help.

Experience points of all varieties come in amounts related to contribution and effort rather than a single value and have been attached to qualifiable and quantifiable events.  The most common even is killed, hence deaths.  Along the base of the screen, are the fields labeled 'K' for "kills" and 'D' for "deaths".  The value that is visible as "kills" is associated with a sum of values found in the tables on the Character Info window, Show Stats Window sub-window.  For our purposes, the statistics reporting system is not fully implemented.  The rows and columns that correlate between empires and exo-suit types update and that updates the sum value for 'K'.  The first column on that sub-window, for a cell, indicates the long-term tabulation of kills, all the kills against that empire and exo-suit since the player joined the game.  The second column on that sub-window, for a cell, indicates this session's kills against that empire and exo-suit, and will reset whenever the client is closed.  It is this value that is used for the "kills" field in the main screen UI.  "Deaths" are tabulated whenever the player is killed, obviously. These are the only parts of the sub-window that will update for now.

This write-up is too damn long.  I'll separate it into the comments after this.


Fate-JH:
___EP___
How experience points has been earned has changed.  As indicated the original method was to offer 1000 of both battle experiecne points and command experience points per kill but the reality is that this systems is a lot more dynamic.  The value is not so fixed.  Following the changes for experience is actually difficult through our packet captures for various reasons, mainly for lack of transparency and disconnect between reason and result.  Members of the Discord, people who have much more knowledge of the game than I do, had nothing constructive to say on the matter.  The numbers and math are things I pulled out of the usual place.
__Battle Experience Points__
BEP is the most basic source of experience, acquired by applying one's firearm against a problem and succeeding.  Basic experience value earned by one player is calculated by determining a few conditions from the player who was killed.  The conditions, in order of increase in the base value, are: the length of time after being spawned, whether or not the opponent player has killed anything, whether or not the opponent player is mounted in a vehicle, and whether or not the opponent player has ever donned a mechanized assault exo-suit in their now-previous life.  Beneath the spawn time qualification, the value is one point and should not be affected by any further modifications.  After that, other modifiers include a battle rank disparity and a threat level assessment called menace.  Only the player who performed the formal killing gets full experience and everyone else who also used their firearm gets a fraction of that original value, modified by the number of shots discharged, and amount of damage inflicted or retained at the time of termination.  It is also possible to earn BEP through successful facility battles, but the value will be dependent on the individual experience rewards of the players who have participated in the contested facility (see below).
1) has killed an opponent player (fired bullet that leads to that opponent player's death, including the ten second grace period to ward against griefing suicide)
2) has assisted in killing an opponent player (fired any of the bullets that caused the opponent player's death, as long as it is not the last one)
3) be present at the end of a facility capture event and not be a squad leader or the player that hacked the facility capture terminal
4) be present at the end of a facility resecure event and not be a squad leader or the player that resecured the facility capture terminal
5) deposit nanite transfer units into an ally facility's resource silo using an advanced nanite transport vehicle

__Support Experience Points__
SEP is experience earned by one player having done something to benefit other ally players earning BEP but does not include dealing any damage to a mutual opponent player.  Any form of damage to a mutual opponent player is called an "assist" and anything that benefitted the ally player is called support or, internally, "contribution".  Contribution can be active.  Contribution can be passive.  In effect, there is no observable connection between the initiating activity (support) and the resulting reward (sep) because another ally player is performing the qualifying action (killing) independently.  On top of this disconnect, the reward does not happen at the exact moment of qualifying action (killing).  The resulting SEP pools with the player and then releases with an irregular schedule until all of the pool has been earned properly.

SEP is split into two periods - a short contribution period and a long contribution period.  These periods consider when the initiating activity is counted and depend not only on how long the player who performed the qualifying action has been alive but on the evaluated skill level of that player.  If the player has not been alive long enough, there is obviously no "long contribution" tally; same if that player's current skill demonstration is lackluster.  The short contribution period is always evaluated from the time of the qualifying action backwards.  The long contribution period goes further back than that.  Support actions of additional ally players and the additional support actions of ally players not found in the short period may accumulate in the long period.  The imperfect list of currently supported SEP qualifying actions that can be performed by "this player" append "ally player" with the clarification "who has killed an opponent player".
1) healed an ally player
2) revived an ally player
3) repaired an ally player
4) repaired the mountable entity used by an ally player
5) repaired the amenity used by an ally player to spawn or heal or repair themselves
6) repaired the amenity used by an ally player to acquire equipment
7) hacked the amenity used by an ally player to acquire equipment
8) repaired a vehicle that is mounted by an ally player
9) an ally player spawned at this player's advanced mobile spawn vehicle
10) an ally player acquired equipment from this player's advanced mobile spawn vehicle
11) an ally player acquired equipment from this player's lodestar vehicle
12) an ally player recently dismounted this player's vehicle
13) an ally player in their vehicle recently had that vehicle dismount as cargo from this player's vehicle
14) an ally player recently transported using this player's deployed router system

__Command Experience Points__
CEP is a form of experience for personnel management and large-scale planning - command, as the letter 'c' means.  It accumulates differently from BEP or SEP and provides a completely different set of rewards that improve the experience of personnel management and large-scale planning.  The primary expression of these skills is squad leading but some independent actions are included.  CEP rewards its own ranking system and general access to lines of communication and use of unique equipment.  CEP is calculated as the accumulation of BEP earned by all players who were involved in the facility capture event, awarded to specific players in the sphere of influence of the facility when the major capture event gets resolved, factoring in population peaks and popultion shifts, reported region activity trends, and, where necessary, hack duration.  One only gains CEP during these facility capture events.  Without any combat during the capture event, there is no CEP or BEP to be earned.

Due to squads being disabled, fewer players than intended will be awarded and that amount will be scaled to the lowest rank of squad management.  Only players interacting with the facility capture terminal and/or the facility's lattice logic unit earn CEP, but normally all ally squad leaders in the sphere of influence of a contested facility should also earn some CEP.  Not all of the abilities and the equipment are implemented in the game proper.  As a consequence, the main reason to earn CEP at the moment is a cosmetic reason and for clout.
1) hack the facility capture terminal of an opponent facility and be present when the capture completes
2) resecure the facility capture terminal of a friendly facility after it has been hacked
3) deposit the lattice logic unit of a major facility into the capture console of a capture target allied facility
4) kill any opponent transporting (holding) the lattice logic unit of a major facility capture
Any player who does not earn CEP during the resolution of a major capture event is rewarded equivalent BEP that has not scaled by squad constraints.  Earning CEP through interaction with the LLU in any mechanic that is not perfectly vanilla is mainly to offset the lack of formal squad operations in the game.  When determining squad participation, the members of the current squad must have also been fighting within the facility sphere of influence during the capture event.

Fate-JH:
___Play versus Progress___ **IMPORTANT**
PlanetSide Classic is properly played as a game of character development.  Its roots are that of an MMORPG with firearms.  As a player, you start the game without skill or access and everything you do while playing teaches you skills and earns you accesses.  The more the player does - exploring, killing, helping, capturing - they earn experience and through that experience they gain combat effectiveness.  That is classically how the experience pans out.  During development of PlanetSide Forever, however, players have been gluttonous and have demanded more and more and soon gained all access that was possible at the time.  Initially, this is seen as a method of testing various things through voluntary offerance and, for that reason, was permissible.  This is not without its controversy.  Some people feel they can not play the game without a specific something.  Others have cited this as spoiling of the proper PlanetSide experience and have demanded a return to clawing through the world, tooth and nail, approach.  Partial accommodations have been suggested including replacement of the default exo-suit, the default rifle, and even granting certain vehicles without cost or penalty.  I myself aim for the shangri-la that is vanilla - only what was in the game prior to shutdown, no change more than is necessary by our limitations, only offering the most basic of creature comforts to offset the degeneracy of the playstyle.  That is neither here nor there.

While thinking upon the reserves program SOE utilized once upon a time to fill out recruitment goals with inadequate unshaven vagabond, I arrived at this opportunistic revelation.

Upon start of a new(?) player at battle rank one, that player is given the option of advancing immediately up to a higher battle rank (see the configuration file) with all of the benefits of the said battle rank.  The player may actually select any battle rank up to and including the limit.  If the offer is accepted, however, that character has all further battle rank advancement suspended until catching back up to their current battle rank in experience.  The player must earn the experience equivalent to the advanced battle rank they have selected before they may earn any more experience to advance to the battle ranks beyond that, and they may only do that through SEP or through BEP earned during facility capture events.  Additionally, CEP is set back to zero and no new CEP may be earned until they player gets back their freedom of advancement.  No CEP is ever maintained.  This is a debt that must be paid but one is allowed to abandon this artifical advancement and return to battle rank two with CEP-earning privileges restored as well.  No BEP is retained upon reset.

The chat command for battle rank skipping is `!progress` with a number to indicate the battle rank.

An event chat message will indicate when BEP or SEP is earned but the exact amount will not be stated.

An event chat message will indicate when CEP is earned but the listed amount will always be zero.

The chat command for undoing the battle rank skip is `!progress 2`.

To be very clear about this, the player is to become a grunt and only a mildly comfortable grunt.  That player can purchase more equipment and may use implants and may even pull a variety of vehicles beyond what a typical new player may, but never a lot.  It is recommended they serve a role.  They will kill only to live longer.  They will live longer only to give aid their allies and glory to their empire.

Fate-JH:
___Features___
__`HackCaptureActor`__
When a facility capture event ends, either through a completed capture or a resecure of the capture, instruct the facility to award points to players who are involved on the victor's empire.

__`ParticipationLogic`, `MajorFacilityHackParticipation`, `TowerHackParticipation`, `NoParticipation`__
Collects information from players within a sphere of influence by piggybacking the normal sphere of influence checks to gather physical player presence each second, for up to fifteen minutes.  At the end of a facility hack situation, the collected information is utilized to analyze contribution for the prior facility capture event and distribute experience points to the victor empire depending on the results.  This contribution data involves population changes, the length of time a player has been in a sphere of influence over the previous capture event duration or in the sphere of influence in general (whichever is applicable), and a series of hot spot information across the duration (where applicable).  Only major surface facilities and towers should be accommodated right now.  In the future, cavern facilities will require special considerations because of their enhanced verticality.

__`ExperienceCalculator`, `SupportExperienceCalculator`__
Zone-level actors that are the entry points for performing the math for rewarding experience for kills.  The experience is separated into kill and assist activity and into all manner of support activity because the distinction is most efficient.  The complexity of the list analysis necessary to manipulate the data for producing the experience values, as well as need to step away from the problem to collect additional information when switching over to support activity, makes the pacing mechanism more convenient.  It takes a significant amount of work to compute and we should avoid tasks being suspended behind ongoing math problems, even if in the grand scheme it does not take that long at all.  Messaging overhead should also be considered.

__`KillAssists`__
The calculation of experience from a player being killed, covering both the primary killer and anyone who added another tick of damage into the effort.  After the person who performed the kill shot earns their experience, every player who also caused damage to the victim also receives an amount of experience based on that value and the effort they undertook to deal damage to the target.

__`KillContributions`, `RecoveryExperienceContribution`, ...__
The calculation of experience for a player helping one player kill another player, covering a variety of rewards given to accomplice players for having done things for the benefit of the killer.  It feels weird explaining it like that.  Common contributions include healing and repairing but more complex ones include having driven the player between facilities.  The different values that go into the calculation of this experience can be found in the configuration file.  Contribution is collected from other entities at the time of the event, the interaction with that entity.  Contribution is usually expressed in terms of heal and repair history but features a number of additioal situations that add contribution.
(There's a chance this feature is going to cause the game's memory footprint to explode.  A different method of shorthand calculations will become necessary if that is the case.)

__`AvatarActor`__
Improved battle experience assignment and command experience assignment and implemented periodic support experience rewards.  Accommodation for the battle rank progression system's interactions with experience.  Composes updates from kills and deaths.

__`ScoreCard`, `Life`, `EquipmentStat`, `KDAStat`, `Statistic`__
A virtual record system that was introduced into the server codebase in an earlier update but not properly utilized until now.  Score cards keep track of ongoing weapon statistics, a list of player kills, and a list of a player's previous lives and revivals during this gameplay session.

__`AvatarStatisticsMessage`, `type/Statistics`__
Packet and structure data for displaying statistics on the Character Info window, Show Stats Window sub-window.

__`ZoningOperations`__
Takes avatar updates from kills and deaths and turns them into statistics records in a character's score card and displays that data in the UI along the base of the screen and on the Character Info window stats subwindow (these are only kills).  The message for the play versus progress system is displayed upon spawning under specific conditions.

__`WithShields`__
Entities that have shields are now connected by their own parent class.

__`CaptureTerminalDefinition`__
This feature isn't new but facility capture terminals will maintain facility capture times.  This shouldn't be a problem because terminals and the buildings they are installed in are connected at the hip.  This installation determination also sets the capture participation logic for that sort of facility.

Fate-JH:
___Tables___
The tables have two primariy purposes - record events that can be used to populate stat fields, mainly for benefit of the client, and keep track of experience gains per player per event.  Some basic housekeeping is applicable both for raw table data but also functions and triggers that manipulate the tables.  Even though the rows do tabulate experience per event, the experience values on the `avatar` table supercede any value that can be produced by adding up table rows.
__`sessionnumber`__
The concept of a "session" was originally a distinction between different spurts of gameplay intended to compartmentalize statistics but the idea never materialized properly.  It can still be used to that ends but the time rules were changed.  Different "sessions" will currently queue if a user logs into the game after an hour since the last log in and the current session will stay relevant as long as the user remains logged in.  I don't know what you would call this now.  The table for `kdasession` and `weaponstatesession` use this concept of session directly, whereas all other tables will use timestamp information.

__`kda`, `kdasession`__
A summation of KDA values, including all revives, for a session.  When rows are deleted from `kdasession`, they automatically get appended into a row stored in `kda`.  When this happens, it no longer can be divided into "sessions" and will represent unified KDA data.

__`weaponstat`, `weaponstatsession`__
Information about weapon use, superfluous information which PlanetSide Classic didn't normally keep track.  We don't currently have use for it either, but we're keeping track of it because these count as traditional FPS stats.  When rows are deleted from `weaponstatsession`, they automatically get appended into a row stored in `weaponstat`.  When this happens, it no longer can be divided into "sessions" and will represent unified weaponstat data.

__`killactivity`, `assistactivity`__
All kills are itemized and sufficiently explained here.  Can and will be used for a variety of things, but mainly builds campaign kills and deaths in the client statistics window right now.

__`supportactivity`, `ntuactivity`__
Sufficient information to express the events surrounding support activity.  Support events may be translated back into a statistical categories (most of the time).

__`buildingcapture`, `expbuildingcapture`, `llubuildingcapture`__
Sufficient information to express the events surrounding building capture.  Includes separate tables for CEP rewards, BEP rewards, and LLU rewards.  When rows are deleted from `buildingcapture`, they automatically summate into a value stored in either `expbuildingcapture` or `llubuildingcapture`.  When this happens, data no longer can be divided into individual facilities and will represent either of two groups of unified building capture data.

__`progressdebt`__
Keeps track of experience required until a player is no longer in debt by choosing expedited battle rank advancement.  All players who have logged in once will have an entry in this table and most will probably be zero'd (probably).

Fate-JH:
___Caveats___
1)  Although a lot of things have been discovered about the statistics tracking interface found under the Character Info Window Display Other Statistics sub-window, some fields do not populate just yet.  Additionally, since the system has been cherrypicked to make use for the K/D UI, not much thought has been put into how to make it work properly before kills by empire and by exo-suit.
2)  In the case of kill assists, only players who still have active damage against a target are rewarded experience for the engagement (maybe?).  In the case of heal support or repair support, all players who contributed will receive experience, with health recovery that still benefits the killer or health recovery that was lost during the engagement rewarded better than health recovery that was lost before the engagement.  The actual scope of when someone should get rewarded for support activity and when it should no longer count is another unclarified aspect of the experience reward system, but this reasoning emphasizes the maximum number of players getting rewarded for helping others but personl effort still be rewarded over tokenism.  "You did help the other person" over "you tried".
3)  Some SEP comes with requirements or limitations that are as arbitrary as they are difficult to explain.  For example, the possibility that a player can be rewarded more for doing something more often but will reach a plateau where their earning increases vanish.  Likewise, they can have earning increases until the plateau, and then an oversatuation of the activity causes it to stop qualifying altogether for a minute or so.  Trying to plan for limitation and against exploitation is difficult.  This may result in some unexpected, perhaps unsatisfying, behavior when the SEP does not trigger and there is no explanation.  It my be equally difficult to tell that the SEP is not triggering but should trigger.
4)  For example, when using a carrier vehicle to transport another vehicle as cargo, after the cargo vehicle has dismounted at the destination and destroyed something, the driver of the carrier should receive SEP.  The action of mounting and dismounting the carrier is not enough to qualify for this SEP alone.  The cargo vehicle must be transported by the carrier a certain distance or must be transported by the carrier for a certain amount of time.  Before that, any successful engagement is just normal BEP to the cargo vehicle driver.
5)  For example, in being awarded SEP for repairing or hacking an equipment terminal that an ally player uses, purchases something, and then goes on to kill an opponent player, the ally who performs the kill does not have to use the equipment that was acquired from the terminal.  That player does not need to have the equipment as weaponry, loaded in other weaponry, or even in his inventory at all.  Anything can be purchased from the terminal.  The actual equipment acquired during the "Buy" action is never recorded - only the terminal interaction is remembered.
6)  CEP through facility capture is mostly untested.  The calculation involves a non-trivial number of people being active in the sphere of influence of the facility capture over the course of the duration of the capture event.  The only formal exercises that have been performed involve whether or not resolution during a near null-state causes any server crashes.  In other words, **_please test_**.
7)  Due to a concern that the facility capture system will excessively request information from the hot spot system, it has been disabled for the time being.
8)  For the play versus progress system, the maximum battle rank selected was 13.  This offers the promoted player a maximum of two implant slots and a maximum of seventeen certification points with which to play.  Proper utilization of these available points is important, not only for comfort, but also for survival, as a promoted player will be dealing with their resulting loadout options for a significant amount of playtime.  Not having to commit to the maximum battle rank may give less certification opportunities but also a quicker resolution of the debt that restricts organic promotion again.  By being limited to ally contribution and facility capture, anyone using the progress system should want to prioritize a role they can serve.
9)  Upon utilizing the progress aspect of the play versus progress system, one's experience UI will be inflated to the value of the promotion.  If that player logs out, it will be reset to zero upon the next log-in.  This is mostly correct behavior.  If the player reverses promotion due to the play versus progress system, the value will be subtracted from your current BEP and, if it is zero, it will plunge into negative numbers.  The client needs to be restarted to reset the UI.
10)  Much to everyone's disappointment but to no one's surprise, there are no formal tests for this update.

Fate-JH:
___Addenda___
1)  All instances of nanosecond time (`nanoTime`) were replaced with millisecond time (`currentTimeMillis`).  Additional effort is required by the server to produce precise nanoseconds and extra precision is unnecessary.
2)  I'm amazed the database tables work finally.  For anyone having not followed the saga: I started working on the tables and triggered functions for this update, then I separated that SQL into its own pull request such that it would not interfere with any minor corrections until the main update was completed, we discovered that the PostgreSQL 10 will not play nicely with PostgreSQL 14 syntax and that caused the server to enter a state of "no updates but not gonna admit not working", finally the tables were fixed in series of "so what will PG10 allow for", a second fix for the SQL was applied, and this pull request will be a third fix applied plus the completion of the system using those tables.  I am still not certain if I am satisfied with their current state but that may only be because much of the data that they collect goes unused.
3)  Entity ownership has changed to resolve an issue with posession by allowing separation between owner GUID modification and owner name modification.  This affects both deployables and vehicles.  See `OwnableByPlayer`.
4)  Every building in the game does not to keep track of every capitol facility on every surface continent on an individual level.
5)  The play versus progress system utilizes the `MailMessage` packet to display its initial login message.  (The "initial login message" is actually displayed as long as one logs in while the conditions qualify.)
6)  Corrected a logging issue when a non-CTF base's `CaptureTerminal` amenity correctly failed to locate a CTF component in the installation facility  - a `CaptureFlag` or, more specifically, the `LLUSocket` - but then incorrectly raised complaint that the facility could not find the `CaptureTerminal` it started with.
7)  Fixed some issue with vehicle cargo'ing.  I really don't remember, sorry.
8)  The chat command to check how much experience remains to be earned while under the progress system is `!progress x`, for any battle rank other than 2.  You may also type `-h` or `-help`, but that resolves to the same thing internally.

ScrawnyRonnie:
A few initial observations:
- The priority message is sent every time you spawn/zone
image
- The session K/D HUD resets kills to 0 if you die. Deaths do not reset after respawning.
- Two NC characters I created were both broken on login attempt and could never load into the world (client stuck on black screen). 
What I did: I created a TR character and logged in, then an NC (same account) and this happened. I created another NC and it did the same thing. I then made a VS character that worked, then another NC and that one worked too. I tried to replicate it on another account, but every character logged in fine. Just something to look out for or it was a fluke?

Dethdeath:
Can't seem to fill NTU silos with ANTs, was tested at 3 bases. The filling animation works fine though, but the NTU level for the base never goes up.

Fate-JH:
There's a minimum duration the LLU needs to be held before CEP will be granted upon kill.  It's 1 minute, according to the files changed.  I had meant to change that into 30 seconds, and will do so later.
It may still not be working correctly.  I didn't work directly on the LLU drop mechanics so I am curious how much I can stretch it.

ScrawnyRonnie:
Mentioned in my previous comment, but it happened again. I created another account. First character created and logged in fine. Second character had this happen again after selecting them on the character list to log in.
image

Fate-JH:
I thought I had this set for the NC only.

Fate-JH:
Fixes related to the above issues.
- NTU silo should work now (problem wasn't that it wasn't working but that it was working at an incredibly slow pace that rounded up to zero)
- Allies and own self no longer reward exp when killed (sorry, NC)
- K of K/D works correctly between kills and between lives

Additionally, I paced the avatar login instructions thus that only the field necessary for the object create packet are loaded before the rest of the server continues the login process.  Information related to other packets sent during avatar login are allowed to load asynchronously in hopes of allowing the server to load the player quicker, before it attempts to boot them from the world.  This seems to work better but the issue still arises with reduced frequency.  Will have to make this an official examination into the matter later.

Dethdeath:
Tested filling NTU silos and XP being given for killing friendlies, both are fixed.

Fate-JH:
Changes to address issues demonstrated on Discord:
- kill experience being 0 (potential fix; no reproduction steps currently exist)
- neutral facility captures should reward experience as long as there was some combat between the victor faction and some other faction

ScrawnyRonnie:
Comment on previously mentioned thing:
The black screen 'You have left the world' or just indefinite black screen still happens, but I think only sometimes when logging in to another character on the same account. I started making a new account for each character and didn't experience this. You are still able to log into multiple characters on one account, but you'll run into the black screen at least 1/2 the time after the first char.

Base capture oddity:
When a hack goes through it shows that the defending empire resecured the base and the attacking empire captured it. I captured an LLU to see if the same resecure message would appear, but it did not. I hacked the hack & hold base again and saw the resecure message again.
image

ScrawnyRonnie:
Ah, that chat message about resecure is not exclusive to this PR. We don't hack bases very often in fights so it seemed "new". Will make a separate issue for that.

Dethdeath:
Can confirm the "You have left the world" issue when logging in with multiple characters on the same account.
Filling up a base with an ANT gives no XP.

src/main/resources/application.conf
```diff
@@ -223,6 +217,192 @@ game {
...
+        {
+          name = "revival"
+          base = 0
+          shots-multiplier = 15.0
+          shots-cutoff = 10
+        }
+      ]
+    }
+
+    # Support experience points
```
Dethdeath, line 355: Should be Command experience points

Dethdeath:
Getting this when I type !list:

invzone

Fate-JH:
Upon this update, the following changes to characters will occur:
1) current number of unvalidated kills will be retained on the side (cep / 100)
2) battle rank set to 1 (bep = 0)
3) command rank set to 0 (cep = 0)
4) all implants forgotten
5) implant shortcuts will be forgotten
6) certifications set to basic
Unvalidated kills are those kills that are unpaired with deaths and thus can not be measured in terms of the K/D system going forward.  There's nothing really to do with them but I feel people will complain if I just threw them away altogether with the BR reset.
Battle rank being set to 1 will ensure equal footing with the new experience gain rate.  In addition, the play or progression system provides enough chance to acquire the specific weapons and equipment a player may think he needs to be good.
Command rank being set to 0 will ensure equal footing with the new experience gain rate.
Implants will be lost with the reduction of battle rank anyway.
Certifications will be lost with the reduction of battle rank anyway.  In addition, the system going forward will involve proper character development.

I do plan to rework the concept of "free certifications" into the game afterwards.

theonetrueduddy:
> **Support Experience Points** The imperfect list of currently supported SEP qualifying actions that can be performed by "this player" append "ally player" with the clarification "who has killed an opponent player".
> 
>     1. healed an ally player
> 
>     2. revived an ally player
> 
>     3. repaired an ally player
> 
>     4. repaired the mountable entity used by an ally player
> 
>     5. repaired the amenity used by an ally player to spawn or heal or repair themselves
> 
>     6. repaired the amenity used by an ally player to acquire equipment
> 
>     7. hacked the amenity used by an ally player to acquire equipment
> 
>     8. repaired a vehicle that is mounted by an ally player
> 
>     9. an ally player spawned at this player's advanced mobile spawn vehicle
> 
>     10. an ally player acquired equipment from this player's advanced mobile spawn vehicle
> 
>     11. an ally player acquired equipment from this player's lodestar vehicle
> 
>     12. an ally player recently dismounted this player's vehicle
> 
>     13. an ally player in their vehicle recently had that vehicle dismount as cargo from this player's vehicle
> 
>     14. an ally player recently transported using this player's deployed router system

Aware that this is an imperfect list so asking for clarities sake; shouldn't "an ally player in this player's vehicle killed an opponent player" also be on this list?

Not had chance to test this in an actual play session yet, but a quick test resulted in the driver of a vehicle where an ally killed and opponent player receiving no SEP.

NoDriverSEP
NoDriverSEP2

Reproduction steps;
1. Had all the test subjects travel some distance so they were alive for a while
2. Had "SEPtestVStarget" kill multibox0 to gain some value as a kill, 51 BEP awarded
3. Had "multibox0" respawn, hop back into the prowler and then kill "SEPtestVStarget" (who was in a vehicle), 102 BEP awarded, 0 SEP awarded to driver "SEPtestTR"
4. Had "SEPtestVStarget" come back, repeat the kill on "multibox0" (102 BEP)
5. Had "multibox0" respawn, kill "SEPtestVStarget" again (this time on foot). 102 BEP, 0 SEP awarded to driver "SEPtestTR"

Issue #812, opened 2021-05-09
If a player bails from their Vehicle, or is killed from the Vehicle Pad spawning their Vehicle, /warp will no longer work until getting into a Vehicle


NickPSF:
If a GM is killed from the Vehicle Pad spawning their Vehicle, /warp will no longer work until getting into a Vehicle

ScrawnyRonnie:
No longer occurs; good to close.

PR #1139, opened 2023-11-12
Address Vehicle Spawn Bug and Neutral AT Terminals

### LINKED ISSUE #1133: Dismounting a vehicle as it spawns can make you invisible
When you pull an aircraft, before it flies off the pad you can press 'G' and exit the vehicle. This causes your client to freeze up and look something like this:
image
You can't type or click anything. If you Alt F4 and then log back in before persistence ends it will put you back at the terminal. Other players in the zone are then unable to see you (bottom left client):
image
A player that enters the zone after you are in this state will be able to see you. After you die it corrects this and everyone can see you again.

This only happens with aircrafts. With certain ground vehicles you can Alt G to bail as it starts driving away which causes the pad to become blocked, but this doesn't freeze your client or cause any other issues.

Is there something that could be put in place to prevent you from exiting/bailing from vehicles as they are spawned?

ScrawnyRonnie:
**Vehicle Spawn Bug**
_The change in VehicleControl.scala_
Made an attempt to resolve #1133 and from my testing it seemed to be successful. A vehicle spawn activity, as far as I know, is either when a vehicle is created from a vehicle spawn pad or when a vehicle enters a new zone. This change prevents dismounting (pressing "G") within the first few seconds of one of these events. Why 8.5 seconds? That was roughly how long it takes for a ground vehicle to spawn, auto drive, and give control back to the player at certain ground pads. As mentioned in the comment in the linked issue, it was possible for the client to lock up at that point if 'G' was pressed. I don't fully understand the vehicle spawn process so this may not be the best way to address the issue, but for now it is good enough maybe? The only negative I found was if a player's zone loading happens pretty quickly, there may be a few seconds where they can't get out of their vehicle, but I also don't foresee players wanting to jump out of their vehicle right away after zoning.

**Cavern neutral terminals**
_All other files changed_
I found that simply by allowing terminal access and exchanges to neutral owned terminals that the terminals throughout the caverns are accessible by all factions without hacking first. It sounds like this would also change the behavior of terminals in neutral bases and towers in the overworld (or even neutral bases in the caves themselves), but those behave normally and show the message that they are not owned by your faction and still require hacking first. I lazily went through the map json files and set the owner ID of all of them to the vanu_core in each cavern instead of the actual building they are in. Didn't seem to matter as long as they had a building owner that wasn't an actual base and I don't foresee any issues with this approach.


Dethdeath:
I bailed out of an air vehicle as it was spawning and it broke /warp for me. Pulling another vehicle after this and not bailing this time fixed the problem. I tried to reproduce this on the regular server too and nothing broke there.

ScrawnyRonnie:
I did notice that I was still able to Alt + G "bail" out of an aircraft, but no longer just 'G' exit before it lifts off the ground. I didn't know that it broke command(s) from working though. I was still visible to other characters so it's at least better and is a different issue than the client freeze bug. You can't bail from ground vehicles anymore during that initial first few seconds and block the spawn pad.

ScrawnyRonnie:
Looking at the logs while bailing, the dismount is being prevented, but I guess aircrafts have a different behavior because it shows that I'm kicking myself out. This may be the cause of the issues with commands as you shouldn't be able to exit the vehicle during this time, but it is still able to happen via a different means.
image

ScrawnyRonnie:
Made futile attempts to prevent aircraft Alt + G bailing during the liftoff auto pilot ascent, but I still could with each dismount change I made. I'll have to concede that this isn't a perfect "fix", but it's better than a client freeze and invisible players.

Dethdeath:
Maybe 8.5 seconds is too long for air vehicles too? If the pad is being camped by AA the pilot might want to bail out as soon as possible.

ScrawnyRonnie:
If bailing was possible before having full control of the aircraft on live (or would make for a good QoL change), then we'll consider it a positive that this change doesn't affect the ability to bail right away and we shouldn't try to prevent it with aircrafts 😀 It was the exiting the aircraft before it left the pad that caused major issues, so that being prevented now is good.

Dethdeath:
I checked some neutral cavern terms and they all worked.

If #1133 can occur just as the auto drive ends too, then 8.5 seconds is suitable for ground vehicles.
Ideally we could have a separate timer for air vehicles, as their auto pilot ends after like 7 seconds. However, since you can always bail from those anyway it won't matter in practice.
It does mean that #812 is back again.
But obviously having players go invisible is still worse than that.

ScrawnyRonnie:
Split air vehicles into their own dismount condition for 3 seconds. It's enough time that you can't 'G' and exit the vehicle before it leaves the pad. From my testing I had to be spamming Alt G to bail just after getting airborne. So the issue is with commands not working still there, but much less chance of it happening at this reduced time.

Dethdeath:
Nice! Should be very rare for commands to break in that manner now.

PR #1141, opened 2023-11-16
A few more fix attempts

### LINKED ISSUE #440: Wall Turret - Turret Upgrading - Upgrading a turret as someone gets in gets you stuck upgrading with no option but /suicide
Upgrading a turret as someone gets in gets you stuck upgrading with no option but /suicide.

Edit: The turret did already have an upgrade, might be relevant

### LINKED ISSUE #1126: Cargo Vehicle Occupants DC
Requirements to DC:
-Your empire owns an Amp Station on the continent
-You are within a friendly SOI that has a **direct link** to the Amp Station base

Steps:
1) Enter a Gal or Lode as a cargo vehicle and you and any passengers/gunners will DC upon entering it.

If you are not within the SOI of a friendly base that is linked to your Amp Station, you will not DC. As soon as the aircraft enters an SOI that is directly linked, the cargo occupants will DC as seen in the image. The Gal slowly approached the SOI and BAM, bye bye cargo.
image

ScrawnyRonnie:
**Cargo vehicles and shield charging**
Fixes #1126 . If a cargo vehicle attempts to charge shield, it crashes the client of all cargo occupants. Cargo vehicles are now excluded from shield charge attempts.

**Turret upgrading**
Fixes #440 . This was a doozy to come up with a janky solution for. The isUpgrading flag was not being set when the upgrade started so other players were allowed to mount the turret. This caused the upgrading player to get stuck. The flag is now set to true in the correct place when the upgrade starts. Everything worked fine as long as the upgrade finished. But what if the upgrading player gets shot, killed, or decides they don't want to finish the upgrade? That would leave the turret in a state with isUpgrading true and nobody could mount it unless someone went back and finished the upgrade. As with other hacking logic (upgrading a turret uses some of the same things as hacking objects), there was no means to cancel the upgrade or set that bool back to false as far as I could tell. 

_Insert solution I came up with_. A variable is set to keep track of the system time during the upgrade process. If a player stops in the middle of the upgrade, the time stops updating. If someone then attempts to mount the turret at least 1.5 seconds after the upgrade progress is stopped, it will allow them to mount it despite the isUpgrading flag still being set to true. Once mounted, it immediately sets that to false and everyone can go on their merry way mounting the turret even though it never finished upgrading.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionData.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionData.scala
@@ -963,13 +963,26 @@ class SessionData(
     log.debug(s"$pkt")
   }
 
+  // This is for excluding cargo vehicles from attempting to charge their shield while mounted in a carrier vehicle.
+  // As per issue 1126, this would cause the client of the cargo vehicle occupant(s) to disconnect
+  def handleGetCargoVehicles(vehicle: Vehicle): List[Vehicle] = {
+    val mountedVehicle = vehicle.MountedIn.flatMap { mountedGUID =>
+      continent.GUID(mountedGUID.guid).collect { case mountedVehicle: Vehicle => mountedVehicle }
+    }.toList
+    mountedVehicle
+  }
+
+  def handleIsCargoVehicle(vehicle: Vehicle): Boolean = {
+    handleGetCargoVehicles(vehicle).nonEmpty
+  }
+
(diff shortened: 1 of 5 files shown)


Dethdeath:
No crashes with cargo vehicles, seems fixed.

Dethdeath:
Turret upgrade issue is also fixed.

PR #1142, opened 2023-11-24
Item Swapping

### LINKED ISSUE #1108: Shooting with no weapon out
This bug has been around for years and happens every so often during infantry combat. It is caused by a player equipping a weapon that has just been placed into the weapon slot.
The result is that you see a player shooting at others with the weapon still holstered on their back.

It seems to be some sort of timing issue where 2 actions happen, the client performs the action of pulling out a weapon and the server sends a holster action to other players instead.
From this point forward the actions are reversed for the observer.

Here's some videos that display the problem:

https://github.com/psforever/PSF-LoginServer/assets/77748540/d3091237-60ba-437b-aa66-86428d7b3f10

It can happen with both weapons and tools:

https://github.com/psforever/PSF-LoginServer/assets/77748540/209c1589-b017-45c0-b3d1-9e039f4f02c7

The reverse also happens, where someone can be seen surging with a weapon out, while on their screen it is holstered:

https://github.com/psforever/PSF-LoginServer/assets/77748540/21aaf390-0c37-44dd-b92a-99259f79fead

If you have a weapon out before dropping the next weapon into the slot, it seems to happen much more frequently.

Reproduction steps:
Put a weapon/tool into a slot and pull that item via your slot hotkey at the same time.

ScrawnyRonnie:
Crafted this to fix #1108 
Right click swapping an item into a slot already behaved correctly. Left clicking an item and freely moving it to a slot would cause an issue if that slot was currently equipped (being held in the player's hands). This should fix that issue so the player and others both see the item holstered or unholstered correctly. This applies to both moving items around in your inventory as well as looting a corpse inventory and moving an item from that one into one of your slots. No more shooting while it appears there is no weapon drawn (I think).

Also threw in the quick fix for one of the spectator comments to not allow a spectator pull a vehicle. By accessing the terminal, then enabling spectator allowed the player to pull a vehicle, which spectators shouldn't be able to do. If someone were to mount a vehicle, then enable spectator mode, this already didn't seem to cause any issues. When a player destroys a vehicle with a spectator inside, they are rewarded with the kill and not kicked.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionTerminalHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionTerminalHandlers.scala
@@ -102,7 +102,7 @@ class SessionTerminalHandlers(
         lastTerminalOrderFulfillment = true
 
       case Terminal.BuyVehicle(vehicle, _, _)
-        if tplayer.avatar.purchaseCooldown(vehicle.Definition).nonEmpty =>
+        if tplayer.avatar.purchaseCooldown(vehicle.Definition).nonEmpty || tplayer.spectator =>
         sendResponse(ItemTransactionResultMessage(msg.terminal_guid, TransactionType.Buy, success = false))
         lastTerminalOrderFulfillment = true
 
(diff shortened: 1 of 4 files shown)


ScrawnyRonnie:
Added a fix for LLU bases when the hacking empire only has a warpgate/geowarp link to the base, thus making the hack a hack & hold. Any existing LLU will be removed. If the base was already hacked by another empire that had a linked base on the zone, it was not removing the LLU when the hack switched to hack & hold as seen here:
image

Issue #1106, opened 2023-06-03
Client crash in vehicle leads to corrupt session with potentially invulnerable player


Dethdeath:
Crashing in a vehicle often leads to a familiar black screen with accompanied "expansions_missing_avatar" messages:
![blackscreen](https://github.com/psforever/PSF-LoginServer/assets/77748540/bbf79827-18b5-446c-a85d-b6a4caac533c)

Eventually the player is presented with a loading screen and it places them back to where they were before the crash.
From what I understand this process is basically patchwork and might be changed in the future?

Either way, coming back from a crash can lead to a corrupt persistence state. Using medkits (or stamina kits) in this state will tell you to "Please wait ..." after the first medkit is applied.

Another thing that is affected is the way second wind works, this one is quite confusing as it's supposed to heal the player for 25 health and reset, but what happens instead is this:

https://github.com/psforever/PSF-LoginServer/assets/77748540/1b162ee2-6168-4bb5-bd0a-8fa5d44ca8d9

https://github.com/psforever/PSF-LoginServer/assets/77748540/c80703df-6ac7-4300-9a36-552341d63456

At first I thought the black screen was required to reproduce this bugged state, but it's not. Even when the player is correctly placed back into the vehicle without seeing a black screen, they can be in a bugged state.

I have tested it with various vehicles and the most reliable way for reproduction is to get into a gunner seat of a vehicle, I recommend liberator/vulture, prowler, juggernaut and skyguard, but others can work fine too. The driver can also get bugged, but in my testing it occurred less often.

Reproduction steps:
Pull a vehicle.
Kill the Planetside client with task manager.
Relog.
If you get a black screen, wait for it to go away.
Exit vehicle if still in one.
Get damaged by something and attempt to heal to full with medkits.
If using second wind wait for it to initialize (if you come back after a crash and it shows as initialized, you still have to wait the full charge up period before it will activate).

ScrawnyRonnie:
I found that this also happens when a character is not in a vehicle. It seems like as long as you log back in before persistence ends you'll end up in this state until you die or zone somewhere else. Here is one of the logs that may help:

[...]

PR #1146, opened 2023-11-29
Keep the jam going & other misc fixes

### LINKED ISSUE #542: Implants - Spawning always resets implant initialization
It should only reset when the player died. It should not reset when the player "redeployed" via deconstruct, IA, recall.

### LINKED ISSUE #813: Vehicles - Players randomly get stuck in Vehicles and Wall Turrets
Players randomly get stuck in Vehicles and Wall Turrets

### LINKED ISSUE #1017: Terminal condition glitched at NC base, Ikanam
Time of: *Monday, 5:49 PM CET*
In-game Location: *Ikanam / Amerish*
Factions: *Character is TR, base is NC*
<hr>

Upon REK engaging the vehicle terminal, the timer does not appear to start and the message is simply "Access Terminal." 

<a href="https://cdn.discordapp.com/attachments/295670273544290304/1069765184907911188/vehtermbugged.mp4">6 second video clip</a>

### LINKED ISSUE #1085: Vehicle wrecks appear on the minimap and have full health

### LINKED ISSUE #1140: Can't get this game to work at all keep getting unknown connection error
image
Any help would be appreciated, but so far nothing works; setting launcher and game to run as admin had no effect. I get either "Unknown Connection Error" or "Unknown Login Error".

### LINKED ISSUE #1143: LLU location lost when player holding it disappears
When a player holding the LLU's session times out because of a crash or DC, the LLU is moved to 0 0 0 on the map.
I've looked at what happens with 2 LLUs at the same time and it appears they both go to 0 0 0, but because of a collision or some other reason, one or both of these LLUs will be physically gone even though it still shows them as being there on the map.

DudewheresmyLLU

Reproduction steps:
Hack an LLU base and grab the LLU.
Close the client with ALT + F4, or kill it with task manager.
Wait until the player's session times out.

### LINKED ISSUE #1145: Jam duration doesn't get extended
If you jam a vehicle, any additional EMP blasts are ignored by the jammed vehicle and they can only be rejammed after the initial jam ends.
This allows vehicles to get shots off inbetween jams, even though they should've been unable to.

### LINKED ISSUE #1150: Geowarps can crash players
Geowarps can make players crash. It doesn't always happen, but when it does it can be quite bad:

https://github.com/psforever/PSF-LoginServer/assets/77748540/5159cbbe-1e33-474e-aa7e-32ef7501b646

The respawn button doesn't work either.

### LINKED ISSUE #1153: LLU invisible to some after vehicle death
When a vehicle blows up with an LLU carrier inside it, only the former carrier can still see it. Other players that were near the death of the LLU carrier can not see the LLU anymore, that goes for the driver too.
Players that get onto the continent after the death of the LLU carrier can see it.

ScrawnyRonnie:
Fix #1145, resolves #1085, resolves #1150, resolves #1143, resolves #1153, resolves #542
Vehicle weapons and turrets should stay jammed if something jams them again.

Close other Issues that can be:
Fix #1017 
Fix #1140 
Fix #813

### DIFF
--- a/src/main/resources/zonemaps/map07.json
+++ b/src/main/resources/zonemaps/map07.json
@@ -8926,7 +8926,7 @@
     "AbsY": 3742.29175,
     "AbsZ": 42.3043633,
     "Yaw": 198.0,
-    "GUID": 713,
+    "GUID": 715,
     "MapID": null,
     "IsChildObject": true
   },
@@ -8939,7 +8939,7 @@
     "AbsY": 3729.643,
     "AbsZ": 42.3043633,
     "Yaw": 288.0,
-    "GUID": 714,
+    "GUID": 713,
     "MapID": null,
     "IsChildObject": true
   },
@@ -8952,7 +8952,7 @@
     "AbsY": 3729.643,
     "AbsZ": 49.8043633,
     "Yaw": 108.0,
-    "GUID": 715,
+    "GUID": 714,
     "MapID": null,
     "IsChildObject": true
   },
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -10,7 +10,7 @@ import net.psforever.actors.zone.ZoneActor
 import net.psforever.objects.avatar.scoring.{Assist, Death, EquipmentStat, KDAStat, Kill, Life, ScoreCard, SupportActivity}
 import net.psforever.objects.serverobject.affinity.FactionAffinity
 import net.psforever.objects.sourcing.VehicleSource
-import net.psforever.objects.vital.InGameHistory
+import net.psforever.objects.vital.{InGameHistory, ReconstructionActivity}
 import net.psforever.objects.vehicles.MountedWeapons
 import net.psforever.types.{ChatMessageType, StatisticalCategory, StatisticalElement}
 import org.joda.time.{LocalDateTime, Seconds}
@@ -235,7 +235,7 @@ object AvatarActor {
 
   final case class SetStamina(stamina: Int) extends Command
 
-  private case class SetImplantInitialized(implantType: ImplantType) extends Command
+  final case class SetImplantInitialized(implantType: ImplantType) extends Command
 
   final case class MemberListRequest(action: MemberAction.Value, name: String) extends Command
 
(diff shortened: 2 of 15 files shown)


ScrawnyRonnie:
While testing different ways to jam something I noticed the big bus EMP capacitor was recharging very quickly. Wiki says 20 minutes to full recharge, but this simply made it 10x longer than it was, so it takes 5 minutes now.

Dethdeath:
Looks good.

ScrawnyRonnie:
Caverns seem to have some kind of poisonous gas that selectively either allows players to explore them or will end their life upon entry. This isn't an ideal gaming experience, so a means to prevent random death upon zoning to a cavern has been introduced.

Some caverns seemed to be safer for entry (c1-3), but c4-6 had a higher chance at ending your life and causing a sad state for the client. This is to address and hopefully Fix #1150

ScrawnyRonnie:
Fixed a few doors in Freyr on Esamir
Added a fix for #1143

Dethdeath:
Freyr doors and #1143 are fixed. Came across #1153 during testing.

ScrawnyRonnie:
cya #1153?
image

ScrawnyRonnie:
This only fixes the issue if the LLU carrier dies after the driver in the kill feed. When the carrier shows up first, the LLU remains invisible.

ScrawnyRonnie:
After more testing with LLU in a vehicle we found that sometimes this fix works and sometimes the LLU still becomes invisible. Investigating...

ScrawnyRonnie:
Given the complexity of #1153 I'll just leave this as is. No additions coming to this PR.

ScrawnyRonnie:
A few more additions:
The black screen in #1106 has been addressed in the change here in AvatarRejoin:
src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
Turns out unmounting and mounting the vehicle isn't required and from my testing you stay in the driver seat now and can continue to operate the vehicle. The other things in that issue are still present.

I noticed CancelZoningProcess was spamming constantly if a player was moving or firing a weapon. I added conditions for each so it only does this when a zoning request is actually being made (IA, recall, quit, etc). Will it reduce load in any capacity? idk

Also, when a zoning request is made it will now use DeactivateActiveImplants() instead of Deinitialize. This is in case the player moves or does something to cancel zoning so they can continue to use their implants. Deinitialize would completely disable them so they could not be used until zoning or dying.

ScrawnyRonnie:
Had another go at #1153 thanks to an idea from @Resaec and it looks like it worked.

Dethdeath:
Black screen seems fixed, so does #1153

ScrawnyRonnie:
#542 is now part of this PR. Implants should now only deinitialize and fully restart the initialization timer after dying. Deconstructing and zoning via warpgate, IA, or recall should keep initialized implants in that state. Since each implant has a certain amount of time to "charge", a calculated time to initialize is set based on the time you last spawned (into the world or after dying). For example:
Enhanced Targeting has a 60 second initialization timer.
You died and respawned. The 60 second timer starts.
20 seconds later you deconstruct.
15 seconds later you spawn again after deconstructing.
Enhanced Targeting will initialize in 25 seconds based on the calculated timer that was set.

The character window where it shows the progress of initialization timers will reset each time if the implant was still under a timer and not initialized. It could be 10 seconds from initializing, but look like it just started if you just deconned. There doesn't appear to be a way to fix the visual progress bar there or make it start where it should be based on how much time is left, so it can go from looking like it barely started going to initialized white bar.

I've only come across one glitch so far with this change. Every now and then Surge would initialize even though it was nowhere near 90 seconds since its timer started. If I deconned or zoned again before that amount of time elapsed, sometimes it would stay initialized, other times it would not be and then initialize after the timer finished. Not sure why it would do this and it didn't happen with the other ones I tested. I guess the implant _surged_ to initialization? 😉

Dethdeath:
Zoning does weird stuff to the timers. I see other implants like darklight also activate right after respawning. Somehow dying is being ignored.

ScrawnyRonnie:
I think the timer issue is fixed now.

Dethdeath:
I think so too.

PR #1149, opened 2023-12-05
Log-Related Fixes (2023-12-5)

Fate-JH:
Most of these were related to option failures - `None.get` - and list traversal failures - index mismanagement and _`foo`_`.head` - and query failures - wrong persistence entity and autoincrement assignment.  The damage recovery calculation, on close inspection, was mostly wrong from start to finish and required retooling.  I feel this is more correct.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -231,7 +231,7 @@ game {
     # Battle experience points
     # BEP is to be calculated in relation to how valuable a kill is worth.
     bep = {
-      # After all calculations are complete, multiple the result by this value
+      # After all calculations are complete, multiply the result by this value
       rate = 1.0
       # These numbers are to determine the starting value for a particular kill
       base = {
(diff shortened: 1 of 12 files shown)


Fate-JH:
I've taken a twenty line function and turned it into a mess of functions that span half of the file it was in and repurposes list data a couple of times to perform calculations in the hopes of inflating battle experience points into something that convinces people who vaguely remember what battle experience points felt like that it's what battle experience points felt like.  I went from something that was easily calculable and was perfectly predictable to an absolute mess.  This is all of your guy's fault.   I will not take any responsibility for this.

So, this is the code I wrote:
The original conditions for battle experience points - was ever a MAX, killed at least one things, mounted in a vehicle, etc. - those still exist but their conditions have been expanded a bit to ensure they more correctly satisfy a wiki write-up about what was going on under the hood.  They were ported into a category called "short life" modifiers because they would be the most prominent values for a player who died too quickly to have done anything significant.  The majority of the calculations have gone into something called "long life" modifiers.  Everything a dying player has ever driven in or has ever worn is assessed and accounted for second per second of their previous life; the most prominent thing a player has done within that assessment produces a maximum value; and, finally, the life span of a player, numerically counted in factored seconds, becomes part of the experience reward for that given player's death.

The battle rank disparity heavily favors punching over one's weight class (battle rank) and the penalty for punching down still exists but has been reduced tremendously.  The number of kills the dying target has achieved is added to the experience reward, modified by any ongoing menace.  There's also a concept called "revenge" which is about one player killing the player who just killed them.  Moment to moment population balance modification is still excluded from consideration.

I have not tested this code yet, whether it will ever crash, or what kind of bep values it will output.

Fate-JH:
By the way, the new numbers require just as much tuning as the old ones so don't ask why they look like they do in their current form.

Fate-JH:
This is the form ready for testing.
The test server does not seem to be restarting.
Welp.

PR #1156, opened 2023-12-15
Decode Packets, Again

Fate-JH:
- fixed critical issue where iterator failed to work correctly because the underlying data was a stream and the stream was closed once the iterator was produced
- even then, switched from the iterator of an open stream to an outputted list of static Strings because the iterator wasn't well-formed and could fail to extract data (results in a brief moment where the overhead of both the file stream and a copy of the file's data exists)
- added a command line option to include a whole input directory of the type of file indicated - normal `.gcap` data or `-p` for preprocessed ascii; all of these files will be processed
- demoted the `<files>...` command line argument into another option, though still unbound; both it and the input directory can contribute but the decode types must match; all of these files will be processed
- working with the files is a more straightforward process, only needing to evaluate the file type from params once

All of the specifications for the tool as described in the ReadMe should be be relevant.
Command line example: `psforever-decode-packets -p -o "decoded" -i "extracted -f "file1.txt" "file2.txt"`

Addenda: I haven't tested on files produced directly from Gcapy - the natural alternate mode of the tool - because my PATH variable isn't set up to accommodate it and I have not installed Python on this computer to run the application anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ make install PREFIX=$HOME/.local
 ```
 Now you can run the program like that:
 ```
-psf-decode-packets -o ./output-directory foo.gcap bar.gcap
+psf-decode-packets -o ./output-directory -f foo.gcap bar.gcap
 ```
 By default, decodePackets takes in `.gcap` files, but it can also take gcapy ascii files with the
 `-p` option. Run `psf-decode-packets --help` to get usage info.
--- a/src/main/scala/net/psforever/packet/game/objectcreate/ObjectCreateBase.scala
+++ b/src/main/scala/net/psforever/packet/game/objectcreate/ObjectCreateBase.scala
@@ -98,8 +98,9 @@ object ObjectCreateBase {
       }
     } catch {
       case ex: Exception =>
-        log.error(s"Decoding error - ${ex.getClass.toString} - ${ex.toString} ($objectClass)")
-        Attempt.failure(Err(ex.getMessage))
+        val msg = s"Decoding error - ${ex.toString} ($objectClass)"
+        log.error(msg)
+        Attempt.failure(Err(msg))
     }
   }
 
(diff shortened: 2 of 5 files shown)


Fate-JH:
`--skip-existing` is a param that would not overwrite output files that already existed, with the potential use of deleting only certain files with a particular file decode failure, or for a transcoder pattern that has changed, and only getting updated decodings for those files.  The different ways in which files and directories were included in the params and could be discovered otherwise was causing issues with identifying complementary packets from the input source in the output directory.

An easy example of the failures happen when attempting to work with files in code that were defined with the navigation hints `.` (same directory) and `..` (parent directory) and absolute file paths.

I wrote code to allow for error message copying to an additional log file but those logs do not include the errors I wished that they would.  Once again, `log4j` has a particular interaction with `stdout` that does not get properly captured through what seems like obvious methods.

Fate-JH:
I forced it.

Fate-JH:
I decided to dive somewhat deeply into the operation of log4j to determine why calls to logging bypass interuption to the output stream and still get printed to the console.  For reference, `Console.withOut` and Console.withErr` should block all relevant code that uses `println`.  In fact, it should not matter what, at all, is involved since the stream should no longer point to the normal command line output buffer.

At first I took a foray into `StdOutLoger` because I think I'm working with a C language but quickly reminded myself that `ConsoleLogger` is how log4j works.  As usual, I'm a bit of an idiot.  After tracing the code down, I discover the root is ... Java.  log4j is primarily a Java library used by a Scala application.  `ConsoleAppender` doesn't call `println` (`Console.println`).  It uses `System.out.print`.  Both calls utilize the same mechanics, operating as the face of a `PrintStream.write` call, but they interact with the underlying console differently.  Using a `Console.withOut` block with `Console.println` will catch the `println`; but, with a `System.out.println` call, the input is sent to the original stream.  They act like completely different buffers going to the same location - the terminal.

I turn off my Scala mindsdet and go back to Java.  I use `System.setOut`.
The sample I've run dozens of times gets run again.  No extra logging to terminal.

My fist hits the desk. "That's so stupid!"

PR #1159, opened 2023-12-24
Fully implement the VoiceHostRequest packet

Resaec:
This PR implements the VoiceHostRequest packet.

With these changes the client input of the relevant packet can be decoded and encoded. Full packet data coverage.


Fate-JH:
Now that I think about it, that app pscs.exe is probably pronounced as "Pisces".

Issue #1154, opened 2023-12-10
Kills not counting


ScrawnyRonnie:
Some kills are not rewarding xp or adding to the K/D HUD or stats window. For the player that dies it also doesn't count their death in the K/D HUD. It's as if the kill never happened, except for the part where it shows in the feed and someone dies. So far no rhyme or reason has been discovered, but it happens frequently.

lazer1982:
To add on to this tested it with a fell Alts of mine Base Cap Rescues don't grant XP even with 9 kills
![PSScreenShot0049](https://github.com/psforever/PSF-LoginServer/assets/99909749/b1414c99-827c-4972-8670-1fa10d17f68d)

PR #1163, opened 2024-01-16
More Experience Fixes

Fate-JH:
1.  Fixed issue caused by calling `tail` on an empty list, which doesn't work
2.  Revenge kill experience rate down (0.15 -> 0.05) and value is capped

### DIFF
--- a/server/src/main/resources/db/migration/V012__ScoringPatch3.sql
+++ b/server/src/main/resources/db/migration/V012__ScoringPatch3.sql
@@ -1,0 +1,3 @@
+/* Original: V008__Scoring.sql */
+ALTER TABLE assistactivity
+RENAME COLUMN attacker_id TO killer_id;
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -332,11 +332,13 @@ game {
         # The kill event must have been the exact previous life's death after revive or respawn.
         # This only applies if experience is set to 0.
         # Set to zero and experience = 0 to ignore revenge.
-        rate = 0.15
+        rate = 0.05
         # If player A kills another player B who killed player A just previously, deposit this experience.
         # The kill event must have been the exact previous life's death after revive or respawn.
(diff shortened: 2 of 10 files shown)


ScrawnyRonnie:
Could you explain what needs to happen to the killer, killed, or another player that was involved in assisting one of the former that would have caused that so it can be effectively tested?

Fate-JH:
`head` and `tail` are complementary operations on a list of elements, the former taking only the first entry, and the latter taking all of the other entries.  Calling `head` on an empty list is an issue.  Calling `tail` on a list with at most one element is an issue.  The correction is pretty generic.  A specific test case could possibly be established but setup would be complicated.

Such a condition most likely involves repairing enough damage suffered by the victim player in such a way that exactly one entry of a previous weapon's damage remains unresolved.  In practical terms, perhaps, being healed as close to 100 health as possible without actually getting to 100 health, just enough that one more tick of the next heal amount would be necessary?  It would also depend on the weapon and the healing strength employed.  The explanation sounds too commonplace.

ScrawnyRonnie:
Tried a variety of heal/repair/medkit amounts and couldn't not get the kill to count. Based on how we played on Sunday, much of it was respawn, engage. At most many times a med kit would have been used before dying and a majority of kills weren't credited so this seems good. I've yet to meet the conditions to get it to not count a kill (even for the previous issue) since this update was introduced so don't take my word alone.

Fate-JH:
I'm linking issue #1154 because obstruction of the experience system would supposedly lead to failure of the kill count system as well.  This one has to work for the other one to work, so I'm hoping that resolution here fixed problems there.

Fate-JH:
3.  Fixed assist database logging by making column name match expected local entity variable names (required by Quill)

**As this is a database change on top of an existing PR, the test server may have issues starting correctly.  I will just close the PR once satisfied if that is the case.  Similarly, pulling in this PR to test it locally will modify your own database in a way that it would require manual revision if you wish to test any other branch before merging, even master.**

PR #1164, opened 2024-01-16
Restored Purchasing Power

Fate-JH:
Fixed issue where a failed vehicle order causes all future vehicle orders to be blocked until a life changing event (death or being spirited away).


ScrawnyRonnie:
Tested and working 👍

Dethdeath:
Can also confirm.

PR #1166, opened 2024-01-28
Turret Automation

Fate-JH:
```
I detect you, you see me
While you break through enemy lines
And deploy your spawn machine
This bd I'm left to guard
What my servo may overlook
The sensors will direct to my gun
If your actions catch my ire
That tree will not shield

Automatic fire
Will shoot at you
Respawning is all you can do
```

The most difficult of the combat engineering deployables to implement have always been the ones that function without need for a user's input.  The spitfire turret and its cousins - the concealed shadow turret and the ouranophobic cerebus turret - are constructed with the same humble equipment that Auraxian's greatest warriors have had access for the longest time - the adaptive construction engine (ACE) - but present the most astounding of behaviors.  Many of the ACE's many manifestations remained ineffective or mostly ineffective for some time even as the ACE's construction secrets were laid bare.  At first, the more explosive aspects of the ACE were given life through a combination of mechanism and self-incrimination.  The hand-triggered explosive Boomer was simple to execute.  The heavy explosive mine (HE mine), a proximity mine by any name, functions by the player's avatar being unwavering in its brutal honesty to the server about his place in the game world and is rewarded by death.  The dual sensors of the ACE - Motion Alarm Sensor and Dirsuptor - remain ineffective, though they were rendered in proper animation.

For the longest time, the most daunting of the ACE's constructs were the aforementiomed turrets that possessed a great deal of latitude in their discrimination and engagement with their prey.  While simple automation had already been accomplished by other amenities and support workflows, the turrets acting on their own were beyond that.  Through analysis and inspiration of previously applied principles, however, the turrets have been properly conquered to an acceptable extent.  By extension, the automation of unattended major facility turrets are also possible.  The turret will turn and face its opponent when angered.  It will unleash a volley of projectiles upon recognition of a nemesis.  It will also linger on the hope of an escaped target's return to discovery when necessary, or will rapidly switch to the next opportunity, or redirect its aggression towards something that would actively pester it by damaging it.  Essential to a weapon turret's own activity turned out to be inspiration from the proximity explosive mines - the revealing self-incrimination of its would-be targets.

The new contract of the PlanetSide Forever age - that the server will assert authority over the client - has to be compromised.  At the moment, I can not imagine any other way for turrets to work properly than the way they have been rendered here.  Turrets face the same concerns that explosives have in the past, that they can affect targets that otherwise they would have no realistic interaction with were they left their own unbridled devices, but any functionality is preferrable to a complete absence of utility.

They do not make good statues.

___How It Works___
Automated turrets can be thought to have five primary modes or states.

__Idle__
In this neutral state, the turret is unaware of any potential targets and does not react to anything because it has nothing to react to.

__Aware__
The turret becomes aware of at least one potential target.  The target is not yet close enough to shoot at but it can potentially move closer or move farther away.  In this neutral state, the turret will review this selection of potential targets periodically to determine if they remain potential or if they have moved far enough away to no longer be considered possible targets.

__Prowling__
At least one target has moved close enough to the turret that it could begin shooting at it.  To determine if the target is reachable in terms of the game world environment, each potential target periodically has a single projectile fired in its direction.  The turret will continue to shoot at all potential targets.  The turret will aim at each target and remain aimed at the target while it shoots every so often.  Anytime a valid target gets hit, it will confirm the shot with the turret, and the first target that the turret recognizes having confirmed its test shot will become the selected target.  This is a basic confirmation call and response.

A detail that is important to mention for this stage of target enagement is that each target only perceives the turret aiming at itself and firing test shots at itself.  Any entity not drawing test fire does not see the turret engaging in test fire with any target.  Different potential targets do not see each other getting shot at either.  Turret weapon fire does not synchronize across all perspectives for the zone (region) until the first potential targets confirms that it has been shot with the turret and is declared the selected target.  This test shot does not cause any damage when it lands because multiple potential targets can respond to it.  In this manner, the turret is capable of searching for targets without overtly exposing any player.

__Attacking__
The turret has established one potential target entity that is within its weapon fire range and is reachable despite whatever game environment features are between the target and itself.  This is the selected target.  The turret will begin shooting exclusively at this selected target and all other test shots are cancelled.  It will continue to shoot at the selected target until the target is dead, no longer a valid target, has moved too far away, or has not been reachable by the turret's line of fire for a certain amount of time.

__Self-Reporting__
If the target has not been reachable by the turret's line of fire for a certain amount of time, the turret will return to its prowling activity to determine the next available target.  Before that, however, it will fire a single special type of test shot at its former target.  If this test shot has a response message then a second form of target automation will initiate and the target and the turret will begin closely communicating with each other, damaging the target in a predictable pattern.  Normal prowling mode will not resume until this target resolves itself, including moving a bit.  The primary purpose for this self-reporting mode is to avoid a loss of valid targets due to simple shenanigans on the part of the target or its occupants.  Currently only valid for vehicles.

Self-reporting is an optional mode.  Normally, when "attacking" completes, the turret goes back to any of the three early modes.

___Features___
__`InteractWithTurrets`__
Hitherto mentioned self-incrimination.  Each entity to which turret automation can respond is to declare to the game world environment "I exist here" and all nearby turrets are made aware of this existence and do with the knowledge as they will.  For the moment, and potentially for the forseeable future, only player entities and vehicle entities must do this.

__`TurretDefinition`/`AutoMation`__
Information regarding a turret's automatic fire mode, mostly documented.

__`AutomatedTurret`__
A parental trait connecting all weapon turrets that could possess autonomous weapon fire behavior.  Connects `TurretDeployable` entities and `FacilityTurret` entities.

__`AIDamage`__
The packet that is dispatched by the client when an automated turret is instructed to shoot at the player avatar and hits either the player avatar or interacts with something else in between.  We only care about hitting the player avatar or the entity in which the player avatar is mounted.  This packet is only received by the client whose player avatar is being hit.

__`WeaponAndProjectileOperations`__
Receiving and processing the packets for the turret's detected damage is performed herein after being redirected from `SessionActor`.  The workflow mimicks the process of target assessment and projectile selection as is similar to other method of handling damage from "hits" or "splash" or "lashing", though the projectile must be generated on the fly rather than allocated from a previous packet, and there is less need for error detection.  Damage handling does not need to deviate from the standard manner as other entities.

__`EffectTarget`__
A number of conditions were crafted to allow for the turrets to perform target validation - this is something they would shoot at - and target blanking - this is something they would not hoot at and should no longer shoot at.  These functional evaluations are not necessarily good break downs of the way the turrets regard their potential targets but they operate sufficiently to provide a convincing fake agency.

__`MountableTurretControl`__
A control agency base type that is utilized for major over-land facility turret control agencies and subterranean point-defense turret control agencies as well as turret deployable control agencies.  Should primarily focus on mounting behavior, but includes `Damageable` overrides due to common implementation inheritance.

__`FacilityTurretControl`__
Modified to respect the new inheritance structure with `MountableTurretControl` but differentiate it from `VanuSentryControl`.  As a feature shared with `TurretDeployableControl`, inflicting the turret with a jammed status should cause it to silently stare at the target that had caused it to become jammered.

__`VanuSentryControl`__
A new control agency exclusive for the subterranean point-defense turrets.  Isolates messaging for Vanu sentry ammunition recharging and adjusts the process such that it only begins recharging when the turret is no longer being fired, as opposed to stopping and starting every time a projectile is created.

__`AutomatedTurretBehavior`__
Primary one stop shop for logic pertaining to a weapon turret's autonomous weapon fire mode.

__`AffectedByAutomaticTurretFire`__
Response and handling for the turret's self-reporting target check weapon fire mode.  Currently only valid for vehicles.

___Characteristics___
While turrets have functionality, this functionality is only possible due to a combination of compromising behavior and some uncompromising behavior.  Some personality has permitted.
__`spitfire_turret`, Spitfire__
1. target escape range and test firing range are the same, so the turret is easier to escape

__`spitfire_cloaked`, Shadow__
1. does not retaliate against a different attacker if it is already shooting at a selected target
2. no target switch cooldowns

__`spitfire_aa`, Cerebus__
1. only searches for aircraft targets
2. target escape range is twice the distance of the test firing range, making it harder to escape
3. does not retaliate against a different attacker if it is already shooting at a selected target
4. no self-reporting weapon fire behavior
5. cylindrical target search area as opposed to the normally spherical one

__`manned_turret`__
1. target escape range is twice the distance of the test firing range, making it harder to escape
2. retaliation delay is longer than other turrets, making it harder to distract from the current selected target
3. cylindrical target search area as opposed to the normally spherical one

___Caveats___
1.  The method of target detection we are forced to work with is stupidly conspicuous and there is no reprieve from the .  The turret goes out of its way to turn towards a target and fire at it once every few milliseconds (one second by default), flying past the intended target with a wiff sound, or striking something that can't be damaged like a tree of a rock with a crunch, and does it again in a few moments if the conditions are correct for a follow-up.  The shots are fairly audible.  There's no way to hide the tracer.  The turret, dumb to its environment, can not realize it should not shoot at a target because it's blocked.  Shadow turrets suffer for this the most as each shot, whether in test or in proper, must be done with the turret frame uncloaked; and, restoration of the cloak is not quick enough to avoid being seen.
2.  The turret should prioritize targets that are closer to it.  The primary means of determining this is by dispatching test shot packets by target proximity.  In truth, however, the order in which packets are dispatched is not enough to ensure that closer targets report being struck before targets further away.  The speed of the network connection between the client and the server is often more important.  Additionally, rapidly switching between targets by prioritizing closer targets that respond within a certain time of each other causes animation whiplash with the turret weapon facing direction.  Leaving it alone produces a satisfactory appearance.
3.  I seem to recall that auto turrets used to have a target switching ability where it switched to whatever was the closest target.  This is a problematic ability, if so, because I can not replicate it easily right now.  In fact, it is entirely imcompatible with this  test shot method of target selection. 
4.  When a facility turret stops shooting at a target, it will reorient back into a rest position where the weapon is aligned to the approach stairs of the turret.  When a deployable turret stops shooting at a target, it will be stuck facing whatever direction it was previously looking until directed to look elsewhere at another target.  It is not possible to manually aim the turret weapon in a particular direction without rebuilding the whole turret body to face in that direction.  It doesn't work like other weapon turrets where `ChildObject StateMessage` might be able to manipulate it.
5.  If one player is not looking in the direction of a turret when it starts firing, it might not be turned in the direction of its proper target.  This is merely a visual bug.  If it must pan to track the target, it might update properly.
6.  When an auto turret stops shooting, the final shot might be seen flying off in the direction of the original placement of the turret rather than the direction of its weapon.
7.  The method of weapons fire from auto turrets called "self reported" was introduced for a specific reason - without it, a player can stop an auto turret from taregting his vehicle by dismounting it; and then, as long as the player does not also draw the turret's ire, a vehicle and driver are now parked safely in place just nearby an active turret without being attacked.  Initially, the situation sounds like something just worth writing off as harmless quirky behavior.  Originally, the idea was that one should not just park an AMS up against a facility wall and dispel the turret fire with a simple deploy; but, is seems that this is actually correct behavior.  But deployed Routers also exist.  Even a normal vehicle with turreted weapons becomes dangerous again since it only takes two players - one to man the weapon, one to constantly repair the vehicle - to remove most fo the risk from the facility turret.
8.  "Self reported" weapon fire may trigger in unexpected situations but the easiest means of cancelling this fire mode is moving slightly.  Moving slightly will cancel this mode of firing and fire a single normal test shot to determine if autofiring should continue.
9.  The conditions under which an automated turret directs its attention at a target player or vehicle should be reviewed and scrutinized throughly.  There was a bit of disagreement in Discord when we were trying to iron the details out and my implementation here might be too simplistic.  Keep in mind that radar really doesn't exist to the server right now.
10. Despite my strong affinity to vanilla behavior, turrets are so hobbled by being functional right now that I am allowing them to have some personality with the hopes of playing into what should be their strengths.  Some are already implemented, as indicated in the section above.   I am accepting suggestions.

___Addenda___
1.  Forgot to finish the override to the hashing operations of `Avatar` by making the hash code method point to the unique character id.
2.  A field in `CommonFieldData` which used for object creation for a large variety of entities, virtually all of them, was changed from `unk % 1 == 1` to `unk > 0`.  The true impact of this has not been explored just yet but the situation under the change was enacted certainly made sense.  I do not remember that situation precisely but I do remember saying "This makes sense".  I don't say that aloud for many reasons so it must have been particularly good.


ScrawnyRonnie:
A few first takeaways:

Spitfires no longer explode. Instead they just disappear, but after being "destroyed" can still shoot a few times. If they kill someone during that window it doesn't credit the owner, but instead the owner gets a kill assist. I think making them explode again would resolve that.

Did neutral tower or base turrets shoot at things or return fire? If they did shoot, there is an issue now where they remain aggro towards you even after the ownership is changed (even to your own empire). Being in an aircraft caused them to start firing, but after I got out, hacked the tower, then went back outside they were shooting at me instead of the vehicle.

The hit detection works well with soldiers and spitfires, but for some reason with vehicles the turrets will start to hit you after being behind something. I mainly tested with aircrafts. Being visible to them then hiding, a few shots will hit the terrain and not damage you as expected, but then it will start to hit you again for several shots despite never becoming visible to them again.

Overall it's good and reminds me of how scary a facility of unmanned turrets can be 😳
image

Dethdeath:
Spitfires are shooting at players that are moving around inside a base, thus revealing the players' presence and position.
Spitfires should not be shooting at infiltrators unless they have the interlink benefit.
I also tested sensor shield vs spitfires which seems to be working fine.
Some base type names should be changed in the kill spam for wall turrets: Comm Station-Interlink Facility, Comm Station Dsp-Dropship Center, Cryo Facility-Bio Laboratory.

Fate-JH:
> Spitfires no longer explode. Instead they just disappear ...

I'll look into that.  It's pretty strange behavior.

> Did neutral tower or base turrets shoot at things or return fire? If they did shoot, there is an issue now where they remain aggro towards you even after the ownership is changed (even to your own empire). Being in an aircraft caused them to start firing, but after I got out, hacked the tower, then went back outside they were shooting at me instead of the vehicle.

That's even weirder.

> The hit detection works well with soldiers and spitfires, but for some reason with vehicles the turrets will start to hit you after being behind something ...

This is the so-called "self-reported" fire mode.  It's purpose is to stop the turrets from not shooting at an unmanned vehicle, i.e., it stops shooting at a vehicle that parked right in front of it just because the driver got out.  When this happens, you should be able to clear yourself of aggro by driving around a little while staying out of the turret's line of fire.
It also affects driven vehicles less than planes because the latter will roll on uneven surfaces.

Fate-JH:
> Spitfires are shooting at players that are moving around inside a base, thus revealing the players' presence and position.

Other players can see the spitfire going all out at the wall of a facility for prolonged periods of time?  That sounds curious.  If the wall is there, the test shot that should have allowed the turret to realize "I can hit something" would never have connected,  Moreover, test shots should only be able to be seen by the player experiencing them, so another players should be incapable of seeing the turret scanning for targets that are not himself.  Moreover even more, the turret knows about as much as the rest of the server about level geometry - zilch - so it can't tell if your player is inside the facility and it is locked outside.

> Spitfires should not be shooting at infiltrators unless they have the interlink benefit.

Noted.

> Some base type names should be changed in the kill spam for wall turrets: Comm Station-Interlink Facility, Comm Station Dsp-Dropship Center, Cryo Facility-Bio Laboratory.

I'll have to check how that comes about.

Fate-JH:
Test completed.

__Arena__: VS Anguta, Ceryshen, the usual.
__Participants__: VS TestCharacter1 and NC TestCharacter4
__Setup__: TC1 held the main entrance door open for TC4 so that TC4 may enter the lobby.  The door closed.  TC1 waited outside and set up a spitfire turret not too far from the side of the building.  While the direction does not matter, not really, the initial direction was away from the wall.
__Rigor__: TC1 watched the spitfire turret.  TC4 ran and jumped around inside of the facility and fired his pistol into the wall to generate activity and noise (as far as the game is concerned).  From the perspective of TC1, the turret didn't so much flinch regardless of what TC4 did.  From the perspective of TC4, the sound of turret weapon discharge could infrequently be heard as it tried to hit TC4 but was stopped by the facility outer wall.  It was only after TC4 left the lobby back outside that the turret, from both of the client's perspectives, turned to TC4 and laid into him until he was dead.
__Conclusion__: predicted, acceptable behavior for both participants.

Dethdeath:
> Other players can see the spitfire going all out at the wall of a facility for prolonged periods of time? That sounds curious. If the wall is there, the test shot that should have allowed the turret to realize "I can hit something" would never have connected, Moreover, test shots should only be able to be seen by the player experiencing them, so another players should be incapable of seeing the turret scanning for targets that are not himself. Moreover even more, the turret knows about as much as the rest of the server about level geometry - zilch - so it can't tell if your player is inside the facility and it is locked outside. Is the test case really just have someone run about just inside and have another player watch a turret outside?

I just retested this, test shots can only be seen/heard by the player that is being targeted, not anyone else.

https://github.com/psforever/PSF-LoginServer/assets/77748540/b6d8022a-843f-4517-bf59-6c9947056417

The noise from all this is quite excessive, wouldn't it be better to have spitfires act the same way as wall turrets where the test fire packets are not sent at all? Only confirmed hit packets would be sent, that way whoever is being targetted won't see/hear all this?


> Noted. I also have logic that turrets may target players that were recently damaged by other players or who have opened fire with a weapon. Does this logic hold up as it currently is?

The turrets always acted independently of players, to have them support players in that way would be a more advanced AI than previously existed.

-Auto-repair on wall turrets is not working.

-When a wall turret gets jammed, it fires one shot that produces a damage indicator on the client, but no damage is done, this seems fine.

-Wraiths do not get targeted by any turrets whether they are cloaked or uncloaked

-In a phantasm, if you uncloak in front of a turret of any kind and recloak, they keep firing on you for quite a while even while moving away, seems to continue as long as the last shot was a hit. Perhaps it would feel more natural if the turret was told that the vehicle is recloaked and they would stop firing at the same time.

theonetrueduddy:
So I had a mess-around and... I think what you have here is a pretty good feel for the turret behaviour. They're scary when you're unprepared, as it should be.

Couple of things I think need addressing though;

1. Automated turrets deal no damage on return fire to enemy turrets that are manned. Tested this using a Phalanx and firing upon another Phalanx within range and each type of Spitfire and variant thereof.

2. In live automated Phalanx turrets would stop returning fire once below a certain threshold (50% according to the wiki), but on this pull they would return fire all the way until they died. Not necessarily a problem, but unknown if intended behaviour change. Could be problematic for infantry and light vehicles?

3. Spitfires only appeared to open fire on a MAX if it was auto-running. My recollection is that they would open fire on a MAX if walking and within range/sight.

4. Sensor shield preventing fall damage triggering turret response. When taking falling damage as infantry the turrets would not trigger if sensor shield was active at the time. Again unknown if falling/bump/ram damage is intended to trigger enemy turret response.

5. Had fun one I couldn't reproduce, whereby a non-rendered turret could be seen to be firing "test shots" constantly. See attached footage. Turret could be seen by other characters, so might have been a specific character issue. (Also being shot through the barrier sometimes here)

https://github.com/psforever/PSF-LoginServer/assets/89201577/563bd8e7-0617-4462-bfb7-b422091b2e42


https://github.com/psforever/PSF-LoginServer/assets/89201577/4ac7876d-0b05-49e5-81a0-3b6f138e6c1b


https://github.com/psforever/PSF-LoginServer/assets/89201577/6d1b53a1-7af4-468e-9a79-f37aacf2a206


https://github.com/psforever/PSF-LoginServer/assets/89201577/e80c04e1-4a30-466f-a0ed-12ef9ec0b366


https://github.com/psforever/PSF-LoginServer/assets/89201577/56a93de8-195e-41dc-870f-4a751b79076e

6. Phalanx turrets appear to respond with an initial shot back each and every time they are actively shot. Note first video a single shot is made, it then proceeds to kill me. In the second video I rapidly refire my suppressor and are killed **very** rapidly as a result. Not captured was just full-autoing using the suppressor, which still proved to be slower than the rapid refiring.

https://github.com/psforever/PSF-LoginServer/assets/89201577/7cb424fa-6088-4dcb-92c9-02e409319970


https://github.com/psforever/PSF-LoginServer/assets/89201577/8283bc03-0a42-4c50-bda3-eaedca8f87be

7. Last one appears to be a quirk I can't quite define. In the video below the driver of the vehicle fires upon it with the passenger in the gunner seat, and the nearby enemy turrets fire just a little bit and then stop. Not captured is when I swapped the driver in to the gunner seat and fired upon the vehicle using the passenger, at which point the turrets proceed to commence fire upon the harasser until death. Was able to reproduce this, which I had @Dethdeath witness.

https://github.com/psforever/PSF-LoginServer/assets/89201577/2daa9519-da0f-45d6-ac2b-74ba8831e54e

Fate-JH:
> The noise from all this is quite excessive, wouldn't it be better to have spitfires act the same way as wall turrets where the test fire packets are not sent at all? Only confirmed hit packets would be sent, that way whoever is being targetted won't see/hear all this?

That's why I want to play with the possibilities of invisible projectiles - to make an active turret less obvious.
As for the facility turrets, I do not know why it has this different behavior than the deployable turrets.  The same packet order is used for tests - aim, start shooting, stop shooting.  In fact, this method wouldn't work unless if did exactly what was required by those actions in that order.  I believe the reason might be the speed and the amount of animated activity a turret may engage in is too high so cancelling it as quickly as I do stops the animation but the results are still reported.
> The turrets always acted independently of players, to have them support players in that way would be a more advanced AI than previously existed.

I took that from the wiki page for the turrets.
https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/weapons/Adaptive_Construction_Engine.md#spitfire-turret
> Be aware however, using a weapon or taking damage (ie from falling) will reveal your presence to the spitfire regardless of sensor shield or crouch walking, and it will fire upon you anyway.

Speaking of which, I need to check the falling damage thing.
> -When a wall turret gets jammed, it fires one shot that produces a damage indicator on the client, but no damage is done, this seems fine.

It shouldn't fire but should just aim in the player's direction.  The aiming is admittedly a bit buggy but it definitely shouldn't fire that one shot.
> -Wraiths do not get targeted by any turrets whether they are cloaked or uncloaked

I can't find on which page it was listed, I found a mention that ATV vehicles were radar immune on the wiki.  Outdated perhaps?
> -In a phantasm, if you uncloak in front of a turret of any kind and recloak, they keep firing on you for quite a while even while moving away, seems to continue as long as the last shot was a hit. Perhaps it would feel more natural if the turret was told that the vehicle is recloaked and they would stop firing at the same time.

I do have a condition for this.  I have two possibilities for this issue, but the more likely one is that the rate of fire from facility turrets is such that the shots just get confirmed and cancelled and confirmed repeatedly and it takes damage just from the turret being unable to decide for long enough whether the target should be ignored.

Fate-JH:
>3. Spitfires only appeared to open fire on a MAX if it was auto-running. My recollection is that they would open fire on a MAX if walking and within range/sight.

From the wiki:
https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/items/Phalanx.md
https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/patches/3.8.9.md
> 4. Sensor shield preventing fall damage triggering turret response. When taking falling damage as infantry the turrets would not trigger if sensor shield was active at the time. Again unknown if falling/bump/ram damage is intended to trigger enemy turret response.

Refer to wiki:
https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/weapons/Adaptive_Construction_Engine.md#spitfire-turret
> 5. Had fun one I couldn't reproduce, whereby a non-rendered turret could be seen to be firing "test shots" constantly. See attached footage. Turret could be seen by other characters, so might have been a specific character issue. (Also being shot through the barrier sometimes here)

Shooting "through the barrier" is a possibility if the player's model has clipping that accidentally passes through the barrier when standing too close to it.  I have no way to confirm that it doesn't.  The fact that the turret can not be seen is a bigger concern for me since that would normally suggest the `ObjectCreateMessage` packet was malformed for the entity, but the turret still shooting indicates it constituted itself properly on your client.  I don't know.

Dethdeath:
> > The turrets always acted independently of players, to have them support players in that way would be a more advanced AI than previously existed.
> 
> I took that from the wiki page for the turrets. https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/weapons/Adaptive_Construction_Engine.md#spitfire-turret
> 

Sorry I was thinking of wall turrets when replying to that previously. For spitfires that should be fine.


> > 3. Spitfires only appeared to open fire on a MAX if it was auto-running. My recollection is that they would open fire on a MAX if walking and within range/sight.
> 
> From the wiki: https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/items/Phalanx.md https://github.com/psforever/wiki/blob/b314780d2ea504671ba1c59dcdc84438b066d70b/wiki/patches/3.8.9.md

Spitfires should be firing at non-running MAXes, wall turrets should not.

Dethdeath:
Cerberus/Shadow turrets don't count towards the limit shown on screen when they should.

spitlimit

Fate-JH:
I'm not certain what I'm looking at in this image.  Please use the keybind again to show the expanded combat engineering stat list.
That 14/15 should be the number of normal spitfire deployables only, not the limit imposed by the category of all three small deployable turrets combined.

Also, are the messages bugged like that?  I wonder what I changed ...

lazer1982:
Im going to Add to this as well. on the Topic of the Phalanx  Turrets Did not Fire on a Mossy if it was under 25 to 30% Power.   Since the Mossy is a Scout Air Craft as well when under 30% power ( But atm the Scout Radar Does not work as well..)

Fate-JH:
I've made some adjustments.  Turrets should properly respond to MAXes, Mosquitoes, cloaked Infiltration Suit users, and people falling who jump, at least based on the above feedback.  It may affect falling players appropriately but I did not pay attention to that.  Auto turrets now take self-reported damage from other auto turrets.

@Dethdeath I did not look at deployable reporting to the UI specifically just yet but I did reactivate proper turret deployable deconstruction code again.  I stunted that half of the turret deployable lifecycle off to test something, I think, and forgot to uncomment the code.  Parts of the code loop back on itself once the entity state changes and that makes it do something different so that might affect what you are seeing.

ScrawnyRonnie:
Deployable turrets explode again 👍

Wall turrets still continue shooting at you after an ownership change. TR tower was hacked by NC all while the NC char could hear the turrets shooting. The TR character did not see/hear the turrets shooting at the NC player so it appears to be test shots being heard by the NC character. When the NC character went outside and was in view of the turret, the turret shots could then be seen by both characters at that point and the NC turret took the life of its now ally. Something to do with exiting a vehicle and it for some reason not stopping the test shots even though the vehicle was also not in a position that it could hit.

Fate-JH:
> Wall turrets still continue shooting at you after an ownership change.

It doesn't matter what the ownership is.  The facility turret is shooting at **you**, not at your vehicle, which is what the algorithm does to get to your vehicle.  In fact, the above issue mentioned with the cloaking Phantasm might be the same thing.  For some reason, the facility turret, no longer going through cycles of test shots and not having a selected target, is stuck in some kind of loop where it's only shooting at your player avatar though we've told it to stop.  This also affects the occupants out of a ground vehicle.

The affliction can be solved.  If your player avatar gets shot by the turret even once, then goes back into cover, the normal missed target cooldown begins.  The turret stops shooting at you after that time passes.  There's no difference in the `stop` packet it receives.

Like I suggested above with the Phantasm, though the reasoning is different, I believe the problem is that the turret gets the `stop` packet too close to the `start` packet it had just received.  It skips over the `stop` command because it's in the middle of an unskippable part of its multi-shot barrage.  It's also possible the lack of explicit stop targeting packet is involved, because that's no a necessary packet for other turrets.

theonetrueduddy:
> I've made some adjustments. Turrets should properly respond to MAXes, Mosquitoes, cloaked Infiltration Suit users, and people falling who jump, at least based on the above feedback. It may affect falling players appropriately but I did not pay attention to that. Auto turrets now take self-reported damage from other auto turrets.
>

Unfortunately I'm not sure not all those changes are working? Went on 61007 to check and found the following;

1. Spitfires still shoot Infiltrators, including through Sensor Shield.

https://github.com/psforever/PSF-LoginServer/assets/89201577/3baaaddc-e21c-47da-9acc-599ce3b98eed

2. Spitfires not detecting MAXes at all now, walk or autorun. In fact in the video below, they didn't even fire on the MAX when it was shot by the Phalanx turret.

https://github.com/psforever/PSF-LoginServer/assets/89201577/5e34cf75-064e-4ca9-90cf-b10ac6da6ce7

3. Following from the above, I discovered Phalanx turrets **really** want to shoot MAXes that were autorunning. To the point they keep shooting when they can't hit the MAX. Allowing them to hit briefly then causes them to cease shortly after once back in cover. First video was just after the previous test with the Spitfires, second video was an explicit test of this.

https://github.com/psforever/PSF-LoginServer/assets/89201577/d4751f59-82d2-4473-bdcf-32f5f5a7fb19

https://github.com/psforever/PSF-LoginServer/assets/89201577/2894bf6f-d9ff-4d05-b7c2-12cf9a9d8be6

4. Phalanx turrets still shoot Mosquitoes that are flying at a slow speed. In the below video I flew towards the tower at setting 2 and was promptly attack at what seems to be tower SOI range.

https://github.com/psforever/PSF-LoginServer/assets/89201577/081624b4-3582-40a1-b097-dfc9de310900

5. Automated turrets returning fire on manned Phalanx turrets still deal no damage. An indicator happens, but alas not damage is dealt. I also observed the return fire from a Phalanx not dealing damage to a Phalanx I was hiding behind.

https://github.com/psforever/PSF-LoginServer/assets/89201577/1e4cb0d4-1111-48bf-805c-30d0e57c6545

https://github.com/psforever/PSF-LoginServer/assets/89201577/34136a03-78cd-4e4f-b864-133cc73052f3

6. Spitfire shooting falling infantry. I was actually checking the situation with Sensor Shield, but was shot while falling and with Sensor Shield on.

https://github.com/psforever/PSF-LoginServer/assets/89201577/c4f2102f-305a-4a4d-b058-416f5fd85a5c

Fate-JH:
I made a number of discoveries in regards to how turrets work, specifically how facility turrets work, and need to perform some retooling.  At least half of it is already done, to be honest, but I should update people.  In the meantime, I think MAXes and Infiltrators should be better represented by the current form of the branch.  Facility turrets may not be perfectly where they should be, but they're closer.
> Phalanx turrets still shoot Mosquitoes that are flying at a slow speed. In the below video I flew towards the tower at setting 2 and was promptly attack at what seems to be tower SOI range.

Mosquitos should be immune right now.  I need to work velocities out.
> Automated turrets returning fire on manned Phalanx turrets still deal no damage. An indicator happens, but alas not damage is dealt. I also observed the return fire from a Phalanx not dealing damage to a Phalanx I was hiding behind.

I interpreted the previous comment about damage between turrets incorrectly and will address that.
> Spitfire shooting falling infantry. I was actually checking the situation with Sensor Shield, but was shot while falling and with Sensor Shield on.

Just to clarify, that is considered correct behavior when without sensor shield.  This is some argumentive grease between "sound of footsteps when running/landing", which is more difficult to status, and "moving too fast", which is easier.

I have sensor shield logic garbled right now - for example, Interlink > sensor shield - right now so don't worry about testing it.

Fate-JH:
Behavior of the turrets should be sufficiently close to what it would and will look like on Live, I hope.  I am pleased with the operation of the facility turrets, though upgrading may prove jarring, perhaps even breaking the whole turret from time to time.

Target detection (radar) was the most difficult to disect because the information is scattered across various wiki pages, actually contradicts itself on a certain point, and does not necessarily pertain to only the spitfire turrets without actually clarifying that distinction.  In practice, AMS coak bubbles and AEGIS cloak bubbles provide perfect defense against detection even against Interlink Facility radar.  Ignoring those two, sensor disruptors also provide defense, but not perfectly, and sensor shield is the same, imperfect defense.  The Interlink Facility benefit detects all and motion sensors can aid.  To be honest, between conditions and interactions, the only feeling that I am left with by analyzing the security net that powered automated turrets is that it doesn't make enough sense as it exists.  Passive radar and visual cloaking trade strengths and weaknesses to produce convenient gaps, creating a series of exceptions rather than one of rules.

I've not yet spent a minute examining the fake projectile approach to making the turrets operate.

Another quality check at this stage would be beneficial.

theonetrueduddy:
> Behavior of the turrets should be sufficiently close to what it would and will look like on Live, I hope.

I'm pleased to say that behaviour _feels good_, at least from what I have been able to observe. In fact, I'm happy to report back on;
✅Automated turrets (both big and small) deal damage to Phalanx
✅Interactions with Infiltrators and MAXes feels good
✅Interactions with AMS bubble feels good
✅Phalanx interacting with Mosquitoes at variable speeds feels good
✅Automated Phalanx cease fire when reduce to below 50% hp

That said, did find some issues;
❌Automated Phalanx never restore function after being reduced below 50% (destroying/repairing/re-entering/exiting did nothing to resolve)
❌Tower Phalanx don't appear to automate at all. They don't engage targets they should track, nor return fire when fired upon
❌Spitfire and variants don't appear to deal damage to other Spitfire and variants (honestly this is a small one, not sure how often this would arise)
❌Inconsistently still seeing the "test firing" behaviour by Spitfires, but if I read what you wrote correctly: you're yet to look at this again?

What I didn't explicitly test, from what you talked about;
❔Any interlink benefit specific behaviour
❔AEGIS interactions
❔Sensor Disruptor interactions

Would like to see someone else test the basics and verify that my observations aren't unfounded.

Fate-JH:
This hasn't been updated yet but:
- Restored auto repair functionality to facility turrets and restored auto turret functionality upon sufficient repair (when I was shuffling behaviors, I forgot to reattach amenity autorepair to the turrets)
- Field towers with facility turrets now possess auto turret functionality (it was actually explicitly excluded)

Auto turrets still use the test fire system.  I've not started working on the hypothetical "invisible projectile" solution as of the time of the previous update.

Spitfires and spitfire variants (SSV) will not intentionally engage other SSV, but they will deal damage to other SSV owned by the target if they interfere physically with the target engagement process.  This will cause the primary turret to consider their target having gone MIA and will eventually lead to a target switch; however, and they may also stop firing depending on their personality.  While allowing SSV to become searchable targets to other SSV would be easy, there would be no way to validate the interaction, especially in the case of terrain interaction.  No client POV would be able to represent that terrain and SSV would have to inform the other that they are being damaged by the other directly through the walls and trees.
I will be resolving a potential turret ownership issue I have discovered and that may change the above considerations.

Fate-JH:
The soundless invisible projectile idea was a bust.

Excepting other bug fixes, this is the final condition of automated turrets for the time being.  Test everything, test nothing, I don't have much to say that hasn't already been said.

Fate-JH:
I discovered that the main validation timer needs to ensure running during retaliation because, without it, a target that draws the anger of a turret from outside of the turret's detection radius will not lose aggro until a new target draws aggro.  I also discovered that, during retaliation, a turret can hit a target no matter how far away the target is.   As I recall, it was possible to damage facility turrets from outside of their range and the projectiles just stop before they hit you; as it is now, there is no safe range.  Snipers beware.

Also, this claims it restarted the test server somehow.  If that did work, well, okay then.  Have fun with that.

Dethdeath:
-Sensor disruptors work with spitfires in interlinks. If a target is acquired outside of sensor disruptor range, the spitfire will continue shooting whilst having line of sight even if the target moves back into range of the sensor disruptor.
Ideally this behavior would mimic the way sensor shield works where once SS is toggled on, the spitfire stops firing immediately.

-AEGIS shields properly hide infantry from spitfires.

-Vehicles that drive into a cloaking bubble from an AEGIS/AMS continue to be fired upon by spitfires/phalanx, them being cloaked doesn't seem to matter.

-Spitfire health desync when blown up by driving over them:

https://github.com/psforever/PSF-LoginServer/assets/77748540/c970130a-0520-4858-8408-4f3b15eeff47

Fate-JH:
> -Sensor disruptors work with spitfires in interlinks. If a target is acquired outside of sensor disruptor range, the spitfire will continue shooting whilst having line of sight even if the target moves back into range of the sensor disruptor. Ideally this behavior would mimic the way sensor shield works where once SS is toggled on, the spitfire stops firing immediately.

On the topic of sensor shield, from https://www.psforever.net/implants/Sensor_Shield/:
> Sensor Shield will not protect you if you take damage. It will also not protect you if you are within the [SOI](https://www.psforever.net/locations/Sphere_of_Influence/) of a base with Interlink benefits and you try to use an item (weapons, REKs, etc.).

> -Spitfire health desync when blown up by driving over them:

Clients explicitly destroy turrets when they physically interact with vehicles?  I think I can accomplish that.
Your video is interesting for other reasons too - the turret stops shooting temporarily after being run over, direct just for one client, then for both, then it resumes on both.

Dethdeath:
> > -Sensor disruptors work with spitfires in interlinks. If a target is acquired outside of sensor disruptor range, the spitfire will continue shooting whilst having line of sight even if the target moves back into range of the sensor disruptor. Ideally this behavior would mimic the way sensor shield works where once SS is toggled on, the spitfire stops firing immediately.
> 
> On the topic of sensor shield, from https://www.psforever.net/implants/Sensor_Shield/:

I retested sensor shield behavior with turrets, seems I misremembered the on/off toggle tests I did some weeks ago. Taking damage will always result in keeping turret aggro (same as the wiki says), so there is no issue with the sensor disruptor behavior I mentioned.

Fate-JH:
I made the cloak bubbles cancel engagement as a test of the PR's ability to communicate with the branch (it passed).

Dethdeath:
Cloak bubbles work fine with vehicles now. Blowing up spitfires with vehicles also working.

PR #1169, opened 2024-02-19
Environmental Redo

Fate-JH:
There should be virtually no change in behavior.

All zone interactions normally exist on the entity side of the entity/control agency ouroboros model but zone environmental interactions normally violated that principle.  The entity half of the code, which should run every tick, only included the environment field detection code.  The explanation of the interactable fields are then sent to the control agency half which governed the code that determined how the ineraction was to be enacted.  Now, the only part of the zone environment code on the control side manages a timer; all of the code on the entity side manages "what happens".  In reality, porting all of the handling over to the control agency side was probably the more correct representation of how the ouroboros model should operate; but, this was only intended to align zone environment interaction with the other forms of zone interactions.  The "more correct" solution would be more involved.

One thing that did change is how environment field overlap is handled.  Previously, only one field type, e.g., water, lava, death, could be interacted with at a time.  Multiple unique fields can now be enacted at once if the fields have spatial overlap, for example, drowning until the very deep water, then death from being crushed by the pressure at the lowest depths.

Another change is that a player will recover from drowning while they are mounted in a vehicle which also means they will not drown while in a vehicle.  This gives more time to escape from underwater if possible at all when the vehicle is disabled and all players are kicked from their seats.

___Adenda___
I considered having passengers in open seats drown as long as the vehicle was being waterlogged; however, the situation made no sense since passengers in open seats don't take damage from certain sources when a vehicle also takes damage from that source.  The previous situation - both passengers and vehicle drowning at the same time - was due to a misundertood limitation.


ScrawnyRonnie:
Water seems good to me. Lava will now damage you or your vehicle indefinitely after getting out of it until you succumb to your wounds. On main it damages you while in it and stops after getting out. I'm not sure if this was intentional or if there should be a few seconds residual damage?
image

Fate-JH:
I mean, it's lava,  You _should_ be dead just by being close to it.

Nonetheless, I restored the ability to escape death by lava.

ScrawnyRonnie:
It stops damaging when you get out now. If you go back in again it will only damage you for a single tick though (new behavior).

Fate-JH:
Had too many methods to reset.

ScrawnyRonnie:
All good

PR #1170, opened 2024-02-19
Packet: DebugDrawMessage

Fate-JH:
There's really no reason attached to this packet.  We do not have any working examples of it and it serves no purpose in the current server.  Just working on finishing off the packets, maybe?


Fate-JH:
Reminder that the `debugmassive` command is related to ad billboards based on the `ChatMessageType` it produces.

PR #1171, opened 2024-02-19
Death By EMP

Fate-JH:
No longer can drivers hide their war crimes behind their vehicles.

Funny story, I should be able to sufficiently prove that this issue was resolved by dropping an APC EMP on someone's head and then telling them to `/suicide`.  And I did do that.  But, as explosions are intended to inherit information from the source and carry it down a potential chain of detonations to reach victim, I decided to go the full nine yards in this test. I spawned a whole APC with one character and waited the whole time until it was ready to pop.  I had another character on a different faction set up an HE mine and a Spitfire outside of their facility, then dropped a teleported APC on that character's head, just to kill him with an EMP blast that destroyed the HE mine that destroyed the Spitfire that destroyed him.  The APC then teleported back to where it originally was as if nothing had happened.  Sucked to be that guy.

### DIFF
--- a/src/main/scala/net/psforever/objects/SpecialEmp.scala
+++ b/src/main/scala/net/psforever/objects/SpecialEmp.scala
@@ -118,8 +118,8 @@ object SpecialEmp {
         case _                  => OwnerGuid_=(Some(owner.GUID))
       }
       Position = position
-      def Faction = faction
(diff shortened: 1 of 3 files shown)


Dethdeath:
Looks good.

PR #1172, opened 2024-02-21
Loadout Item Issues

Fate-JH:
The linked issue had nothing to do with the inventory tile size of the item.  Two different issues were being manifested:

1)  Mechanized assault exo-suits could deposit their arm weaponry in the backpack inventory space when changing between loadouts and exo-suits, resulting in a comedically huge 1x1 inventory item that shouldn't have been depicted there in the first place, let alone be able to be moved around like an other item.

2)  When adjusting loadout specifications between exo-suits, the item tiles in the backpack space tended to be misplaced or invisible.  This very frequently happened when trying to switch to an exo-suit that the player no longer has the ability to wear.  The issue was actually two-fold.  **Firstly**, the inventory space that the items were going to be added to was not correctly identified in some situations and the items were being calculated for the wrong index positions.  This caused items to skip grid spaces and trail off the edges of the inventory grid and could potentially have lead to the game crashing.  **Secondly**, the messaging for depicting existing equipment in infantry holsters relied on those holsters being unoccupied but earlier messaging for depicting existing equipment in infantry holsters _placed_ the equipment in some holsters, the early messages to self-destruct early.  Meanwhile, a second routine drew from that same list of items and found backpack-space for them without causing the same issue as with the holsters.  The result was items existing in a place where that item was not depicted, and items being depicted in places where that item didn't exist.

Any further inventory issues might require deleting the entity by globally unique identifier before drawing it into the next spot.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -350,6 +350,7 @@ class SessionAvatarHandlers(
             sendResponse(ObjectDeleteMessage(objGuid, unk1=0))
             TaskWorkflow.execute(GUIDTask.unregisterEquipment(continent.GUID, obj))
         }
+        drops.foreach(item => sendResponse(ObjectDeleteMessage(item.obj.GUID, unk1=0)))
         //redraw
         if (maxhand) {
           TaskWorkflow.execute(HoldNewEquipmentUp(player)(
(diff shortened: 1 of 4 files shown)


ScrawnyRonnie:
Did a few tests with the linked issue and had no problems.

Fate-JH:
Restored set exo-suit armor value back to where it should be, I think.

ScrawnyRonnie:
Looked good here.

Issue #1148, opened 2023-12-04
/capturebase causes problems with already hacked bases.


KingFeraligatr:
I've know for several years that using /capturebase on an already hacked base could cause problems and it was something you just didn't do. Fixing this would make using this command safe on already hacked bases.

Fate-JH:
Can you provide some details in regards to how hack-in-progress facilities were spoiled by using this command?

Dethdeath:
I think the timer would keep going and in some instances that meant you weren't able to resecure it. I've tested it on various continents on #1173 and it seems to all work as it should.

ScrawnyRonnie:
Yeah I think this was already fixed in #1147

PR #1173, opened 2024-02-27
Battle Island Facility Names

Fate-JH:
This change is entirely to make the parameters for `/capturebase` work on the Battle Islands in an intuitive way.

Technically, this can be used to give any facility or tower a much more approachable name.  Consider, towers.

### DIFF
--- a/src/main/scala/net/psforever/objects/zones/ZoneInfo.scala
+++ b/src/main/scala/net/psforever/objects/zones/ZoneInfo.scala
@@ -2,15 +2,21 @@ package net.psforever.objects.zones
 
 import enumeratum.values.{IntEnum, IntEnumEntry}
 
+final case class AliasLookup(
+                              zone: Seq[String] = Seq(),
+                              facilities: Map[String, String] = Map()
+                            )
+
 sealed abstract class ZoneInfo(
     val value: Int,
     val name: String,
     val id: String,
     val map: MapInfo,
-    val aliases: Seq[String] = Seq()
+    val aliases: AliasLookup = ZoneInfo.defaultAliases,
 ) extends IntEnumEntry {}
 
 case object ZoneInfo extends IntEnum[ZoneInfo] {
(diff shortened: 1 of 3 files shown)


Dethdeath:
Works.

Fate-JH:
I don't know if it is still a valid issue, but I was going to do something for #1148 when I discovered that the handling for `capturebase` in `ChatActor` looks like this at a certain point:
```
(factionPos, buildingPos, timerPos, buildingsOption, timerOption) match {
                case // [[<empire>|none [<timer>]]
                    (Some(0), None, Some(1), None, Some(_)) | (Some(0), None, None, None, None) |
                    (None, None, None, None, None) |
                    // [<building name> [<empire>|none [timer]]]
                    (None | Some(1), Some(0), None, Some(_), None) | (Some(1), Some(0), Some(2), Some(_), Some(_)) |
                    // [all [<empire>|none]]
                    (Some(1) | None, Some(0), None, Some(_), None) =>
                    /* wth this is actually the valid case!? */
```
... and there's more like it in the rest.  You're using the index of the param array to determine what kind of param a certain field should represent and then validating that it matches and, if it doesn't, then you have to pick a role for that position and subsequent positions to represent and then find out if one of the alternate guesses are permissible ...
Did I write this?

src/main/scala/net/psforever/objects/zones/ZoneInfo.scala
```diff
@@ -244,40 +256,59 @@ case object ZoneInfo extends IntEnum[ZoneInfo] {
...
       extends ZoneInfo(
         value = 30,
         name = "Ascension",
         id = "i2",
-        map = MapInfo.Map98
+        map = MapInfo.Map98,
+        aliases = AliasLookup(facilities = Map(
+          ("Zal", "Base_Alpha"),
+          ("Rashnu", "Base_Bravo"),
+          ("Sraosha ", "Base_Charlie")
```
Dethdeath, line 276: Extra space here prevents /capturebase Sraosha from working.

PR #1179, opened 2024-03-03
Initial Outfit DB structure

Resaec:
This migration (presumably) allows for the storage of all outfit relevant data (except titles).

Outfit
- each outfit has one entry in the outfit table
- each name is unique
- ranks are inlined (static 1:n, join unnecessary)
- faction is limited to 0,1,2,3
- decal is limited to 1 through 26 (inclusive)

OutfitMember
- each avatar can at most be a member in one outfit
- each outfit can only have one rank 1 (leader) member
- rank is limited to 1 through 8 (inclusive)
- there is a "quick access" index on outfit and avatar for rank 1 (leader)

OutfitPoint
- each (outfit, avatar) combination can only exist once
- a (outfit, NULL) combination can not be limited yet (not a big deal)
- points is limited to >= 0
- deleating a avatar will have his points remain as (outfit, NULL)

### DIFF
--- a/server/src/main/resources/db/migration/V013__OutfitStructure.sql
+++ b/server/src/main/resources/db/migration/V013__OutfitStructure.sql
@@ -1,0 +1,91 @@
+-- Tables for Outfits
+--
+-- Includes Outfit, OutfitMember and OutfitPoints
+
+/*
+
+This migration (presumably) allows for the storage of all outfit relevant data (except titles).
+
+Outfit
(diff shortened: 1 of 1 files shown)


Resaec:
- OutfitMember and OutfitPoint should both get a `created` timestamp to indicate date of creation.
- OutfitPoint should allow negative point counts as there are is a gm command to add/substract points.

Resaec:
continued in #1297

PR #1184, opened 2024-03-18
Cargo Fix

Fate-JH:
Non-players could never hope to escape.


Dethdeath:
Both linked issues fixed as far as I can tell.

PR #1186, opened 2024-03-19
Water on Forseral

Fate-JH:
I redesigned the ocean that surrounds the continent Forseral from a sheet of water at sea level to eight rectangular regions that surround the continent in a piecemeal fashion.  These correspond to the larger red rectangules in the image in the comments of the linked issue.  The lakes and ponds in the region were largely unmodified from the way they were (properly) represented in the files, except:
- the pond south of Lugh
- the small lake north of Anu
- the large lake between Anu and Caer

These were previously defined by the all-underscoring sea level and had to be defined individually.  Additionally, the pond at J17 needed to be corrected as it was encroaching the aforementioned lake between Anu and Caer above sea level.  I did not have a chance to check the entirety of the new coastline but I tried to select coordinate points that properly framed the coastal areas without treading too far in-land

The interaction known as `WithWater` was adjusted to remove risk of a stack overflow crash on failure to escape the drowning state.  (Admittedly I was moving at `/speed 6` at the time but the risk independently exists.)


Dethdeath:
Ran around some more, didn't come across anything weird.

PR #1193, opened 2024-04-15
No More Kill Trades?

ScrawnyRonnie:
Tested for a while with WeaponAndProjectileOperations HandleWeaponFireAccountability and couldn't get the projectiles to not do damage if the player shooting was dead according to the server. I then opted to try another way of doing it and from my testing there was a 0% kill trade rate. I just told the server to do nothing (not calculate damage) if the shooting player is dead. It would still log SHOTS_WHILE_DEAD, but those shots do no damage so it looks like a win. Hope you have lower ping than your adversary.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionData.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionData.scala
@@ -2208,6 +2208,8 @@ class SessionData(
         // auto kick players damaging spectators
         if (obj.spectator && obj != player) {
           administrativeKick(player)
+        } else if (!player.isAlive) {
+
         } else {
           obj.Actor ! Vitality.Damage(func)
         }
--- a/src/main/scala/net/psforever/actors/session/support/SessionData.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionData.scala
@@ -2208,8 +2208,6 @@ class SessionData(
         // auto kick players damaging spectators
         if (obj.spectator && obj != player) {
           administrativeKick(player)
-        } else if (!player.isAlive) {
-
         } else {
           obj.Actor ! Vitality.Damage(func)
         }
--- a/src/main/scala/net/psforever/actors/session/support/WeaponAndProjectileOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/WeaponAndProjectileOperations.scala
@@ -609,11 +609,13 @@ private[support] class WeaponAndProjectileOperations(
           if (tool.Magazine <= 0) { //safety: enforce ammunition depletion
             prefire -= weaponGUID
             EmptyMagazine(weaponGUID, tool)
+            (o, Some(tool))
           } else if (!player.isAlive) { //proper internal accounting, but no projectile
             prefire += weaponGUID
             tool.Discharge()
             projectiles(projectileGUID.guid - Projectile.baseUID) = None
             shotsWhileDead += 1
+            (None, None)
           } else { //shooting
             if (
               avatar.stamina > 0 &&
@@ -628,8 +630,8 @@ private[support] class WeaponAndProjectileOperations(
             tool.Discharge()
             prefire += weaponGUID
             addShotsFired(tool.Definition.ObjectId, tool.AmmoSlot.Chamber)
+            (o, Some(tool))
           }
-          (o, Some(tool))
       }
       collectedTools.headOption.getOrElse((None, None))
     } else {


ScrawnyRonnie:
Tested some weapons that don't normally result in a trade to see what this change would affect. This change only applies to infantry vs infantry (MAX included).

For these, if the projectile has been fired and is in-flight, but has not yet hit another player:
Decimator - No damage
Rocklet Rifle - No damage
Thumper - No damage
Grenade - No damage*
*(Plasma grenades will continue to do damage if the initial damage has started before dying)

If fired and you die before projectile hits/does what it does:
Scorpion - Does damage
Radiator - Does damage

For anything I didn't test you'll probably expect similar results. I think it is a worthy tradeoff.

Fate-JH:
The Scorpion and Radiator are subject to server-side damage and do not go through this damage routing.  Most other weapons mentioned do so.

As you have noted, this change exacerbates both ping discrepancies and projectile velocity discrepancies.  In controlled experiments, this should look like a inconsequential matter.  I ask for reconsideration, however, because this eliminates both the validity of projectiles that were produced in valid conditions and, in the greater scope of the project, group dynamics.  If one player throws a jammer grenade, and his ally sees him throw that grenade and moves in for attack, but the grenade does not affect the enemy target because that first player was killed before the grenade goes off, that will change the other player's perception of the account without providing feedback.  If there is feedback, there's no indication it will be timely either.  And, at the tail end of this hopefully, I will see a bug report for "attack did not damage" that I will struggle to disprove the validity of.  A much more applicable 1v1 situation: the tank lives because the player that fired what would be the decisively lethal missile is killed before the missile reaches its target.

I believe there's a change in line with the original intent: return a tupled value for every `if...else` case starting at line 607 of WAPO, eliminating the catch-all return at the end of the `collect`, but specifically return `(None, None)` in the case of finding the player dead.  This stops the server from locating the weapon that fired the shot and should restore the condition of "never producing the projectile in the first place".

ScrawnyRonnie:
I'm the tank scenario the projectile still does damage since I only applied this change to the player case.

If the other way you prefer would eliminate trades as well, I'd have to defer that change to you (or give me something to copy/paste in 😄). I started at that portion of code because it seemed to work at one point, although not 100% of the time, looking back in time at chat. I, as usual, went for a KISS approach (emphasis on the stupid), probably due to my still lack of knowledge. I was sure there were cons of doing this, some known and other unknown, but if it's a bandaid I think there will be less complaints with this than leaving it as-is until I'm guided to another way.

theonetrueduddy:
> ...but if it's a bandaid I think there will be less complaints with this than leaving it as-is until I'm guided to another way.

There might be less complaints, but I'd question what the signal to noise ratio for this really is.

Do some high profile regulars complain about trades? Yeah, particularly with the _current_ style of play. Do trades make much of a difference to your average player playing what is historically the _average_ style of play. I'm not so convinced.

In fact I'd go as far to say; projectiles that have left your weapon should have a chance of killing your attacker, to do away with that feels like dissonance from reasonable expectations. Sure people here to play "shooty shooty stairwell" might hate trades, but in the grander scheme of playing Planetside I don't think they matter at all.

ScrawnyRonnie:
> In fact I'd go as far to say; projectiles that have left your weapon should have a chance of killing your attacker, to do away with that feels like dissonance from reasonable expectations.

If by that you mean projectile fired before the server says you are dead, I agree that would be good. If that's not what you meant, that seems to take away authority from the server. What we're seeing now is Player A is dead according to the server, but they don't know they are dead yet and are able to get off a shot or two that ends up killing Player B. There was already a system in place that detects this (and was meant to prevent it - I guess it was never tested thoroughly to ensure it was working?) so this change isn't trying to do something that wasn't already considered years ago. I think regardless or player base or competitive nature, the server should be the authority, and ping differences, higher being worse, is something any online fps deals with.

When it comes to things like a Deci or grenade that was thrown before the player dies, if there is a way to make them still do damage that would be ideal because they were fired/released before dying (according to the server, not the client). Fate's suggestion sounds like it would do that?

Fate-JH:
As what Scrawny said, what I wanted to do originally didn't work that way because I mesed up the logic.  Originally, when I first wrote it, it might have worked correctly, with the logic being different, and that's lost in the fundamental changes to `(World)SessionActor` over the years.  I don't think it's worth tracing it back to see how badly I mangled things.  What's important is moving forward in a correct and convincing manner.

ScrawnyRonnie:
Okay, I reread Fate's suggestion and think I got it... It wasn't that complicated after all. Undid the initial change to SessionData as well.

From a few quick local tests, trades were still gone. I did test deci and nades and they DID do damage if released before dying so that is a plus.

theonetrueduddy:
> From a few quick local tests, trades were still gone. I did test deci and nades and they DID do damage if released before dying so that is a plus.

Nice! So long as the solution avoids the ill effects mentioned in your second comment, I think that's good for the game.

Looks like you're on track for that :)

ScrawnyRonnie:
Yeah, so long as the projectile left the weapon before dying, it will now hit. This goes for any weapon. With Suppressors, I saw a few trades after testing more, but I'd say those were acceptable because the shots were fired and in the air before dying. In actual fights I still think the amount of trades will be significantly less now that beyond the grave weapon fire will not kill people.

Dethdeath:
Looked good when I did some tests again earlier, the real test will be during the Sunday fight.

PR #1197, opened 2024-04-29
Mine Experience Fix

Fate-JH:
Attempting to resolve the issue where mine kills do not reward experience to mine owners, and comabt the exceptions that sometimes occur when a mine kill ocurs.  Additionally, there are isues regarding possible immuity to mine activation or damage to explore and patch, though the specifics for these undue immunities are not specifically outlined.


Fate-JH:
This will not be in the May 20th event.

PR #1200, opened 2024-05-08
Spectator Role

Fate-JH:
```
A secret stalker, all doors open before me
Like the paparazzi
Just a war correspondant, warping, flying freely
On you like paparazzi
My footage will be fair
Could catch people cheating, don't you dare
You'll never see me, hear me, speak to me, but I'll be
Like the paparazzi
```

Players could always become a spectator, there's a formal command in the game that allows it, if it's allowed, but the implementation we had was always cumbersome.  Asides from the command and a character flag, players who were spectators were just treated by the server as normal players and had all the capabilities of normal players.  All the command granted them is the ability to open most doors locally regardless of empire affiliation (not all doors: sanctuary doors to the HART remain locked unless the HART is docked, for example).  The server had to put in extra work to hide spectators from the view of normal players.  The server had to put in extra work to stop spectators from being damaged (and in some cases other players could still damage spectators and get kicked because of it).  Spectators could still drown; they could get killed by death fields; catch on fire and die in lava.  And, because the server treated them as normal players, they could get into strange situations that neither the server nor the client was prepared to deal with, "normal" to normal players, but abnormal for others.

To remedy this situation, normal behavior for the players on the server have been broken down into operations and functions, which also store state data, and logic, which controls implementation.  Normal players have spectators now have independent logic.  (Operations and functions and data are to be shared between different logic modes, where applicable.)  The logic of these game modes allows for spectators to be handled completely differently from normal players.  Special accommodations only have to be made once, rather than being built into the code, to handle all of the special edge cases that spectators must endure.  Also, this format reduces the number of branches that must be traversed to sequester proper spectator activity from proper player activity.

___Player Modes___
__Normal Mode__
Virtually unchanged from current player behavior, and still intermixed with game moderator functionality.  The distinction between operations and logic is based on normal players doing normal gameplay things.

__Spectator Mode__
Spectators are not characterized by what they can do but what they can not do.  The primary purpose of a spectator account is to be a snitch for the game moderators and a point of view for making films.  As a consequence, their skillset is focused on exerting no influence on the battlefield they observe and being able to obscure from their own point of view.  The typical spectator role grants them access to any door on a local level, not to affect any other player's concept of that door, for example.  Furthermore, a spectator:
- is invisible to normal players (and to each other);
- will never consume stamina for any action (jumping);
- can not mount any vehicle, turret, or machine;
- can not access any terminals or apply any loadouts;
- can not draw any weapon (except for the one they are provided with);
- can not pick up or drop items (and have no items in their inventory, except for the sole piece of equipment they are provided with);
- do not own their combat deployables or their vehicles anymore (might not work);
- will not be allowed to communicate with normal players (but may observe normal chat and may chat with one another);
- do not earn BEP, SEP, or CEP (but can serve as a stepping stone for other players's SEP or CEP earnings);
- will not belong to any squad;
- sees all other players as opponents (they are considered black ops, though it is not completely reflected);
- is not represented on the block map and therefore are not affected by environmental detection fielding or damage;
- has access only to certain commands and bang-commands;
- can not take damage and can not die (test it, but don't test it);
- are returned to their original faction's sanctuary when leaving spectator mode.

When a spectator swaps to the mode, they are outfitted in an infiltration exo-suit.  While a black ops soldier is normally green, black ops does not normally have access to the infiltration suit so it does not look any different from the faction's standards.  All of their equipment is stripped from them but they are given a Flail targeting laser into what is normally their melee equipment slot.  As it is their melee equipment slot, it can not be unequipped or swapped out.  On their shoulder are three of the same ribbon bars - the Bending Movie Actor commendation.  I need to remind that no one sees this but the player who is the spectator.
Spectators have three fixed implants and shortcuts in fixed shortcut slots for those implants.  The implants are always initialized and can only activate or deactivate.
- Advanced Targeting - intended to see other players's health bars
- Darklight Vision - intended to see cloaked infiltration suit users
- Range Magnifier - intended for planning shots and dymanic zooms

Spectators have no equipment but whatevere is assigned to what is normally a dedicated knife melee equipment slot.  The Flail laser pointer, while in this slot, can not be removed or dropped or changed.  It will visually perform lazing but there is no point.  The primary purpose of the pointer is to provide some item that can be used with Range Magnifier but offers the spectator no harmful capabilities.

Spectators have access to the following chat commands: `/zone`, `/warp`, `/fly`, `/speed`, `!list`, `!nearby`, and `!loc`, and, of course, `/spectator` for purposes of turning off the player mode.  Some other allowable commands are treated the same as `spectator off` and will result in the spectator player being sent back to sanctuary as a normal player.

As an unrelated note, spectators should be able to interact properly with telepads and routers and zipelines and teleportation rings.

___Features___
__`PlayerMode`, `NormalMode` and `SpectatorMode`__
Entry points for the logic that will govern this particular user character mode.

__`ChatActor`__
Originally an actor that lives above the operations and logic regions for the server, also known as support operations of the session, it was separated into its own operations funtions and data section, and logic sections for the different player modes. Currently deprecated and should be removed entirely.

__`Zone`__
There's a new disticntion called `AllPlayers` which does the same thing as `LivePlayers` previous did.  `LivePlayers` will now be used to track all non-spectator players.

__`UplinkRequest`, `UplinkResponse`, `UplinkPositionEvent`__
Packets that were originally deciphered for transcoding of packets related to the CUD (`command_detonater`), of at least four are determinate.  The developer was unable to determine how the requests dispatched by the CUD correspondss to the responses sent from the server to the client and was incapable of emulating the harmless activity of CUD operation.  They serve no purpose as use of the CUD has been rolled back and have no associated tests.

___Database___
Added table `Avatarmodepermission` to independently keep track of player mode permissions for both GM and for spectator.  An account can still be marked as GM, but that only means any player character owned by that account must be also marked for GM to assume GM activity.  Players marked as spectators may spectate independently of being GM's.

___Caveats___
1. Spectators can get temporarily stuck in the gantry hallway in the HART building lobby if the shuttle departs while the player is in the hallway.  They can not board the shuttle.  Secondly, the door to the HART shuttle when the shutle is not docked does not react to the spectator's unique status as a "personal master key" and there is nothing I can do about that.  Finally, they would never be kicked out into the lobby since spectators are not represented on the block map and the gantry exclusion fields require the block map to operate.  A simple `/zone` or `/warp` can free the spectator, however, or that player may wait until the shuttle returns.
2. The fixed implants have safe but peculiar interactions because they are not naturally being managed.  The second slot to fourth slots are set up to always have those implant shortcuts and any attempt to change them assert the implant shortcut in that slot.  What happens is that any attempt at replacing an implant shortcut will result in that implant shortcut in the same slot and a copy of that implant shortcut attached to the cursor.  If you place that shortcut into a different slot, it will become a duplicate implant shortcut.  Rinse and repeat as many times as desired.  To clear any copy of an implant shortcut or any shortcut attached to the cursor, exit cursor mode (tab, usually).  When the current spectator mode ends, all implants and shortcuts will be replaced with the avatar's normal anyway.
3. The Character Information window has a button that suggests it would allow a user to quit Black Ops while in spectator mode and it is flanked by a useless timer.  This button does nothing.

___Addenda___
1. GM's (game moderators) and CSR's (customer service representatives) will have their own player modes in the future with distinct allowances within the scope of the game.  When that happens, normal players will only be just plain normal players, unless they promote.
2. An idea is that spectators who are also GM-permitted will not have to return to sanctuary upon ending spectator mode.  For the moment, everyone is returned to sanctuary in the name of fairness.
4. An issue with the database attempting to create a table entry using the session id as a weapon id and the weapon id as a session id has been switched.  Future entries should be created with the proper property values now.
5. `ChatService` apparently had been using a super-stale reference to the player's session for a long time now.  When a channel is joined, that is the only time the "current" session is retained as a hard copy in the service.  To rememdy this, a reference to the dedicated session storage element is passed to `ChatService` instead, meaning it will never not have access to an up-to-date entry.
6. I already riffed Private Eyes in the last major feature update.  I can't use the same song twice.


Fate-JH:
Because the continuous integration tests servers are not equipped to handle it, the requirements for accessing spectator mode have been temporarily lifted.  Anyone should be free to access spectator mode using the appropriate commands.

Fate-JH:
Apparently, spectator mode is tied to the current character, not the client session, so when the character is completely reconstructed for its "spectator mask", the prior-set status of spectating is lost and it just becomes a normal [_faction_] character who has been become isolated from the game world and his fellow allies and closest enemies.  And the doors don't open.  To fix the issue, I pushed back the order of setting up the spectating status.

This prompt is apparently important.
spectator-mode
Whenever you can't see it, the mode doesn't officially count.

This should also adjust some other things that may have been accessible in the prior state.

theonetrueduddy:
Hopped on to retest the previously shared issues, but now /spectator doesn't appear to work at all?

The command is accepted in chat, but nothing appears to happen from the player's perspective, nor does anything appear to change for an observer.

Tested across 3 empires, and different accounts.

Fate-JH:
Ah, I was preparing for the "closing commit" wherein I properly activate spectator mode for database approved characters only.  (I want this over with and merged.)
No one on the test server is approved.

theonetrueduddy:
Can confirm that push appears to successfully resolve being able to open doors whilst in spectator mode 👍

The other less-pressing issues appear to remain as is. (i.e. CE, /setbr)

Fate-JH:
Commands restored, spectator ce icons wiped

PR #1202, opened 2024-05-12
Scoring

Fate-JH:
This resolution was decided after it was determined towers, which should never have given CEP upon capture or defense in the first place, were giving an inordinate amount of CEP under certain circumstances.  I am aware that capture CEP for major facilities is currently bugged - someone else is working on that issue.  Despite tower CEP being incorrect, as mentioned, maintaining it as is benefits healthy testing for future CEP corrections, so it will remain for the time being, but with a flat and severe reduction.

You will get 10 CEP for a tower capture event and you will like it.

Additionally, everyone's CEP will be reset to ~zero.  Most CEP in the database comes form towers anyway.  LLU CEP might be restored since it has nothing to do with towers.  Differentiating between major facilities and towers in the database entries is currently not possible.  Database entries for facilities captured and LLU carrying / hunting will not be affected.

Reminder: CEP currently does little but grant drip.


Fate-JH:
This will not be in the May 20th event.

PR #1203, opened 2024-05-17
Mine Kill Restoration

Fate-JH:
- restored experience awarded from mine kills
- removed situations where mine damage no longer works


Fate-JH:
Did we ever make an issue about this? Did I accidentally close an issue about this too early?

PR #1206, opened 2024-06-17
Multiple Game World Ports

Fate-JH:
Standard operation of the game server seems to have always been one port handling both inbound and outbound packets per component, and one port for the login component and one port for the game component.  In short work, two ports to handle every player that would ever play the game.  The first port is an underwhelming effort that only gets used during the start of session (before the "technical session").  The second port is an overworked monster that manages at least four packets per user per second and at least four times as many packets as connected players per second.  The actual outbound is perhaps greater as the frequency is more rigorous than four times a second and some packets may arrive in between each of those standard four.

The login server is still just one socket-port that everyone congregates their account tinformation, but now the game world ports can exceed the same limitation.  The number of ports that could be used by the game world is defined in the configuration file `application.conf` and each user is assigned the next available socket for their session.  No load-balancing algorithm is currently utilized because socket functionality and middleware functionality is still tightly woven.

The main goal is to reduce UDP packet receive errors and reduce congestion in general.

### DIFF
--- a/server/src/main/scala/net/psforever/server/Server.scala
+++ b/server/src/main/scala/net/psforever/server/Server.scala
@@ -14,7 +14,7 @@ import akka.{actor => classic}
 import ch.qos.logback.classic.LoggerContext
 import ch.qos.logback.classic.joran.JoranConfigurator
 import io.sentry.{Sentry, SentryOptions}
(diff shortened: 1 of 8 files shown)


Fate-JH:
I made a thing.
multiple-sockets-example

PR #1215, opened 2024-07-23
QoL: Character Select Screen

Fate-JH:
Up until now, the characters the you see on the character select screen have only been decorated in the same fake equipment load and appear as if in a fake destination zone.  In reinforced exo-suit, they don faction-specific pistols, standard and medium, the faction-specific heavy rifle, the faction-specific anti-vehicular launcher, and a katana.  (No one probably ever saw the katana.)  The reality is that the game's database remembers what they wore when they logged out; and, if they had not logged in once yet, it knows what they will wear.  The reality is that the game's database remembers where they were when they logged out; and, if they had not logged in yet, it knows where they will go.  These realities are now visible on the paper dolls displayed in the character select menu.


Fate-JH:
The test server is not starting properly.
Need to explore why.

PR #1221, opened 2024-08-02
Implant Terminal Hack

Fate-JH:
Since I didn't know how to actually do it, I gave a hacky-solution my best shot.  In the process, I discovered that we don't really understand the nitty gritty of this hacking packet works even if we can manipulate it in other ways.

Anyway, this should allow an implant terminal to be accessible when hacked by the enemy.  Implant terminal should "resecure" when hacked while the facility's CC has ben hacked, but not before that.  Important things to remember: implant terminals are composed of two entities - the interface that you purchase from, and the mech that you mount and above which a hood lowers around you.  To keep things simple, only one of the two will take damage; but, both may retain functional dispenser anchor points for repairing.  Similarly, both may become hackable.  For best use, I recommend standing aside it, or in front of it, and aiming upwards at the hood.


Dethdeath:
Tested various hackable objects along with the implant terminals, everything works as it should.

PR #1224, opened 2024-08-05
Dead People Implants

Fate-JH:
Only a person who dies will have their implants deactivate and deinitialize when that person dies.

This bug is actually hilarious. When someone died, EVERYONE'S implants deinitialize.


Dethdeath:
Working.

PR #1225, opened 2024-08-05
Drop It While It's Hot

Fate-JH:
Once again, the mysterious curse of the medical applicator thwarts loading the game on initial zone load.  So, we'll be ignoring it whenever it would normally show up on the Character Select screen and instead we'll be replacing it with a Flail targeting laser pointer.

Also, the MAX needs its arm raised when being depicted.


Fate-JH:
<details>
<summary>░M░Y░M░A░X░I░N░B░I░O░</summary>

my-cannon
</details>

Dethdeath:
The crashing problem is fixed. MAX preview also looks more appropriate.

PR #1226, opened 2024-08-10
Yellow Ownership

Fate-JH:
Deployables demonstrate familiarity with the color green or ownership with the color yellow.  For most deployables, a green circle will appear underneath them; for larger deployables, this status merely enhances the visibility of the name.  For ownership, however, the circle will be a yellow that is vibrant against most environments, and it's descriptive name is also yellow; for larger deployables. only the yellow name is visible.  Deployables that are neither owned or faction-aligned are merely depicted as they are in the game world, and lack a descriptive name.

___Caveats___
For all deployables but one this aspect of familiarity or ownership is consistent and unflinching from the time of being created in the game world.  This is a transcoder modification.  For Boomers, however, this familiarity or ownership is transferred according to possession of the complementary Boomer Trigger equipment.  If a player holds it, it's temporarily theirs; and, the rest of their faction will see it aligned to their cause.  The way Boomers are handled to transfer ownership does not reflect how the deployables actually transferred ownership on Gemini Live - complete reconstruction of the deployable entity once ownership is redefined internally - but this method operates correctly.

___Addenda___
1. Most of the changes in this PR are related to converting the `DeployItem` message from belonging to `AvatarService` to belonging to `LocalService`.
2. Some tests were corrected based on a new packet inclusion and the service event bus changes.
3. A bunch of fields have been properly named in the transcoders - `bits` and `size` and such.
4. The packet for `GenericObjectAction2Message` was transcribed because it appears in the packet capture for gaining ownership of a Boomer and it includes the GUID of the trigger being manipulated.  Does it help?


Dethdeath:
Yellow text/circles are working properly.

I came across some non-functioning CE though.

If a vehicle is pulled at the same base as the CE was laid, there is a good chance the CE won't work properly.

https://github.com/user-attachments/assets/ff9b4536-8a5e-428f-a6a0-97a04f64bd85

I placed a spitfire in range of the vehicle too and it didn't engage until the driver got out and back into the vehicle.

If CE is set up something like 2 grid squares away, and the vehicle drives or warps there it will work.

ScrawnyRonnie:
A mine will explode if laid by a corpse or already destroyed vehicle

ScrawnyRonnie:
When you die, any previously laid CE will change to a green circle. They are still owned by you, but the ring changes to green. Deploying more will be yellow, but only stays yellow for that life.

Dethdeath:
And if you zone and come back they are back to yellow 😝

Fate-JH:
Deployables should be yellow in both death and re-zoning.

The other concerns were made into issues.

Dethdeath:
They do indeed stay yellow now inbetween deaths/zoning.

Issue #1227, opened 2024-08-12
Invulnerable Vehicles


Fate-JH:
A report about fresh vehicles, hot off the spawn pad, not being affected by deployables properly.  Spitfires may or may not target but never do any damage.  Mines explode or may not explode but never do any dcamage.

theonetrueduddy:
I'm not confident this is about the vehicles being invulnerable as so much it is the turrets dealing no damage. [See this clip](https://youtube.com/clip/Ugkxumb_0YbHo-7snnVMRSR-E6LgMKCUMwcV?si=ItviHe-P6i-q0-6V) from 2024/08/11 from the Sunday fight where knifing the turret causes it to shoot me but do no damage.

Also got other footage of the turrets dealing no damage to me in a MAX, as well as not damaging other vehicles, all in that video.

Can't comment for mines other than they have felt like that they don't always work for a while.

Will try to reproduce my findings on test, but that'll have to wait sadly.

Dethdeath:
Afaik the new CE/mines issue only happens on #1226

I've made a lot of progress in unraveling the mystery.

It's not just invulnerable vehicles either, it's infantry too:
(also note how friendly fire from the boomers never stops working)

https://github.com/user-attachments/assets/6a84f3d5-e474-4741-bf43-0766c349a0f5

The mines in the video below were all laid by the same character:

https://github.com/user-attachments/assets/9c391513-cae2-4b94-b8aa-6437b0893f0a

Now for the explanation:
The bug is caused by the minelayer spawning on the same continent as the vehicle is pulled on. If the minelayer zones to another continent and comes back, their newly laid mines/boomers will now damage that vehicle, their old ones won't. That is how I laid the alternating working/not-working mines in the video above. The vehicle will stay immune to a minelayer that spawned on the same continent even if it's empty, as seen in the video featuring the Fury. As for the driver, their mine/boomer damage immunity entirely depends on where they spawned at. If they came from another continent before pulling the vehicle, they will be vulnerable to mines/boomers once they get out of the vehicle. If they spawned on the continent where the vehicle was pulled, they have a high chance of being invulnerable to mines/boomers.

Reproduction steps:
-Go to a continent with 2 opposing empire characters and make sure they each have a base they can spawn at.
-Suicide on each character and spawn at their respective bases.
-Travel or warp to one of these 2 bases with both characters.
-The character that the base belongs to pulls a vehicle.
-The character that is hostile to the base lays down some mines or boomers.
[...]

PR #1230, opened 2024-08-22
Facility Turrets Do Damage Again

Fate-JH:
I resolved at least one case of a turret, formerly incorrectly not doing damage to a vehicle target, now again correctly performing damage under the same conditions.

1a. Warp to enemy facility
1b. Travel to enemy facility
2. Hack enemy vehicle terminal
3. Pull vehicle
4. Move close enough to alarm the nearerst enemy facility turret
5. _Damage_

Refer to #1227.
I don't know if this covers all situations.
Now that a non-`Sidedness` solution to turret damage has been uncovered, consider justifiably merging #1226 and referring all turret issues in future PR's as native to `master`.


Dethdeath:
I tested some of the things Duddy mentioned and the turret damaged the vehicle every time. He should probably double check though.

theonetrueduddy:
Re-tested the cases presented by myself with regards to facility Phalanx turrets in #1227, happy to report that damage was correctly applied in all cases. 👍

Issue #1218, opened 2024-07-29
Drop pod client freeze


Dethdeath:
Originally reported by Equinaux in #bug-report. Whoever is in a drop pod will experience a client freeze if their pod is shot at by a wall turret or spitfire turret.
See video:

https://github.com/user-attachments/assets/d47aa76e-21ca-4d89-b8d7-8f963614151d

Fate-JH:
Low-effort prognosis: the drop pod is deleted from the client before the occupant gets to disembark, resulting in, naturally, the occupant being deleted too.  With the POV entity lost, the game doesn't know what region of the portal system it should display anymore and it culls out the world.  Also, the client softlocks, because it lost its POV entity.

Fate-JH:
I've never been certain whether the player is "deleted" or if it is merely lost forever in the game's memory when its parent entity loses the reference to it before truning over control.  It'd be more or less the same thing.

Fate-JH:
Boom.
``[ERROR] n.p.a.session.support.SessionData - AIDamage/Target: [player] has an invalid reference to thing with [droppod guid] in zone [name id]``

PR #1231, opened 2024-08-27
Vehicle Gating, Again

Fate-JH:
Attempting to change zones in a vehicle, e.g., a drop pod, caused a client crash because the client attempted to create an unmounted player without legs.  The lack of unmounted detailed placement data is an invalid condition.  This issue is not affecting Live as it was only introduced in #1226 and that has not yet been pushed to the server.  The circumstances of this problem were a line that had been dragged and dropped.  The line was undoing an assignment that had _just_ been changed to run a command that would be different if the field was set; the drag accidentally dropped the undo right after the assignment and then ran the command.

This has actually been an issue for me since I have started using my current mouse, where attempts to click something are interpreted  as an attempt to drag and has left me scrambling, on more than one occasion, to hit a "cancel" button before half the server's files drop themselves into the wrong directory.


Fate-JH:
Now I can actually test #1218

PR #1232, opened 2024-08-29
Turret-proof Drop Pod

Fate-JH:
A little birdie told me that a drop pod caught in turret auto-fire would cause the world to end.


Dethdeath:
Crashes are gone.

PR #1242, opened 2024-10-01
Spawning, Beep Boop

Fate-JH:
Spawn tubes are paired with special effects and sound alerts.  The spawn tubes attached to an Advanced Mobile Spawn vehicle only has the visual effects, but the spawn tubes installed inside of facilities and towers have both an effect and a sound.  The visual effect should technically have a flourish is related to the faction-affinity of the player who is spawning.   For perfect clarification, the effect should always appear under the player's feet rather than at the location of the spawn point (tube).

___Caveat___
The spawning player appears to stand on top of the spawn effect for a few moments after that player manifests, as if elevated by collision with the animation.  This is actually a product of the player character model being created in the game world before that player's client is ready to tell the character to "fall" to the ground.  It's only a coincidence that the animation should finish about the time the model plants its feet back onto the ground again and that should vary by latency more than anything else.  This levitation is never perceived by player's client.


Dethdeath:
Very cool to see and hear this working.

Seems to function in all the places it should.

PR #1248, opened 2024-11-04
Death Planes

Fate-JH:
Added death planes at 10m under cavern Hunhau and cavern Byblos.

Someone mentioned a disconnected bugged player being found having been ejected at the bottom of Hunhau and I noticed that this was not a feature of either zone.

Ascension's environmental setup is kinda suspicious but I don't think death planes would be useful to fix any issues there.


Dethdeath:
This can be merged?

Issue #1199, opened 2024-05-07
Capture Experience


ScrawnyRonnie:
There is an issue with capture experience being rewarded consistently. All of my testing has been done with towers due to the quick nature in capturing them. During our regular weekly fights, capturing and resecuring bases has rarely (if ever) granted experience. I've found there to be a few reasons why it results in 0 experience.

In TowerHackParticipation.scala, RewardFacilityCapture does some calculations for contributionVictorSize and contributionOpposingSize. For whatever reason, both or one of those will come out to 0. There is also a chance that baseExperienceFromFacilityCapture or totalExperienceFromKills (from FacilityHackParticipation.scala calculateExperienceFromKills) will be 0. Any of these resulting in 0 will make it so you do not get rewarded.

To demonstrate the randomness of this issue, I recorded myself hacking the same tower over and over again. Right before starting to hack it I killed the opposing player and had them respawn and stay within the SOI. There was a slightly less than 50% success rate in being granted cep. I had the different values print to the console to see why it didn't reward each time. Most of the failures was contributionVictorSize being 0, but the last one was contributionOpposingSize.
https://github.com/psforever/PSF-LoginServer/assets/73139382/04665ea5-5093-4efe-832f-83a59f102bd0

ScrawnyRonnie:
After reading over #1138, I found a few more things related to scoring that are not working.

- Attempting to delete rows from the `buildingcapture` table results in a SQL error.
![image](https://github.com/user-attachments/assets/b9962f0e-1881-4017-9c99-c41cdbb32636)

- The `kdasession` table never populates with data

ScrawnyRonnie:
A database error can occur under the right conditions. I believe the SQL function needs to be changed to handle conflict. Here is how to make this error happen:

A player gets multiple kills with a weapon such as a boomer or mine
These kills are the first kills by that player with that weapon during that play session

This can happen with each individual weapon (cause) that kills multiple players during that session

This is the error seen
[...]

PR #1253, opened 2025-01-19
Facility Capture Experience

ScrawnyRonnie:
Made a few changes after testing and println'ing basically everything to figure out the cause of 0 experience being granted upon capturing a base that was contested. I locally tested both 1v1, 1v2, and 2v2 scenarios. A few things of note to clarify how experience awards are calculated for anyone testing:

- The player that hacks the CC will receive CEP. Other players in the SOI will receive BEP. Once squads are no longer buggy, this will change and CEP will be rewarded to squad leaders and BEP to squad members (and solo players?)
- Every player does not need to get kills to receive capture experience. I had players just stand there and do nothing and receive experience because they were in the SOI. Collectively as an attacking faction, kill experience is summed up and rewards are calculated based on that, among other things.
- ~~You must be in the SOI at time of capture to receive a reward (with one exception - LLU). I do see an issue with this as LLU carriers typically require transport, so some unlucky driver will miss out because they leave the SOI. To be addressed later by actual devs (aka Fate)~~ Players that were in the SOI at some point during the hack will receive reward as long as they are still in the same zone when the hack, or resecure, completes.

I also did one test LLU capture to ensure the LLU runner received experience.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -59,7 +59,7 @@ class SessionAvatarHandlers(
     //TODO squad services deactivated, participation trophy rewards for now - 11-20-2023
     //must be in a squad to earn experience
     val charId = player.CharId
-    val squadUI = sessionLogic.squad.squadUI
+    /*val squadUI = sessionLogic.squad.squadUI
     val participation = continent
       .Building(buildingId)
       .map { building =>
@@ -117,7 +117,10 @@ class SessionAvatarHandlers(
           exp.ToDatabase.reportFacilityCapture(charId, buildingId, zoneNumber, modifiedExp, expType="bep")
           avatarActor ! AvatarActor.AwardFacilityCaptureBep(modifiedExp)
           Some(modifiedExp)
-      }
+      }*/
+    //if not in squad (temporary)
+    exp.ToDatabase.reportFacilityCapture(charId, buildingId, zoneNumber, cep, expType="bep")
+    avatarActor ! AvatarActor.AwardFacilityCaptureBep(cep)
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ScrawnyRonnie:
Oh, and don't test with towers. Something with how updatePlayers works results in players constantly being in the list and the next cycle it runs not being in the list. That explains the inconsistent results I had with #1199 when testing with towers.

ScrawnyRonnie:
I had forgotten to test rewards for resecuring a base. There was an issue causing the attacking and defending faction to be the same, so no experience was given. Seem to have fixed that now and received resecure experience in testing. The same things apply as far as getting a reward. Kills during the time the base was hacked are used to calculate the reward and a player does not need to contribute to the kills to receive it.

Another important note for testing: For reasons unknown, if you use /capturebase then test this, it will not (ever?) grant experience the first capture. The next capture has rewarded as expected.

Dethdeath:
I hacked a neutral base with 2 TR players and killed an NC player a few times. No BEP or CEP was given for this capture, is it because the accumulated XP was too low or because the NC player was not counted as a defending faction?

Fate-JH:
> I hacked a neutral base with 2 TR players and killed an NC player a few times. No BEP or CEP was given for this capture, is it because the accumulated XP was too low or because the NC player was not counted as a defending faction?

If I recall, for hacking NEUTRAL facilities, the game will attempt to determine the defending empire by checking who originally owned the facility.  If, for some reason, the original owner was "no empire", it will then attempt to find the empire that had the most contribution, ignoring the empire that hacked the facility.  If it still can not find _anything_, it will try to use TR or, as a fallback to the fallback, NC.

If the TR force a VS facility to go NEUTRAL and then attempt to capture it, but the NC go to fight them for it instead, the TR will probably not get capture rewards for it should they succeed due to the lack of VS interaction within the hack period.

ScrawnyRonnie:
Not sure what happened with your neutral base test. I just tried on this server with TR/NC at a neutral base and got exp. 1 TR hacked base and killed 1 NC once close to the hack completing.

Dethdeath:
I did another test at a neutral base, LLU base this time, same characters and got XP. Might've been a fluke then.

PR #1254, opened 2025-01-26
Doors Again Change Player's Sideness

### LINKED ISSUE #1249: Boomers don't work sometimes
Boomers don't always work, some bases are affected and others are not.
I first noticed these problems at Akkan and Hanish on Ishundar, but there seem to be more bases where boomers don't always work, provided certain conditions are met.
The conditions that had to be met in the tests I did was that the target of the boomer had to spawn at that base. The boomer deployer had to spawn at an AMS at that base.

The footage below was recorded at Ikanam on Amerish.

https://github.com/user-attachments/assets/bb97bb15-6478-479d-8ca6-6f1d9834eb3b

I tried various other bases on Amerish as well, but couldn't reproduce this outcome. Even Ikanam isn't always consistent. Hanish is probably the most reliable for it.

### LINKED ISSUE #1251: Sidedness ignored
There's a problem with CE ignoring sidedness. A PR (#1192) previously fixed the excessive noise from spitfires in the courtyard, but it has returned.

End result is the noises heard in this video are back: https://github.com/psforever/PSF-LoginServer/pull/1166#issuecomment-1915417117

ScrawnyRonnie:
Should fix #1251 and fix #1249 

Enabled a debugging feature that shows a chat and on screen message when sideness is checked by a door. I played with lower values for the testingDepth, but I was able to "miss" the door check at a Tech Plant vehicle bay door due to its size. As unlikely as this scenario is, parking a vehicle right against the door and using surge with light armor (infil) I was able to jump high and far enough for the sideness to not change. 4 seemed to prevent that possibility.

### DIFF
--- a/src/main/scala/net/psforever/objects/avatar/interaction/WithEntrance.scala
+++ b/src/main/scala/net/psforever/objects/avatar/interaction/WithEntrance.scala
@@ -65,7 +65,7 @@ class WithEntrance()
                              obj: InteractsWithZone,
                              door: Door
                            ): Sidedness = {
-    debugInteriorCheck(obj, door)
+    //debugInteriorCheck(obj, door)
     strictInteriorCheck(obj, door)
   }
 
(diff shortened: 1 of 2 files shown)


Dethdeath:
Both issues are fixed from what I can tell.

Fate-JH:
Excellent work, Scrawny!

Issue #1256, opened 2025-01-27
Capturebase all lag spikes


Dethdeath:
Capturebase all can still cause a desync and/or DC for the player using the command, probably because they are on the continent it is used on, as reported in #731
Apart from that players on other continents will also suffer from lag spikes when the command is used, especially if it's used a few times in a row.

Here's a clip from Scrawny's stream around the time the command was used:

https://github.com/user-attachments/assets/40c7cc73-a932-4f9a-a4b4-2fa587b05b6f

ScrawnyRonnie:
Another minor thing with this command. Players in other zones will not see the towers change empire, just major facilities. Logging out and back in or going to the zone shows their new correct faction owner.

![Image](https://github.com/user-attachments/assets/6f590c09-008a-4613-ae39-9eb1d8fcc912)

ScrawnyRonnie:
Correction: It's not just this command with towers. Hacking a tower normally does not update it for players in other zones.

PR #1258, opened 2025-01-28
Spread The Wealth (With My Squad)

ScrawnyRonnie:
This is my humble attempt to introduce shared squad experience when a squad member gets a kill. Let me preface this by saying I don't know how this all works. I simply looked at existing things that the code does and tried to replicate. Is it _a_ way to do this? 👍 Is it **the** or a _right_ way to do it? 🤷‍♂️ It did result in the desired outcome, but if it is inefficient it can be binned now or when a better way is introduced.

**My squad members will get a split of my kill experience!?** Maybe... The requirements are:
- Be in the same zone as the squad member that got the kill
- Be alive (or still a corpse on the ground). Deconstructing and sitting at the map screen also counts as being alive

If dead or not in the same zone, you won't receive this benefit. Dead = having clicked the countdown timer button to select a respawn location.

The amount received by the other squad members is based on the amount of experience the killer received and the size of the squad. I just did basic math: experience / size of squad. The experience given to the killer is in no way effected by this. This is just an added benefit to squad members.

Added a temp(?) fix for #1259

### DIFF
--- a/src/main/scala/net/psforever/actors/session/AvatarActor.scala
+++ b/src/main/scala/net/psforever/actors/session/AvatarActor.scala
@@ -3039,6 +3039,8 @@ class AvatarActor(
     if (exp > 0L) {
       setBep(avatar.bep + exp, msg)
       zone.actor ! ZoneActor.RewardOurSupporters(playerSource, historyTranscript, killStat, exp)
+      zone.AvatarEvents ! AvatarServiceMessage(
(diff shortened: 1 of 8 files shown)


Dethdeath:
Squad XP is working well. Tested it with a squad of 4 and 2 squads of 2.

ScrawnyRonnie:
Addressed my comment about towers in #1256 here

PR #1260, opened 2025-02-03
Deployable Fixes

### LINKED ISSUE #1246: CE turns against its own faction
When a player crashes and relogs, their CE can turn against their own faction. When 3 mines were placed during a life before the crash, sometimes only one of them loses its ring, at other times they all lose their rings and become hostile. 

I'll just include the entire footage of the mines losing their rings:

https://github.com/user-attachments/assets/82936357-fcdb-40aa-b21a-b11c64fbc799

Can also happen with Spitfire turrets:

https://github.com/user-attachments/assets/e9f6fe15-8151-4719-9a33-c1849b1c9976

Reproduction:
ALT + F4 works fine to reproduce this, but it doesn't happen every time. The player who owns the CE getting killed before relogging might also play a role in this, I haven't been able to confirm it 100%.
I recommend placing 3 mines/spits at a time and then have the player "crash" and relog.

ScrawnyRonnie:
This aims to address a number of issues with deployables. The changes are made to:
- Fix #1246 by retaining the original owner's empire so CE does not turn against its own faction. I couldn't think of a scenario where ownerless CE would need to belong to Neutral. I also have no idea why _sometimes_ a random number of CE has this happen.
- The owner of a deployed router telepad should be able to see it and use it again after dying. Current behavior: If the owner of the pad were to die and respawn, the telepad would appear to be stuck in a state of getting ready to deploy. The owner could still teleport from the router to the telepad, but not go back the other way.
- If your field turret, for example, is hacked, this change will remove that deployable from your on screen counter and "toolbox". Current behavior: If your field turret were to be hacked and you deployed another one, the first one that now belongs to another empire would deconstruct, likely confusing the player that could be mounted inside it.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -2494,6 +2494,11 @@ class ZoningOperations(
           reclaimOurDeployables(continent.DeployableList, player.Name, reassignDeployablesTo(player.GUID))
         )
       )
+      //do this to make my deployed telepad appear that way
+      continent.Vehicles.filter(router => router.Definition == GlobalDefinitions.router && router.OwnerName.contains(player.Name))
+        .foreach { obj =>
+        sessionLogic.general.toggleTeleportSystem(obj, TelepadLike.AppraiseTeleportationSystem(obj, continent))
+      }
       //begin looking for conditions to set the avatar
       context.system.scheduler.scheduleOnce(delay = 250 millisecond, context.self, SessionActor.SetCurrentAvatar(player, 200))
     }
--- a/src/main/scala/net/psforever/objects/ce/DeployableBehavior.scala
+++ b/src/main/scala/net/psforever/objects/ce/DeployableBehavior.scala
@@ -68,7 +68,12 @@ trait DeployableBehavior {
       if DeployableObject.OwnerGuid.nonEmpty =>
       val obj = DeployableObject
       if (constructed.contains(true)) {
-        loseOwnership(obj, obj.Faction)
+        if (obj.Definition.DeployCategory == DeployableCategory.Boomers) {
+          loseOwnership(obj, PlanetSideEmpire.NEUTRAL)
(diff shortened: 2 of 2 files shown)


Fate-JH:
> I couldn't think of a scenario where ownerless CE would need to belong to Neutral.

When you drop a Boomer Trigger, the Boomer turns Neutral.  If you did not, the check that determines whether it can be affected by formerly friendly jammer grenades would collapse.
> The owner of a deployed router telepad should be able to see it and use it again after dying.

Sounds like a message being left out of the party.  Clarification: is the telepad within render distance of the death and respawn? Separately: is the telepad be in render distance but on the opposite side as the owner, e.g., owner outside, telepad inside?

ScrawnyRonnie:
> When you drop a Boomer Trigger, the Boomer turns Neutral. If you did not, the check that determines whether it can be affected by formerly friendly jammer grenades would collapse.
 
Ah, yes thank you. Boomers should be treated as they were.

> Sounds like a message being left out of the party. Clarification: is the telepad within render distance of the death and respawn? Separately: is the telepad be in render distance but on the opposite side as the owner, e.g., owner outside, telepad inside?

From recent fights where I started to notice this when I was the pad owner, the sidness on spawn relative to the sideness of the telepad didn't matter. It would still not function for only the owner. I was always within render distance, but a quick test on the main server also had the same outcome when respawning far away and traveling back to the telepad (no commands used to warp/etc.)

Dethdeath:
-CE no longer turns against its faction inbetween crashes/relogs.
-Telepad owner continues seeing it as deployed inbetween deaths.

Dying and respawning removes the ability to deconstruct CE and telepads via the map, zoning or relogging fixes it.

ScrawnyRonnie:
> Dying and respawning removes the ability to deconstruct CE and telepads via the map, zoning or relogging fixes it.

This is something I noticed too. I'd go separate issue for this since it pre-existed to this PR and I couldn't get it with the few things I tried.

Dethdeath:
Sounds good.

PR #1265, opened 2025-02-20
Lock The Door To My New Ride

### LINKED ISSUE #1264: Driver seat defaults to Empire
When you pull a vehicle, the permission for the driver seat is 'Empire'. It should be locked to just the owner.

ScrawnyRonnie:
Two things with this PR.

Fix #1264 
Tested pulling a vehicle as well as unlocking and allowing another player to take ownership of the vehicle. In both cases, when the owner dismounts, the driver seat (and trunk for good measure) were locked to other players. ~This also turned out to fix a vehicle appearing unlocked, but would say "you are not the driver" when attempting to mount and the driver door was in fact locked (this would happen prior to this new issue appearing).~

Also made a change to the fix I applied for router telepads in #1260. It now applies this for the telepad owner, not the router owner. This is because they aren't always the same player (as I experienced this past Sunday when my telepad stopped appearing "ready" to me when someone else jumped in my Router).

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -2495,9 +2495,13 @@ class ZoningOperations(
         )
       )
       //do this to make my deployed telepad appear that way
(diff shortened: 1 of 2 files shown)


Dethdeath:
Briefly tested the vehicle locks, didn't run into anything weird.

PR #1266, opened 2025-05-21
Capturebase Tweak & Misc. Things

### LINKED ISSUE #1256: Capturebase all lag spikes

ScrawnyRonnie:
- An attempt to fix #1256 by adding a delay between the captures to not overload the client. Seemed to work fine locally.
- Squads off
- Minor log fix for hardcoded:
image


Dethdeath:
The staggered capturebase all is working fine, it makes me wonder if 500ms between captures would also still not cause any lag. I think it would also be neat if bases were captured before towers, but if that requires a lot of extra code then it's probably not worth the effort.

Can confirm squads are disabled.

ScrawnyRonnie:
I played around with lower values, but things like reloading a weapon and zoning would sometimes delay by a few seconds. Never had a desync happen, but it just felt laggy.

PR #1267, opened 2025-05-22
Enemy Faction Equipment in Favorites

ScrawnyRonnie:
Needed to remove a filter that prevents you from being able to load other faction equipment in a favorite. This will allow favorites to be used outside of Desolation and still give you the equipment, but it will be reverted after Sunday.


Fate-JH:
The drop predicate also influences cavern equipment storage, router telepad management, and boomer trigger (command detonator) management.

Recommendation: adjust the predicate directly in https://github.com/psforever/PSF-LoginServer/blob/0621cb4fcd21fd8b1ad5661ba895f09c90e6068d/src/main/scala/net/psforever/objects/serverobject/containable/ContainableBehavior.scala#L675 by commenting out the last line `(faction != tplayer.Faction && faction != PlanetSideEmpire.NEUTRAL)`.

ScrawnyRonnie:
Good call. I did look at that code, but hastily dropped the filter thinking this will cause enemy equipment to no longer drop on the ground when you change loadouts. Who cares... get more from the terminal 😄 In my few minutes of testing I had already littered the ground with so much stuff so it not dropping will probably be better anyways.

PR #1268, opened 2025-05-26
Undo All Access and Disable BFRs

ScrawnyRonnie:
Desolation set back to normal and no more enemy equipment in favorites

BFRs have long been suspected of causing crashes. It seems like when one of them is around, people crash. It could be coincidence, but there's only one way to know for sure. If crashes continue during vehicle fights, we may need to disable more over time.


ScrawnyRonnie:
During the fight I noticed spitfire turrets trying to shoot me after I respawned in the base. The same turrets tried to shoot at me outside as well so it's as if they ignored sideness or had the wrong sideness (in between?) applied to them. After testing a bit I was able to reproduce this. I think forcing the sideness of spitfires will fix this.

Dethdeath:
Spitfire issue seems fixed.

PR #1271, opened 2025-06-09
The Pilot Goes Down With The Plane

### LINKED ISSUE #1077: All Your Bases Belong To Me!
When a specific empire captures the bases on a continent, all of the towers remain unchanged.

What is supposed to be happening is when a specific empire hacks all of the bases on a continent, all of the towers are supposed to flip to the empire that captured the continent.

### LINKED ISSUE #1083: Driver bail vehicle momentum
This past Sunday I captured an Skyguard appearing to fling into the base on the minimap. I believe it was caused by the driver bailing. During the animation of exiting the vehicle, it continues forward to other players. I was able to reproduce this:

https://github.com/psforever/PSF-LoginServer/assets/73139382/6c16f243-a905-44ca-88e0-1bf9c68dbe00

### LINKED ISSUE #1205: Repair messages not showing
while in a max and get reps from teamates does not show who and how much they rep you for

### LINKED ISSUE #1234: Generators showing as critical even at full health
In the base status window on the map, generators can be shown as critical even if they have been repaired back to full health.

ScrawnyRonnie:
There's nothing more annoying than chasing an aircraft that has a sliver of armor remaining and seeing the pilot bail out. This introduces the bail malfunction system. The conditions are _mostly_ based on what is available in the wiki, which is:
- Only applies to aircraft pilots and gunners. Passengers can freely bail at any time
- Once the aircraft is below 35% of its maximum armor, bailing may be prevented
- As armor decreases, so do the chances of being able to bail

The additional condition is how long it has been since you have taken damage. As armor decreases, more time since the last damage taken is needed for the chance to bail to be allowed. While this favors the pursuer, it was my best effort at adding a "delay" between bail attempts as I couldn't see a way to prevent spamming the bail hotkey to roll the dice more quickly to hit the RNG needed.

Also added some fix attempts to fix #1234 and fix #1083.  Don't ask on the latter; I tried different things and that worked... The vehicle can tend to slowly inch forward to where the driver bailed. I think it has to do with how long the exit animation takes, so the vehicle will initially be slightly behind the driver to other players. While it is "moving" you can still enter/exit the vehicle, so this is better than what was happening before. 

Edit: Fix #1077 using the same process that capturebase all uses to capture any towers not owned by the faction that just captured the last base on the zone.

Edit edit: Fix #1205 added in messages sent to the player being repaired or the owner of a vehicle being repaired.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/csr/MountHandlerLogic.scala
+++ b/src/main/scala/net/psforever/actors/session/csr/MountHandlerLogic.scala
@@ -298,6 +298,10 @@ class MountHandlerLogic(val ops: SessionMountHandlers, implicit val context: Act
         }
 
       case Mountable.CanNotDismount(obj: Vehicle, _, BailType.Bailed)
+        if obj.Health <= (obj.MaxHealth * .35).round && GlobalDefinitions.isFlightVehicle(obj.Definition) =>
+        sendResponse(ChatMsg(ChatMessageType.UNK_224, "@BailingMechanismFailure_Pilot"))
+
+      case Mountable.CanNotDismount(obj: Vehicle, _, BailType.Bailed)
         if {
           continent
             .blockMap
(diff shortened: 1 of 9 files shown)


lazer1982:
Keep the good work up guys

PR #1276, opened 2025-06-17
Mine Distance

Fate-JH:
Corrects an underappreciated trigger distance on mine checks.

Now everyone should die more easily.

This is a ... good thing?


Dethdeath:
Infantry stuff is fixed now.

PR #1277, opened 2025-06-29
Two Little Fixes

### LINKED ISSUE #1269: MAX armor amount desync
An avid MAX user told me yesterday that whenever they disconnect and relog their MAX appears to have full armor again.
This is quite easy to reproduce by closing the client while in a damaged MAX suit.
The end result is that the MAX appears to have full armor to the user, but other players see the actual armor/health amount.

-If the MAX has no more armor left, upon relog it will appear to be full to the user and it will stay full even after they take damage, only their healthbar will go down until they die.
-If the MAX has armor left, upon relog it will appear full to the user until they take damage and it resyncs to what the value actually is.

Reproduction:
Grab a MAX and take some damage.
Alt + F4 the client.
Relog and check armor.

### LINKED ISSUE #1272: Kicked players are timing out instead of logout
When kicking a player the session is not closed correctly.
Other players see the kicked player timing out for 60s instead of disappearing like with a proper logout.
Is this intended behavior?

ScrawnyRonnie:
Closes #1269, closes #1272

One liners that seemed to do the trick for each issue... For the max armor, it was also with normal log out/in and zoning in general.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -3134,6 +3134,7 @@ class ZoningOperations(
       statisticsPacketFunc()
       if (tplayer.ExoSuit == ExoSuitType.MAX) {
         sendResponse(PlanetsideAttributeMessage(guid, 7, tplayer.Capacitor.toLong))
+        sendResponse(PlanetsideAttributeMessage(guid, 4, tplayer.Armor))
       }
       // AvatarAwardMessage
       //populateAvatarAwardRibbonsFunc(1, 20L)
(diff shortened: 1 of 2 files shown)


Dethdeath:
Logout for kicked players works.

MAX armor amount for the MAX user is fine now, also during logout and zoning. For other players, the MAX shows as having full armor if they zone or relog, until the MAX takes damage and it resyncs.

ScrawnyRonnie:
Deth uncovered even more. Turns out nobody is shown the actual armor of a MAX. So this third edition is for a player entering the zone to be able to see the actual armor of MAX units already in the zone (that are not full armor and have not taken damage or been repaired) since they got there. I think this covers everyone now 🤞

Dethdeath:
Everything is working now.

PR #1279, opened 2025-07-06
Warpgate Fixes

### LINKED ISSUE #1237: Ceryshen Geowarp
This geowarp keeps sending me to around 174 on the Z axis, which is under the map.

cgeo

cerygeowarp

ScrawnyRonnie:
Discovered a few things with the different types of gates and made some changes to fix them:
- Cavern and small gates were using regular warpgate definitions instead of their own. Also adjusted their innerRadius a bit to prevent spawning too close to the center and immediately zoning back where you came from. This was only happening to me in caves, but I changed both to match what is was in the adb. I haven't had any instances of spawning in/under the ground in a cave after this change.
- Should fix #1237
- Vehicles have been maintaining their momentum while going through a direct gate (not a broadcast gate). A vehicle is now stationary on the other side instead of smashing into something or already being outside the gate by the time the zone finishes loading.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -1269,6 +1269,7 @@ class ZoningOperations(
         ICS.FindZone(_.id == zoneId, context.self)
       ))
     } else {
+      vehicle.Velocity = None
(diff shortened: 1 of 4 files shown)


Dethdeath:
Ceryshen geowarp is fixed.

PR #1281, opened 2025-07-15
Login Database Test

Fate-JH:
Originally, I was working on separation of the login functionality with game world functionality, an age-old dream for this project, though not an absolutely required one.  Much like previous attempts, I abandoned my effort before completing it; but, I also have a fairly good idea what we could do to facilitate the split when we finally get around to finishing the work off.  In the mean time, these are the changes I performed to the `LoginServer` operation within the current dual-purpose functionality of the project, without features like referencing a completely separate database of server connections and names to display to the client.


ScrawnyRonnie:
I logged in a few times. Is there anything else to test?

PR #1282, opened 2025-07-17
The Command Uplink Device (CUD)... Kinda

ScrawnyRonnie:
After combing through the old captures, I thought to try to tackle a new feature. While it would be better to be fully implemented, there are many mysteries remaining. That said, this is a partial feature release.

Working (as far as my testing has shown):
- EMP Blast
The size of the blast increases with higher CR (CR 3 minimum to use)
- Orbital Strike
Size also differs between CR 4 and CR 5. I also filtered potential targets to ensure they are 'outside'.
- Reveal Friendlies
- Reveal Enemies

Not working:
~- Reveal Friendlies~
~- Reveal Enemies~
- Squad Leader Recall
~I tried to brute force the reveal ones by returning many combinations of values (map id/base id and testing on different zones & bases based on those values), but couldn't figure those out.~

### DIFF
--- a/server/src/main/resources/overrides/game_objects0.adb.lst
+++ b/server/src/main/resources/overrides/game_objects0.adb.lst
@@ -23,7 +23,7 @@ add_property boomer_trigger equiptime 500
 add_property chainblade equiptime 250
 add_property chainblade holstertime 250
 add_property colossus_flight requirement_award0 false
-add_property command_detonater allowed false
+add_property command_detonater allowed true
 add_property command_detonater equiptime 500
 add_property command_detonater holstertime 500
 add_property cycler equiptime 600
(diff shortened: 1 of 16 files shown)


Dethdeath:
EMP seems to work fine.

OS also works, but being able to launch multiple is probably not intended 😛 

https://github.com/user-attachments/assets/a44b3e62-4cdd-46e4-9486-f8223efad843

Dethdeath:
Multiple strike is fixed.

ScrawnyRonnie:
Reveal friendlies enemies have been added. Cooldown timers are set to their normal lengths as well.

ScrawnyRonnie:
Oh and the cooldowns should persist between sessions

PR #1283, opened 2025-07-19
No Safe Spaces

Fate-JH:
Still a WIP but this should show you what I am going for.

Currently, all zones are cleared for live fire activity.


Dethdeath:
Some stuff I tested (I assume most, if not all of it, is intended):
Firing weapons and killing players at sanctuaries is working.
HART can't be used by players from another empire, they are told the HART is not present.
Equipment terminals and other terminals can be destroyed.
Sanctuary respawn tubes can't be destroyed.
Sanctuary teleporters work for players from other empires.
Sanctuary towers can be captured by other empires.

captured

Fate-JH:
I don't remember if I tested this lol

PR #1284, opened 2025-07-21
Red Alert!

ScrawnyRonnie:
There was nothing more exciting than logging in to see a red alert at a base. This update brings those alerts back. For testing, I set the threshold of enemy players in an SOI very low:
1 - Yellow
2 - Orange
3 - Red

While testing, I found that PlayersInSOI did not include vehicle occupants so players in vehicles are now counted. This was preventing those occupants from receiving certain messages, such as generator under attack, so it serves a double purpose to include them now. 

I also ran into an issue with a player dying and respawning. While they are a backpack, they were not being counted so the alert level kept going down, then back up when they spawned. The client only gets a chat message for the higher level alert, but it was still sending quite often. To try to solve that (and prevent the DensityLevelUpdateMessage packet from constantly spamming), corpses are included in the condition before sending the update. This isn't foolproof because if a single player has two corpses on the ground, it no longer counts it. 

It will trigger the packet to send if the difference between the current enemy count and previous enemy count (last time TryUpdate happened - every 5-6 seconds) is at least 1. Some adjustments will need to be made (maybe timed delay between them as well), but I think this is a good start.

Also, these packets are sent during these events:
- When a player logs in, they get one for every base. This way if an alert is already active, they will see it straight away.
- When a base changes factions
- As population in the SOI fluctuates

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -508,6 +508,18 @@ game {
     capture-experience-points-modifier = 1f
     # Don't forget to pay back that debt.
   }
+
+  alert {
+    # When a certain number of enemy players are within the SOI of a facility, an alert (DensityLevelUpdateMessage)
+    # will be dispatched to all players. Players of the owning faction will receive a chat warning (if in
+    # the same zone) and the map will flash the alert level over the facility until it changes
+    # Wiki says 25-30
+    yellow = 1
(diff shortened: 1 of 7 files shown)


Dethdeath:
Everything is working 👍

PR #1285, opened 2025-07-22
Lukewarm Potato

Fate-JH:
Zero'd out the kill reward for an LLU carrier as per vanilla requirements.  This was always a placeholder.  The event should still be added to the database but the player who kills the transporter will no longer be rewarded CEP for it.

All that said, other factions, please continue to kill the enemy person making off with your LLU's.


Dethdeath:
Does what it's supposed to.

PR #1287, opened 2025-07-29
Animations and Turret Range

### LINKED ISSUE #393: Vehicles - Deconstruction - When a Vehicle deconstructs, it simply disappears, rather than going through the proper deconstruction animation
When a Vehicle deconstructs, it simply disappears, rather than going through the proper deconstruction animation.

### LINKED ISSUE #1057: Spawn Tube - Cannot deconstruct from repaired spawn tube
Spawn tubes that have been dead and then repaired cannot be used to deconstruct your character afterwards. Spawning itself works fine.

ScrawnyRonnie:
Should close #393 and fix #1057 and added a few other things:

Animations should play when corpse loot bags and vehicles despawn/disappear, a player zones, and when a player deconstructs at a tube/AMS. The deconstruct animation is wild. I do not remember it being like that 😆

Repairing a spawn tube should now allow you to deconstruct from it. DestroyMessage only breaks that for tubes? So... don't send that.

Turrets, specifically wall turrets (manned turrets) have extreme range. Some have mentioned how far the auto fire will continue shooting. There is an escape range set, but it seems to ignore it and keeps firing up to 400m. What I did is probably a band aid, but the other things I threw at it that seemed like they should make a turret stop shooting didn't work. There is also a 3-4 second delay in it stopping shooting once you've reached 150m away from it, so it keeps shooting a little beyond the range. Still, much less distance than it was before so escaping them is more possible.

Lastly, emotes will be seen by other players now. The classic /em cabbagepatch can be enjoyed by surrounding players.


Dethdeath:
-Repaired spawn tube issue is fixed.
-Deconstruction animations work on characters, vehicles and backpacks.
-Emotes work.

-GMs cannot see others doing emotes.
-GMs cannot see the character decon animation.

I also don't remember the tube deconstruct animation being like that, but I like it :)

PR #1288, opened 2025-07-31
Yet Another Squad Update

Fate-JH:
I don't even remember most of the changes I was working on.

Based on what it looks like, the following three things are the main attempts and one bug fix:

1. withholding messages and activity from being displayed until a sufficient expectation of in-game readiness is achieved
2. management of invite queue via a custom chat command
3. encapsulated the capabilities and properties of different squad join attempts (may be by squad leader, specifically)
4. squad search was broken somehow and I changed it

Do any of this work?  _Does any of this work?_


Fate-JH:
Yes, something's really wrong.  Not even gona sugarcoat it.

Issue #1176, opened 2024-03-01
Remaining Squad Issues


ScrawnyRonnie:
Listing some things that are still wonky with squads. 

**Disabled features:**
- Promoting a new SL. This has caused an error of squad leader not found. No known reproduction steps.
- Squad searching. Could cause an error by attempting to return a squad that no longer existed. (might be resolved by #1157 already, but still disabled).

**High;**
- Sometimes (for reasons tbd) a client will crash or be kicked to the character select screen when the player joins a squad.

**Medium:**
- Joining a squad while respawning or zoning will make squad UI elements never appear. Squad cards and numbers on map never show and the squad tab shows no squad members. Squad chat does work in this state and other players can see the player's card so the player is actually in the squad. Leaving and joining the squad while alive resolves it and UI elements appear.

**Low:**
- Squad cards remain on screen after a squad disbands or when you log back in if you were in a squad when you last logged out. Pressing Alt makes it go away and you are not actually still in the squad; just the UI stays for some reason.
- If player A is invited to multiple squads and A accepts player B's invite, player C that sent the other invite never gets an indication that player A joined another squad. If player A leaves the squad and player C tries to invite A again, A never receives the invite.

ScrawnyRonnie:
Just adding some of the cursed behavior that sometimes happens the moment you join a squad, view listed squads, click into a squad to view available positions, or click join to request. Each of these things has your client receive a SquadDetailDefinitionUpdateMessage packet, and the cursed one has a lot of extra random things in it causing its size to be much larger than the usual packet of this name.
![image](https://github.com/psforever/PSF-LoginServer/assets/73139382/96331e75-eecc-42b1-9d16-e5e0f745c7e5)
There has been some discussion around squad index (the squads themselves and positions within a squad). From this last test after a server restart, only one squad had been created since the restart (confirmed from logs) and the position shown as available was accurate across all clients and this bug occurred.

PR #1290, opened 2025-08-02
Let's Try Squads Again

ScrawnyRonnie:
Baby steps here... and I just noticed a println snuck in there 😨 but that part was why characters weren't getting the Membership packet responses. So squads are functional and players now see the squad UI and chat messages for invite actions. I have only tested a few things so far. One of them was having your invite accepted while you are respawning and the invite did go through after I spawned, so that queued/delayed response to the player is working properly (this was mentioned in #1176 and seems resolved - maybe test zoning while invite is accepted too?). Going to just PR it now so Deth and others can test.

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/ZoningOperations.scala
@@ -3853,7 +3853,8 @@ class ZoningOperations(
     }
 
     def startEnqueueSquadMessages: Boolean = {
-      sessionLogic.zoning.zoneReload && sessionLogic.zoning.spawn.setAvatar && player.isAlive
+      println(s"${sessionLogic.zoning.zoneReload} ${sessionLogic.zoning.spawn.setAvatar} ${player.isAlive}")
+      !sessionLogic.zoning.zoneReload && sessionLogic.zoning.spawn.setAvatar && player.isAlive
     }
 
     def enqueueNewActivity(newTasking: SpawnOperations.ActivityQueuedTask): Unit = {
--- a/src/main/scala/net/psforever/services/teamwork/SquadInvitationManager.scala
+++ b/src/main/scala/net/psforever/services/teamwork/SquadInvitationManager.scala
@@ -368,6 +368,7 @@ class SquadInvitationManager(subs: SquadSubscriptionEntity, parent: ActorRef) {
               charId,
               SquadResponse.Membership(SquadResponseType.Cancel, charId, Some(0L), name, unk5 = false)
             )
+          case None => ()
         }
         None
       }
--- a/src/main/scala/net/psforever/services/teamwork/SquadService.scala
+++ b/src/main/scala/net/psforever/services/teamwork/SquadService.scala
@@ -274,8 +274,8 @@ class SquadService extends Actor {
         SquadActionMembershipCancel(cancellingPlayer, tplayer)
 
       case SquadAction.Membership(SquadRequestType.Promote, _, _, _, _) => ()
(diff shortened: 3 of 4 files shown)


ScrawnyRonnie:
Double check me of course, but...
> -If you send out multiple invites at the same time and <invited player 2> accepts first, the squad leader gets notified that some invitations were cancelled due to that action, which is good feedback. However, the other players that were invited still have their yes/no window will see nothing happening if they click yes. The invites have to be accepted in the order they were sent out or only the first one that was accepted succeeds.
>
This should only happen if the squad doesn't exist yet. If the player inviting others has created or listed a squad already, the order they are accepted shouldn't matter. When the first invite is accepted and squad is created, it does drop any of the other invites as they will need to be resent. Not sure about how to inform the other invitee that accepts after and nothing happens.
> -If a squad member other than the leader sends out invites, players who accept these invites will fail to join the squad. The squad leader gets a message after this saying the invited player would like to join the squad and /accept does nothing here as the invite already failed.
>
Fixed
> -When a player declines a squad invite, they get 2 messages. One says they have declined the invitation, which is correct. The other message says "your request to join 'squadname' has been refused", which is incorrect.
> 
Fixed
> -When players attempt to join a squad via /invite squadmember or /invite squadleader, the squad leader gets a message saying this person would like to join and then to accept/reject it. If they type /accept however, the invite fails and the squadleader gets a message saying "your invitation to squadleader was accepted, but failed". This is a wrong message and probably also being sent to the wrong person.
> 
Fixed
> -If you toggle "location follows squad leader" off in squad options, the BEP waypoint stays at that location forever. Might not be implemented yet.
 -Selecting a role as the leader or assigning a squad member to a role isn't working but this probably hasn't been implemented yet.
> 
Haven't dug into this, but may not be implemented
> -GMs can receive squad invites, but if they accept one, nothing happens and any subsequent invites are not received even if they toggle GM mode off.
>
Fixed

Dethdeath:
> > -If you send out multiple invites at the same time and <invited player 2> accepts first, the squad leader gets notified that some invitations were cancelled due to that action, which is good feedback. However, the other players that were invited still have their yes/no window will see nothing happening if they click yes. The invites have to be accepted in the order they were sent out or only the first one that was accepted succeeds.
> 
> This should only happen if the squad doesn't exist yet. If the player inviting others has created or listed a squad already, the order they are accepted shouldn't matter. When the first invite is accepted and squad is created, it does drop any of the other invites as they will need to be resent. Not sure about how to inform the other invitee that accepts after and nothing happens.

Yep can confirm, only happens when the squad doesn't exist yet.

I got sent to the character select screen when I looked at "Roles" as a squad member. Tried it on another character later on and the client crashed:

https://github.com/user-attachments/assets/8774e77a-b30c-4a73-ae4a-ec0942a6a2d9

Also saw a bunch of outfit garbage text appear when clicking Roles here:

https://github.com/user-attachments/assets/36a41923-1731-4360-b1f1-73e79db5b1ee

Another character got this:

OutfitG

ScrawnyRonnie:
I unfortunately had a few crashes and also the 'You have logged off' message upon joining a squad as the 10th member. I did packet captures...

ScrawnyRonnie:
Testing a change with how `SquadDetailDefinitionUpdateMessage` is delivered. Hopefully this can stop the random client issues that can sometimes/rarely occur when this packet is delivered. Haven't thoroughly tested, but initial packet captures show the expected behavior.

Squad chat should also be working. It was too easy that it feels like I had to have done something wrong. Was it supposed to be this easy? 🙃

PR #1295, opened 2025-08-10
Issue Closer

### LINKED ISSUE #1278: Bailing without bailing
When you're driving an open cockpit vehicle and press the dismount button the vehicle stops instantly and lets you dismount. Bailing from these types of vehicles should be possible, but dismounting like this should not be possible whilst driving. Also works for gunners/passengers.

https://github.com/user-attachments/assets/cb7d8872-0e5d-4e6f-9ae6-89d04c21e9ec

Repro:
-Drive an open cockpit vehicle.
-Press G as either the driver or gunner/passenger.

### LINKED ISSUE #1292: Wrong message when attempting to bail in a WG as a gunner/passenger
Gunner and passenger in closed vehicles inside a warpgate get the message shown below:

https://github.com/user-attachments/assets/1209267d-2772-4bde-b973-70592a9a077c

These are not pilots 😛

### LINKED ISSUE #1293: Hack goes through even when link is gray
This happened during a recent fight and should not have.

When 2 bases that are linked to each other get hacked at the same time, the last one to complete should fail if it is the only link they have to the base. See video:

https://github.com/user-attachments/assets/6cabd20c-4219-4dee-b066-2696e01563c3

The map perspective is from a 3rd soldier that is on NC. You can see the link to Neti turn gray and the hack completes anyway.

### LINKED ISSUE #1294: AMS can be deployed inside Warpgates and Geowarps
AMSes can be deployed inside active warpgates and geowarps, shouldn't be possible.

ScrawnyRonnie:
Should close #1294 close #1293 close #1292 close #1278

If both bases are hacked within a second of each other they both might still finish. Apart from that, it should interrupt the hack while it is in progress.

Also enabled squad promoting new leader. I didn't have any issues with it so far so maybe that is fixed?

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
+++ b/src/main/scala/net/psforever/actors/session/support/SessionAvatarHandlers.scala
@@ -56,14 +56,14 @@ class SessionAvatarHandlers(
   }
 
   def facilityCaptureRewards(buildingId: Int, zoneNumber: Int, cep: Long): Unit = {
-    //TODO squad services deactivated, participation trophy rewards for now - 11-20-2023
     //must be in a squad to earn experience
+    val delay = continent.Building(buildingId).map(building => building.CaptureTerminal.get.Definition.FacilityHackTime.toSeconds / 6)
     val charId = player.CharId
-    /*val squadUI = sessionLogic.squad.squadUI
+    val squadUI = sessionLogic.squad.squadUI
     val participation = continent
       .Building(buildingId)
       .map { building =>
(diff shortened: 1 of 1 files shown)


Dethdeath:
-AMS deployment in WGs and Geowarps fixed.
-Base hack at the same time issue fixed.
-Dismount in moving open vehicles is fixed.
-Wrong message when bailing as a gunner/passenger in a WG is fixed.
-Squad leader promotion works.

Still getting the wrong message when bailing in a WG as a GM.

PR #1301, opened 2025-08-26
Update GitHub Action to fix broken pipeline

Resaec:
Pipeline update, connected and spawned in sanc.
Test still fails to run :/


Resaec:
Other updates will come after this is merged.

PR #1306, opened 2025-09-11
A Few Fixes and Tweaks (CCs and Outfits)

ScrawnyRonnie:
Should resolve a few issues:
- Hacking a tower CC
- Resecuring a base if there are no neighboring friendly bases
- Outfit member last login time. Saw some errors in the logs `Could not encode OutfitMemberEvent` because the value for seconds was negative if the last login time was very recent.
- Buffed SL CEP for a capture event.


Dethdeath:
Successfully hacked a few towers and did a resecure with no friendly bases around.

PR #1310, opened 2025-09-23
No Stamina No Life

Fate-JH:
How long has it been since you properly had no stamina after you died?  October 2020 by the looks of it!

It looks like implant re-activation might have been affected, the backend of this process especially.


Dethdeath:
AA MAX damage also fixed.

PR #1311, opened 2025-10-06
When Squad Leader Leaves...

ScrawnyRonnie:
Don't disband the squad anymore. Should apply if SL uses /leave or logs off.

Clear skies are also in the forecast. Persistent heavy precipitation was visually and audibly impairing as some have commented so we can play with that again when storms can move. Attempts to decease the intensity did not work. It seemed to require both a minimum size and intensity for it to fully do weather and not just be dark clouds above (100 might be the minimum for each to make it actually rain/snow).


Dethdeath:
All working fine.

PR #1313, opened 2025-10-28
Mr. Ditkovich gets his rent

ScrawnyRonnie:
door

The doors are fixed...

Hurakan on Hossin had two door IFFs that were swapped. Top lobby and on roof
image

Also disabled a few items that are not yet functional or are currently broken


ScrawnyRonnie:
I want to say the scorpion and maelstrom's 2nd/3rd fire modes both stopped working for the same reason, so you may be able to test that weapon. I can undo disabling those if that would follow soon. I've seen the mael being used recently and the poor guy didn't know it was doing no damage.

PR #1321, opened 2025-11-30
CSR Dismount

### LINKED ISSUE #1298: GM related vehicle lag
Whenever a GM is in a vehicle and gets the message "You are driving too fast to dismount" for any reason, it gets spammed a huge amount. This is different from non-GMs as they only receive such messages once. The effects of the spam the GM receives affects the entire server and causes disconnect screen popups and ping spikes to 3000+ for everyone online. The vehicle they are in is also in some catch-up mode where it takes several minutes to catch up to actions like players entering/exiting the vehicle or driver input.
The server lag lasts for quite a while and even causes login errors.

Some footage:

https://github.com/user-attachments/assets/1adf1956-c485-4686-8d32-16730f178f7d

https://github.com/user-attachments/assets/a3e584d6-7522-408d-ae71-67157750c1af

Fate-JH:
CSR players were failing to bail from moving vehicles and that resulted in breaking the dismount routines, causing lag, spamming event chat, and even crashing the client.  The policy was always that CSR players have priority to dismount/eject/bail from vehicles and that's why this fix was composed.

Exhaustive conditions for the bail system failing and for avoiding situations where the player should not be allowed to bail from the vehicle based on how long since it had been spawned, even if they were working correctly, made narrowing cases an minor issue.  I condensed five cases into two.  Someone more familiar with the bail failure code and the code that stopped people from bailing too early should make certain the logic for that holds up to your expectations.

Literal instructions:
- be CSR-level player
- spawn vehicle
- get in vehicle (automatic)
- drive vehicle forward (once control has been restored)
- dismount (bail or otherwise)

Expectation
- player should get out of vehicle

### DIFF
--- a/src/main/scala/net/psforever/actors/session/csr/MountHandlerLogic.scala
+++ b/src/main/scala/net/psforever/actors/session/csr/MountHandlerLogic.scala
@@ -40,7 +40,39 @@ class MountHandlerLogic(val ops: SessionMountHandlers, implicit val context: Act
 
   def handleDismountVehicle(pkt: DismountVehicleMsg): Unit = {
     //can't do this if we're not in vehicle, so also not csr spectator
-    ops.handleDismountVehicle(pkt.copy(bailType = BailType.Bailed))
+    val DismountVehicleMsg(player_guid, _, _) = pkt
+    player
+      .avatar
+      .vehicle
+      .flatMap { obj_guid =>
+        (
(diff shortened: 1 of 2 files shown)


Fate-JH:
Was it working before my change here?  I assume, given all these different cases blocking exiting the vehicle too quick after spawn, that it was not, but I haven't checked the blame file to compare actual dates for what I wrote and when those were written.

Fate-JH:
The current situtation of this branch is difficult to recover.  Going to redo from scratch if the original problem still exists.

PR #1327, opened 2025-12-05
LLU Handling / Wall Turret to MAX autofire conditions

### LINKED ISSUE #1324: LLU lost to warpgate bubble when in vehicle
If you're on foot and attempt to walk into a warpgate whilst carrying an LLU, you will be correctly bounced out of the warpgate or prevented from entering in the first place. This always leaves the LLU intact.
If you are a passenger of a vehicle and the vehicle gets too close to a warpgate, the LLU disappears and the hack fails.

### LINKED ISSUE #1325: Wall Turrets fire at falling MAXes
See video:

https://github.com/user-attachments/assets/df1d5bae-b07a-47fd-a175-9b32b32b8c32

They even continue firing after the MAX has landed.

ScrawnyRonnie:
They continue to have more edge cases that haven't been addressed yet. 

- This should fix #1324
- Disallow LLU carrier to mount a vehicle in a warpgate to prevent them from entering it in that manner
- Not allow LLU carrier to use recall, instant action, or quit without first dropping it
- If a player is in the middle of one of the above zoning actions and picks up the LLU, it will cancel that zoning process.
- Fixed some swapped values that were being written to the database for facility and zone ids

Some of these were to prevent the LLU from being warped to 0 0 0 because the carrier left the zone.

Also added a log when a contested capture event happens. There is still a small chance of players not receiving a reward when kills happened during the hack or resecure. Hopefully this can shed some light on that rare occurrence.

Edit:
- Should fix #1325
- Wall turrets should only auto fire on a MAX if the max is in auto run mode (which sends packets when enabled/disabled), but aggro also continues if cover is not found after ceasing auto run

### DIFF
--- a/src/main/scala/net/psforever/actors/session/csr/GeneralLogic.scala
+++ b/src/main/scala/net/psforever/actors/session/csr/GeneralLogic.scala
@@ -524,6 +524,10 @@ class GeneralLogic(val ops: GeneralOperations, implicit val context: ActorContex
           if (avatar.lookingForSquad) {
             avatarActor ! AvatarActor.SetLookingForSquad(false)
           }
+        case GenericAction.MaxEnableAutoRun =>
+          player.maxAutoRunEnabled = true
+        case GenericAction.MaxDisableAutoRun =>
+          player.maxAutoRunEnabled = false
         case _ =>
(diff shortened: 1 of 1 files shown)


Dethdeath:
Wall turrets no longer fire at falling MAXes.

PR #1328, opened 2025-12-05
Lead Plating

Fate-JH:
Radiators, prior to being deactivated, were incorrectly applying damage to targets in the sense that too much damage was being applied to targets that should have been shielded from radiation.  Vehicles and mechanized assault exo-suits, in particular, were being incorrectly affected, but so would have mountable turrets and the implant machines been left unshielded.  By adjusting resistance lookup, entity references, and clarifying how targets and projectiles were sorted, the amount performed by damage should be correct now.  Any single type of radiation damage encountered from multiple sources simultaneously should be applied only once per tick.

Previously, projectiles that generated secondary proximity damage radiation clouds but struck a functional target would only induce direct hit damage.  A good example of this working correctly is the OICW (laughs out loud at the notion of the `oicw` working correctly) which does not spawn its secondary projectiles - little buddies - if they stike a target before traveling a set distance.  The Radiator and other such weapon systems do things differently where directly striking a valid target generates the radiation cloud, the same as if the projectile strikes a surface that does not belong to a valid target, on top of whatever direct damage it causes.  The Radiator, furthermore, uniquely differentiates how it affects the target between its bounce projectile hitting certain targets and its on-contact projectile hitting those same targets.  This should be experienced properly now.

___Features___
__RadiationCloudInteraction__
A new parent class for all entity zone interactions that involve radiation cloud detection.

___Caveats___
1. Though the amount of damage per tick has been tuned, the pace of ticks is not necessarily perfect.  I am adapting a method of damage aggravation that operates on a different timing system to fill in the gaps in places where damage would have been felt.  "Damage being too quick" is a legitimate concern that I will address if escalated.
2. Though vehicles now provide radiation damage shielding, not all vehicles provide radiation damage shielding.  Going by just the values in the ADB file, the following vehicle categories do not shield against radiation: ATV's (Basilisk, Fury, Wraith) and medium transports (Deliver, Raider, Thunderer, Aurora).  The TR Marauder, unique to all other buggies, does not offer any relief from radiation.  The Wasp also does not offer any relief from radiation.  There's no rhyme or reason to this distinction.
3. The droppod and HART shuttle provide perfect radiation shielding.  This is not vanilla but is being offered as a courtesy.  Both operate on a timed release method that would cause anyone affected by a radiation cloud to suffer unavoidable radiation damage through no fault of their own, possibly leading to death before return of control.


Dethdeath:
Not sure what to make of this. I thought it had to do with the vehicle shield at first, so I tested that. 2 ANTs, one with a vehicle shield, one without. Made no difference. The difference in damage came from getting out of the vehicles and getting back in. 
So basically if you purchase a vehicle, get out and back in, you will take more damage from the radiator bubble. The extra damage varies a bit and health can end up anywhere between about 25-5hp left.

(In 2 parts and some pieces cut out to reduce filesize)

https://github.com/user-attachments/assets/3bcb9c4c-846b-4865-bcb7-7b7c1ce49a0d

https://github.com/user-attachments/assets/2988fa70-f846-40ac-80f1-c9864e1041d0

Dethdeath:
Issue is fixed now.

PR #1335, opened 2025-12-16
EmpireBenefitsMessage packet

Resaec:
This PR adds de/encode support for the EmpireBenefitsMessage packet.

There is only one test yet to allow for rapid development on top of this packet.
More samples will be tested against to ensure correct behavior.


ScrawnyRonnie:
I sent that exact packet on login to see what it would do. New pretty messages in chat:
image
And the global map benefits window:
image

ScrawnyRonnie:
All's well after merge and commits

PR #1342, opened 2025-12-30
Pool's Closed

Fate-JH:
It should no longer be possible to drown indoors in the basements of facilities and towers.  As a result, it's no longer necessary to carve oddly-shaped water pool maps on certain continents, specifically Hossin, Ceryshen, and Ascension.  Forseral will remain a hassle because various parts of the outdoors terrain - especially the dips in the fields north of Eadon - are below sea level.

I resolved an issue with the player's oxygen meter not being displayed while recovering because the recovery time calculated to zero seconds (datatype conversion issue) causing the mechanic to completely reset instead of slowly revert.  Vehicle waterlog recovery seemed to work correctly when testing.

### DIFF
--- a/src/main/scala/net/psforever/objects/serverobject/environment/EnvironmentAttribute.scala
+++ b/src/main/scala/net/psforever/objects/serverobject/environment/EnvironmentAttribute.scala
@@ -1,6 +1,7 @@
 // Copyright (c) 2020-2024 PSForever
 package net.psforever.objects.serverobject.environment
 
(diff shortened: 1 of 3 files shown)


ScrawnyRonnie:
Recently lost an LLU in Heyoka vehicle bay due to water. Appears to be safe now
image

Fate-JH:
Eh?  Heyoka's garage actually is flooded?  I need to check this.

PR #1350, opened 2026-01-19
Capitol Force Dome

### LINKED ISSUE #1291: Capitol Force Dome remaining issues
-Doesn't kill enemies within range when it comes up.

-If the force dome generator kills a player they have no way to respawn:

https://github.com/user-attachments/assets/f4cc24bd-60cf-444c-b83f-59ba902d32f2

Fate-JH:
> Big push, friendly base, hacking, resistance
> "Hurry expand the force dome"
> Big push, Capitol base, timer's insistence
> Dread under the force dome
> 
> All those nearby bases that were taken
> Were leading to "one last" fight
> Hack over and wall immediately energizing
> In a facility betrayal, we all died
> 
> Someday, anticipate the same perspective as defenders swarm
> The subcapitols once again captured and klaxons be alarmed
> It bewilders high command to think this capture console ever compromise
> But, Capitol force dome, we will make you deliver our enemies' demise *

The capitol force dome is a feature of capitol facilities on six of the game's major overworld continents, triggered when one factions owns at least three facilities - a certain central facility with the force dome generator and two facilities comnnected to that one across a continental lattice.  Calling a "dome" is a bit generous.  The actual shape is much better described like an multi-level irregular pyramid where the lowest angles are shallow enough to be treated as walls.  The dome generates from a telescopic tower installed onto the roof of the protected capitol facility, originating from its bulb cap.  When activated, it extends high into the air and then the protective field awkwardly cascades out and downwards, covering the entire of the facility from wall to wall and even enveloping some bunkers adjacent to some entrances.  Enemy players may not pass through this field making all of the entities underneath it virtually undefeatable.

Other functionality exists but had previously not been handled properly.  One effect that did exist before but was not being handled correctly was an instance of immediate death.  Touching the bulb on top of the telescoping tower caused a fatality so instant and decisive that the client didn't wait for the server to decide it. The player was just liquidated locally and the server had to be told of the fact.  Without the server being ready to catch this state, the client was left in a state of unresponsiveness where nothing was possible.  Resolving this improper state, though it was rare to be encountered, was one of the first matters to be resolved and allowed for me to anaylze the actual conditions of the capitol force dome tied down to an in-game object that exists on the server.  Three objects of note are represented.  One is the specific variety of force domes that exist based on the type of capitol facility it protects; these ones exist.  Another is a global force dome definition object for positioning data specific to different types of domes; these ones doesn't exist, but are used.  The third is a dome generator that provides additional positioning information; this one does not exist.  The instant death is caused by reported as collision with the former-most specific variety objects.

When the facility or one of the facility's connecting facilities changes faction alignment, a message is dispatched to all of its installed amenities informing of the event.  The force dome checks whether the local lattice is in any condition to make the capitol facility become protected. When the force dome expands, the thing it primarily is known for doing and the thing it is now capable of doing is causing certain demise.  Infantry, vehicles, and combat deployables that find themselves underneath it - technically anywhere within the walls of the irregular pyramid energy field - are killed outright.  This only happens once, when the dome first envelops the facility; and, following that, a sense of security is placed unto the base.  Certain elements like the capture console become unhackable.  All automatic amenity repairs are treated as free (but still take as long as normal to complete).  Any player or vehicle that can find its way into a facility, even if it were aligned to an enemy force, becomes unkillable, indestructible, for the time it spends under the force dome.

Be mindeful of your abandonned Router telepads.

___Features___
__``GlobalDefinitions``__, __``GlobalDefinitionsMiscellaneous``__
Definitions for all of the different types of capitol force domes in the game, separated by facility of installation.  With five overworld major facilities, and six zones with capitol facilities, only one type of capitol force dome is used more than once.

__``ForceDomeDefinition``__
Keeps track of the bounded energy field offsest for the fore dome perimeter and which entities are protected by the resulting field (as long as they are also installed in the facility).

__``ForceDomePhysics``__, __``ForceDomeControl``__
Entity representing a capitol force dome entity.  Keeps track of whether the dome is active (visually energized) and whether it respresents a bounded region that influences entities beneath it (its perimeter).  If the dome state energizes (activates) naturally, it also controls the enemy forces purge within the said perimeter.  (Note: "force dome physics"

__``ForceDomeExposure``__
The causation for any damage from the capitol force dome, both direct contact and purges.  Since the force dome represents an exotic variety of affliction, all deaths will count as "collisions".

__``ResourceSiloControl``__
Introudced a message stcuture that modifies the rate at which NTU is drained per repair request.  The same amount of repair is still performed per request.

__``MajorFacilityLogic``__
Removed capitol force dome management logic for use in ``ForceDomePhysics``, but included conditions for changing NTU drain and amenity vulnerability due to a detected change in the capitol force dome.

__``Zones``__
Constructs capitol force domes and assigns them as amenities to some facility ("the capitol").

__``InteractWithForceDomeProtection``__, __``InteractWithForceDomeProtectionSeatedInEntity``__, __``InteractWithForceDomeProtectionSeatedInVehicle``__
Zone interactions for including dynamic entities that can mount players under the protection plan of the capitol force dome.  The vehicle interaction passes the baton off to any vehicle that might be ferried in the vehicle cargo holds, if it is a carrier-type.  Please refer to caveat 2.

__``Damgeable``__
Includes a flag that causes the target control agent to reject all messaging in regards to sustaining damage.  This is dangerous messaging and should not be dispatched flippantly.

___Commands___
The following CSR-level commands allow manipulation of the capitol force dome without requiring conditions to support the capitol force dome's current state.  Normal changes to the capitol or to the subcapitols will not affect the force dome, unless specified.  In all instances, a capitol force dome must be available in the same zone as the CSR-level user; that user does not need to be within the sphere of influence of the capitol facility.  If there is no capitol facility in the current zone, nothing will happen, but a message will indicate the lack of force dome to operate upon.
``!forcedome``, ``!forcedome o``, ``!forcedome on``
Enables a capitol force dome.  The force dome will appear as expanded around the capitol facility regardless of the conditions with the subcapitol facilities.  Protections gauranteed by the force dome were not established.  No hostile forces purge occurred.
``!forcedome of``, ``!forcedome off``
Disables a capitol force dome.  The force dome will not be expanded around the capitol facility even if the conditions with the subcapitol facilities suggest that it would be.  Protections gauranteed by the force dome were not revoked.  No hostile forces purge occurred.
``!forcedome protect``
Extends the protections of a force dome to the affected amenities, allied forces, and hostile forces within its perimeter.  The capitol force dome was neither expanded nor disabled.  No hostile forces purge occurred.
``!forcedome purge``
Causes a hostile forces purge within the perimeter of a capitol force dome at that moment.  Applied protections that are already in place will, of course, interfere with the purge action.  The capitol force dome was neither expanded nor disabled.
``!forcedome normal``
Restores the operations of a force dome to apply to the normal conditions rather than any command-determined state as described above.  The capitol force dome may or may not enable or disable.  The protection state of various capitol amenities, allied forces, or hostile forces within its perimieter may be extended or may be revoked.  Regardless of what other changes are made ore skipped, a hostile forces purge will be triggered.

___Caveats___
1. Capitol force domes are not normally counted as amenities and are therefore never owned by a facility.  They are not the reason why a facility is promoted as that continent's "capitol facility" either (there's another flag for that).  Internally, these are separate mechanisms: that which designated the capitol facility, and the region maintained by the force dome.  For our purposes, it's easier to link the force dome to the facility since existing systems such as power, the facility generator, hack state of the subcapitols, and such, are all managed through messaging between amenties and their installed facility and that will properly reflect the state of the dome.
2. The force dome extends damage protection against certain entities that are physically located underneath it: the facility generator and most importantly players.  The generator is an explicit and simple case.  Players create a more complicated issue.  To reduce opportunities for death, anything a player character can be mounted in must also be protected such as vehicles, the implant terminal machines, and the turrets, both facility-installed and deployed.  These latter entities are not counted as officially protected from the force dome and are piggybacking off of the logic of an adjacent entity "being protected". It's not vanilla acording to any of our official sources.
3. The capitol force dome generator - the expanding stalk with a bulb of instant death on top of it - does not exist.  This is a vague statement because it can be seen and interacted with.  While it has an object id defined by the game (332), and you can crash your Mosquito into it, and can most certainly die from contact with it, it never gets instantiated as an actual entity in the game world.  It has no globally unique identifier required by any known interaction with the game distinct from the field physics itself, whether silently as certain terminals attached to the facility or specifically as an entity in the zone map files.  For our purposes, it doesn't need to exist.

___Addenda___
1. All turret upgrade variables were being stored and referenced at a global level on `GenericHackables`.  Horrifically, this means all turrets across the intergalatic cluster of zones would be blocked from being mounted while at least one of them was in the process of being upgraded; and, if even two of them were being upgraded simultaneously, there's a chance that both (all) of them would break from the concurrent strain.
2. Turret moutning and dismounting was not previously represented as a historical event.  I am uncertain if this was important in regards to experience tracking but a distinction to vehicle sourced events was included in cases where transportation was to be rewarded.
3. Viral infections were introduced during the making of this feature branch and most certainly lacks proper engagement with the capitol force dome mechanics, e.g., where the NTU drain is reduced (to zero).  If any interaction between the capitol force dome and viral infection exists, it needs to be researched and integrated properly in a different feature branch.
4. I reviewed and rewired at least half of the vehicle/turret mounting code in pursuit of an operational zone interaction chain of responsibility,  I adding case handling statements to differentiate between gunner seats and passenger seats, moved mount blocking cases into their own mix-ins, referenced normal mounting protocol where custom was used, and then introduced an easy to override callback method to make the process more responsive to the specific implementation.
5. Following from the above addenda point, I also reviewed all of the upstream packets to ensure that at least and at most one of them was handling the cone interaction trigger.  Vehicles have switched from reporting on their own zone interactions whenever there's no driver to whenever there's no occupants of the vehicle at all.  This retains the ease of integration with the driver's upstream and extends it to the occupant closest to the driver seat, reducing the load per vehicle independent of normal client-server messaging for the player.  If it weren't for HE mines and automatic turrets, it would probably not be necessary to continue to test zone interactions for completely vacated vehicles.
6. I do not remember when I started writing parodic lyrics as the opener of major PR's.  The ones I do recall the fondest are the riff on The Daring Young Man on the Flying Trapeze (George Leybourne) in the PR when ziplines started working (unknown) and 'Disco Inferno' (The Tramps) for the PR that introduced #529.  Only now, do I realize that I never properly credit the original and not everyone might know the meter to follow.  Going forward, the last addenda will be the credits is such a riff is included.  Having said that ...
7. * Sung to the tune of Bus Stop - The Hollies (1966)


Dethdeath:
Force Dome commands seem to work.

-Dying to the bulb still gives no way to respawn, both with infantry and vehicles.
-Players remain invulnerable after the force dome goes back down, but only in the area where the force field used to be, so not the entire zone.

Dethdeath:
I destroyed the spawn tubes and lobby term at a base where the force dome was up. The force dome came down and they were not being repaired by the base. Other equipment terminals that were destroyed afterwards did get repaired by the base with the force dome still in the same state of being down.

Fate-JH:
The force dome gives protection ands the force domes takes it away.  If you play with the CSR commands, remember to cap them with `!forcedome normal`.

All players should properly die now if they touch the bulb of death.  Normal player death and spectators death are permanent (this life is over), but CSR players should revive in a moment.

Dethdeath:
-Dying to the bulb now works correctly. Regular players get a respawn countdown, GMs get stuck in a die-respawn loop, but they can just /warp away to get out of that.
-Players no longer invulnerable after the force dome comes down.

Fate-JH:
I realized that landing dead on top of it should trap you into an infinite death loop when as CSR but didn't have many good options on what to do.  The kind of death state the game puts you in traps your player character in place until the client thinks you are alive in which case you will promptly die again, leaving you with little to no opportunity to be moved.  It's good that `/warp` is available.

Dethdeath:
Forgot I took this screenshot the other day.

AMS and enemy player were out of range of the force dome shield, but got killed when it went up. I suppose the killzone is just a giant rectangle, so makes sense why they died. Maybe it can be adjusted a bit?

Anguta

Fate-JH:
No, it's actually that polygonal outline that is recognized as the force dome.  It's entirely possible the center or rotation is wrong. That's Anguta Ceryshen?

Dethdeath:
Yes it is.

Fate-JH:
This should resolve the weird kill radius upon the force dome purging targets.  Mesy refactoring after a change was to blame.

I should correct some things I stated above about the force dome generator: we keep track of it now as I use its height to assist in force dome ceiling inclusion approximations.

PR #1355, opened 2026-04-06
Initial implementation of VR Training

Subsonic154:
This is the basis of implementing the VR Training zones to the server.
With handling for TrainingZoneMessage, the transfer to VR Training zones from the Virtual Training buildings in the Sanctuaries now works.

As for the zones themselves...

What works:
- All terminals in the VR Shooting Range zones work*.
- All terminals and the Repair/Rearm Silo in the VR Driving Area zones work*.
- Players and vehicles are invulnerable to friendly fire and have access to all equipment without cert/module requirements or purchase timers in the VR zones (not sure if lack of purchase timers in VR is accurate to live, but it's nice QoL) .
- The players loadout is reset to the default respawn loadout upon exiting VR Training to prevent them from smuggling equipment out.
- The Shooting/Driving zones no longer display "The data on your client does not match the data on the server. Please patch or update."

What doesn't work:
- The targets in the VR Shooting Range are not implemented yet.
- The zones have no proper spawn points, triggering the respawn in Sanctuary fail-safe if the player somehow dies despite being invulnerable (/suicide or map edge kill plane), I am unsure how live handled respawning in these zones, so I just left it at triggering the fail-safe for now.
- The logic to display the @TrainingGriefWarning message is implemented, but the function logic itself is commented out because I haven't found the correct ChatMessageType for it yet, it's supposed to use one that causes the pop-up to have a 5 second timer before you can close it.
- No work has been done on the VR Combat Area zones yet, as they are unused/inaccessible without /zone.

Additional technical details:
- Adding handling for new objects; spawn_pad (id 800), spawn_zone (id 815), vr_training (id 955), vt_air_vehicle (id 976). spawn_pad and spawn_zone use a heavily stripped back class/constructor since their logic is handled client-side, this is just enough of a representation for a GUID check against them to not cause the server to send an ObjectDeleteMessage to the client when they are hit; this also fixes the teleporters in the Sanctuaries deleting themselves when hit because they use the spawn_zone object type.
- Added a set of shared property overrides for all VR zones, and modified various checks/conditions to behave differently when the player/object is in a VR zone.
- *These zones have a unique map wrapping behavior that live seemed to use to create multiple "instances" of sorts of the zones to keep them from getting too crowded. This behavior is not implemented by the server yet, it is possible to reach the other instances by using GM mode to get past the map edge kill planes, but no objects in those instances will function because I haven't created data for their zonemap representations yet.

I would've liked to have had the Shooting Range targets working before making this PR, but that is going to be a massive undertaking for me that could end up taking almost as long as the rest of the work on it's own, plus I've been busy with other projects... But I figured in the meantime it would be useful to go ahead and try to get this out there for people to utilize. Being able to test all equipment without the certs in a safe environment is useful even without targets to shoot.


Subsonic154:
I've been informed of an inaccuracy in the invulnerability behavior, players and vehicles are only supposed to be invulnerable to friendly fire in VR, not all damage. Please do not merge this until I refactor that behavior.

Subsonic154:
I've made the necessary changes and the PR is now ready for review again.

PR #1357, opened 2026-04-12
TimeOfDayMessage implementation

Resaec:
corrected packet decoding
added helper functions
added zone logic to store and access local time
updated zoning logic to use local zone time
added chat commands to change zone local time and time speed


Dethdeath:
Commands work.

PR #1359, opened 2026-04-16
Workaround world connection issues caused by routers that lack Hairpin NAT support

Subsonic154:
This acts as a built-in workaround to world connection issues caused by this network configuration (which usually can't be fixed by the user without getting a new router, only worked around). I ran into this when I started having other people connect to my dev server for play-testing changes and figured I'd do something about it.

What this does is use separate world server forwarding logic depending on if the connection is coming from the local network or not, if the connection is from outside of the local network, it _should_ behave the same as before, I've tested this extensively and nothing appears to be broken. If the connection _is_ from the local network it uses different logic, localhost connections just get forwarded back to localhost, ignoring the public address, connections from other machines on the local network get forwarded to a new local address parameter in application.conf instead.

Adding a local/private address parameter to the config is not how I would've _liked_ to have implemented this fix for other machines on the local network, but it's the only solution I could come up with that was cross-OS compatible without system file tweaks, and I wanted to keep setup simple.

If there's a better way I could've handled any of the changes here please reach out.


Subsonic154:
Connections seem to be working fine on the test server instance this PR spun up, if that's a close enough representation of a container environment to know this doesn't break that, let me know and I'll mark this ready for merge.

Subsonic154:
I'm fairly confident this is safe now, so marking as ready for review.

PR #1361, opened 2026-04-19
Battleframe Destruction

Fate-JH:
Based on analysis reported by @jgillich, the destruction of battleframe robotic vehicles does not require repeated use of unconfirmed explosion-causing packets and the actual vehicle should be destroyed on each client before the final explosion to cause characteristic jittering.

Battleframe robotics can be pulled from appropriate terminals again but all previous restrictions and limitations are still in effect - excluded subsystems, siphons restricted, no remote projectile weapons, and such - save that flight variants have been restored.

Crashes to follow.


Dethdeath:
Jittering works.

PR #1362, opened 2026-04-19
Vehicle Destruction

Fate-JH:
In the client, a variety of destroyed models of some vehicles exist; and, where possible, these alternate models were utilized to represent the vehicle when it was destroyed.  They appeared after zone loading, usually, by segregating those particular vehicle definitions where it was defined.

Recently, @jgillich has discovered that the Reaver wreckage was causing issues during player interaction (what of it was possible) and during zone loading.  This alternate model wreckage was disabled across the board.  As far as I recall, no vehicle required its alternate wreckage model and I only included this as an attempt to reduce packet size.


Dethdeath:
Crash involving the reaver wreck is gone.

PR #1363, opened 2026-04-28
This Side

Fate-JH:
After discovering that people were no longer drowning on relog until they passed through a strict outside setting trigger, e.g, an external door, I attempted to concoct a variety of clever tricks to approximate whether a player might be considered inside or outside at any given time, from any given coordinates in a zone.  It didn't have to be perfect perfect, but it did have to work.

It didn't work.

So I just added it to the database for relogging purposes and called it a day.

___Commands___
Players engaged in customer service representative mode (CSR) can not set their sidedness state by interacting with a sidedness determination trigger and, instead, the last valid state is retained between calls to `gmtoggle` for the most part.  Players engaged in spectator activity have the similar issues but they are designed to respawn if they leave the observation mode.
CSR-level players can set their sidedness to `InsideOf` with the commands: `!setside i`, `!setside in` or `!setside inside`.
CSR-level players can set their sidedness to `OutsideOf` with the commands: `!setside o`, `!setside out` or `!setside outside`.
CSR-level players can check their current sidedness state with the command: `!setside check`.

___Caveats___
If you are already in an incorrect state and start in a way that does not involve respawning, the initial login will not put you into a correct state.  It's not designed to check that, only that it sets it properly moving forward.  If you check the SQL, the default value is `false` which is treated as `OutSideOf`.  Please interact with a sidedness determination trigger, e.g., an external door, or respawn, before assuming whether it works correctly or not.

### DIFF
--- a/server/src/main/resources/db/migration/V017__Sidedness.sql
+++ b/server/src/main/resources/db/migration/V017__Sidedness.sql
@@ -1,0 +1,3 @@
+/* New */
+ALTER TABLE savedplayer
+    ADD COLUMN sidedness BOOLEAN NOT NULL DEFAULT FALSE;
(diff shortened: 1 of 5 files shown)


Dethdeath:
Drowning issue fixed.

PR #1368, opened 2026-05-05
O I C Which Side

### LINKED ISSUE #1351: Scorpion damages players inside
The Scorpion can be used to damage players and amenities inside buildings. This is obviously because the server still has no concept of where walls are, hopefully this can be resolved with a sidedness check.

Fate-JH:
Exploding Scorpion (`oicw`) little buddy sub-projectiles are wrapped in special entites because the little buddy itself doesn't have any vitality information and certain aspects of the damage system account chain requires vitality for that particular element.  This wrapper element, however, was not properly transmitting sidedness in regards to the little buddy (even though the little buddy's originating projectile underwent special sidedness determination and attempted to pass that onto the little buddy).

As is my policy with all matters regarding use of the `oicw`, it's up to someone else to prove it works.

### DIFF
--- a/server/src/main/resources/overrides/game_objects0.adb.lst
+++ b/server/src/main/resources/overrides/game_objects0.adb.lst
@@ -64,7 +64,7 @@ add_property mini_chaingun equiptime 750
 add_property mini_chaingun holstertime 750
 add_property nano_dispenser equiptime 750
 add_property nano_dispenser holstertime 750
-add_property oicw allowed false
+add_property oicw allowed true
 add_property pellet_gun equiptime 600
 add_property pellet_gun holstertime 600
 add_property peregrine_flight requirement_award0 false
--- a/src/main/scala/net/psforever/actors/session/csr/ChatLogic.scala
+++ b/src/main/scala/net/psforever/actors/session/csr/ChatLogic.scala
@@ -558,9 +558,9 @@ class ChatLogic(val ops: ChatOperations, implicit val context: ActorContext) ext
     var postUsageMessage: Boolean = false
     contents.headOption match {
       case Some(side) if side.matches("i|in|inside") =>
-        session.player.WhichSide = Sidedness.InsideOf
+        ops.customSetSidednessOfTarget(session.player, Sidedness.InsideOf)
       case Some(side) if side.matches("o|out|outside") =>
-        session.player.WhichSide = Sidedness.OutsideOf
+        ops.customSetSidednessOfTarget(session.player, Sidedness.OutsideOf)
       case Some("check") =>
         val whichSide = if (session.player.WhichSide == Sidedness.OutsideOf) {
           "outside"
--- a/src/main/scala/net/psforever/actors/session/normal/ChatLogic.scala
+++ b/src/main/scala/net/psforever/actors/session/normal/ChatLogic.scala
@@ -6,6 +6,7 @@ import net.psforever.actors.session.SessionActor
 import net.psforever.actors.session.spectator.SpectatorMode
 import net.psforever.actors.session.support.{ChatFunctions, ChatOperations, SessionData}
 import net.psforever.objects.Session
+import net.psforever.objects.serverobject.interior.Sidedness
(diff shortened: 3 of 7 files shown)


Dethdeath:
Still damages players/amenities inside.

https://github.com/user-attachments/assets/61493d90-b2d5-41e4-b40b-f568df85e1ae

Fate-JH:
What.  How do the little buddy projectiles travel through the ground.

What was your setup for this test?
[A] Spawn into the lobby, grab the Scorpion, run outside
[B] spawn in(to) the same facility spawn room, wait by the terminals
[A] fire and observe

Dethdeath:
The player firing the scorpion spawned at the AMS you can see in the video, no buildings were entered.
Player in the spawn room spawned at the base.

Fate-JH:
I'm so against working with the `oicw` so much I boiled it down to this:
```
   if (isCrouching && !player.Crouching) {
      //dev stuff goes here
      import net.psforever.actors.session.support.WeaponAndProjectileOperations
      import net.psforever.objects.{DummyExplodingEntity, GlobalDefinitions}
      import net.psforever.objects.serverobject.interior.Sidedness
      val projectile = Projectile(
        GlobalDefinitions.oicw_little_buddy,
        GlobalDefinitions.oicw,
        GlobalDefinitions.oicw.FireModes.last,
        player,
        GlobalDefinitions.oicw.ObjectId,
        pos + Vector3.z(value = 1f),
        Vector3.Zero
      )
      projectile.WhichSide = Sidedness.OutsideOf
      val obj = new DummyExplodingEntity(projectile, projectile.owner.Faction)
      obj.Position = pos + Vector3.z(value = 3)
      WeaponAndProjectileOperations.detonateLittleBuddy(continent, obj, projectile, projectile.owner)()
    }
```
This shortens a bunch of code from `WeaponAndProjectileOperations` that contributes to transforming a Scorpion rocket into its little buddies and then setting up the explosion.  I skip the parts that are unnecessary for the explosion on the server such as the downwards fanning vectors used for the animation.  Essentially, every time I crouch, a little buddy explodes above my character's head and hurts her for 60 HP and 15 armor, but it should only hit her while she is outdoors (`OutsideOf`).

In any event, it does only hurt while outdoors.  Upon respawning in the spawn room, it fails to explode as she is indoors. Upon respawning at an AMS, and precautiously moving away from the AMS since that's not a volatile part of the experiment, the explosion indeed hurts.  One side of the door is an explosion; the other side of the door is a dud.  The `oicw` should work correctly as long as the sidedness of both attacker and target are correct.

Dethdeath:
Another test.

I've included the entire process in the footage below, to show where both characters spawned.
(In 2 parts because of filesize.)

https://github.com/user-attachments/assets/32e2c5fe-cb65-487c-aa6f-2591bc08e7f4

https://github.com/user-attachments/assets/7bb4e823-bf58-4311-9bf8-fa473504b574

I don't know where the sidedness breaks if it even does. Perhaps adding sidedness debug output ingame would help, but every test I've done has the same outcome. It doesn't matter if the guy firing the scorpion spawns at an AMS, base or tower. It keeps doing damage where it shouldn't.

Fate-JH:
Realized an issue with my quick line segment sphere intersection test caused by setting the start and the end point of the line segment to the same position, resulting in a `Vector3(NaN, NaN, NaN)` that acted like the projectile was passing through a door that wasn't there and setting it to `InBetween`.  In an `InBetween` state, the projectile can hit both `InsideOf` and `OutsideOf` targets.  The only affected damage proxies, sub-projectiles, and remote control projectiles, so little buddies hitting entities underground was within the realm of possibilty previously.

I dry tested it by shooting an Scorpion at nothing and ject checking the sidedness of the projectiles and am entirely relying on the above code test as proof that interior targets can not be hit by exterior explosions.

Also included was a normal player variation of the CSR-level bang command `!setside` which is exclusively for `!setside check` functionality.

Dethdeath:
It's working correctly now. 👍

PR #1371, opened 2026-05-17
Wrapped VR Zones

### LINKED ISSUE #1366: Two Air Vehicle Terminals in the VR Driving Area are broken
The Air Vehicle Terminal's at X-2433, Y-1180, Z-18 and X-1309 Y-1770, Z-17 appear to have vehicles spawn from them below the sea level, which instantly disables the vehicle and ejects the pilot, and results in the vehicle getting stuck on the pad and breaking the spawn pads queue system (it gets stuck forever). I didn't catch this issue initially because the player sidedness issue seems to have hid it.

Fate-JH:
The virtual reality zones contain a unique map property in that the terrain get repeated.  The `adb` refers to this as being "`wrapped`".  An instance of the map is smaller than the largest size map, but multiple instances of the map are placed side by side into the largest size map.  This, in essense, means that the VR zones are (the largest) full sized maps but have been dissected into smaller disconnected instances where a player can exists.  There was probably a distribution system that ensured that too many players weren't crowded into one VR zone building but such a system doesn't exist right now.

The kill planes and warning regions for the kill planes have been tiled in such a way that all instances should be bounded off from one another properly.

### DIFF
--- a/src/main/scala/net/psforever/objects/zones/MapInfo.scala
+++ b/src/main/scala/net/psforever/objects/zones/MapInfo.scala
@@ -310,20 +310,20 @@ case object MapInfo extends StringEnum[MapInfo] {
       extends MapInfo(
         value = "map14",
         checksum = 4276645952L,
-        scale = MapScale.Dim1024,
+        scale = MapScale.Dim8192, //Dim1024,
         hotSpotSpan = 0,
         environment = List(SeaLevel(EnvironmentAttribute.Water, 10)) ++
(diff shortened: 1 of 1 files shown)


Dethdeath:
Warnings and killzones work.

Fate-JH:
Made exceptions for the air terminals that miscarried vehicles.
It's been a long time since I made a reference to vehicle creation as child birthing.
Those who remember the now archived Trello post will know.

Dethdeath:
Terminals described in #1366 now work properly.

PR #1372, opened 2026-05-19
Router Demonstration (OCDM)

### LINKED ISSUE #1347: Proximity Map vs Router Pads
Enemy player location data is still being sent when they go through routers. This happens both with the regular proximity radar and also while using audio amp. Seems undesirable to say the least.

Using regular proximity radar:

https://github.com/user-attachments/assets/26c6ea45-3ab3-4507-af9c-8cca9bff590c

Using audio amp:

https://github.com/user-attachments/assets/9f8eec87-b5a9-437e-b328-93da9c9f1960

Fate-JH:
As per the attached ticket submitted by @Dethdeath, I changed the router telepad system to use `ObjectCreateDetailedMessage` and `SetCurrentAvatarMessage` to do a complete reconstruction of the teleporting player in an attempt at bypassing minimap updates and just have the player appear somewhere else.  I've not completely tested whether a second party observer sees anything wonky, either from a faction-aligned or an enemy faction perspective.

From the persepctive of the teleporter, however, this method is incomplete because I have not fully realigned the new(?) avatar and game world completely.  A good demonstration is that the person going through the router does not own any vehicle thbat they origijnally owned coming out the other end.  Other disconnects may also be present.  Once again, this is just to test the adjustment to the teleportation method.

CLOSES #1347

### DIFF
--- a/src/main/scala/net/psforever/actors/session/support/GeneralOperations.scala
+++ b/src/main/scala/net/psforever/actors/session/support/GeneralOperations.scala
@@ -1432,13 +1432,14 @@ class GeneralOperations(
       val pguid = player.GUID
       val sguid = src.GUID
       val dguid = dest.GUID
-      sendResponse(PlayerStateShiftMessage(ShiftState(0, dest.Position, player.Orientation.z)))
+      player.Position = dest.Position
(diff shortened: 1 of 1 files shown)


Fate-JH:
Note: may not close connected issue, do not close connected issue, just testing keywords.

Dethdeath:
Apart from router teleports feeling delayed by what might be almost a second, I'm unfortunately getting the same outcomes as described in the original issue.

PR #1379, opened 2026-06-23
Generic Event Bus

Fate-JH:
> Unhappy fortune! By my brotherhood, the letter was not nice but full of charge of dear import, and the neglecting it may do much danger.

The event system is the realization of our ability to have a proper multi-player experience.  All players communicate with it and all players get correspondence from it.  It has also always been a chokepoint that bottlenecks our proper multiplayer experience.  Just one player send s a message to it and every other player has the potential to get correspondence from it!  Frustrating about it, owing to the straightforwardness of its execution, is that it is designed in such a way that it can be considered wasteful just by operating normally.  Since about 2022, a plan to make the system much more generic existed; but, attempts always fell short and sputtered out, especially in favor of development of a much more presentable mechanism that required development.  It's not technical debt as much as a shallow but self-sufficient design.

If you're going to skip everything else, please read the the first part of "Problem Statement" (below) and the first sentence of "Solution" (also below).

___Technical___
An event system is just a pipeline imposed by the game to create a coherent timeline to the asynchronous packets from a variety of connected clients through their client-facing contexts.  It provides both communication pathways and organizes events into a before-next-after.  The clients are on the network side and the client-facing contexts are on the server side.  From the side of the network, at any given millisecond, a packet arrives from any of the various clients, is given pre-processing by the client-facing context, and then resulting data is queued into a communal pipeline based on its analyzed purpose.  The pipeline takes the submitted data(s) one at a time depending on the indeterminate order of submission and also performs processing on it.  This twice-processed data is hoisted onto another pipeline that is part of the same pipeline as the event system except it is tasked with pushing that data back out onto all of the known client-facing contexts that have communicated with it.  This twice-processed data may be processed further - thrice-processed - but is ultimately sent out by each client-facing context back to their own respective client.  These messages within the pipeline are routed towards predetermined scopes and occasionally blocked from their origin context and all of this determines which of the client-facing contexts actually have to transmit the final packets back to the network.  The basic execution of multiplayer game mechanisms laid bare!  While the pacing of these connected processes is a potential concern, it's the structure that is going to be adjusted herein.

___Problem Statement___
Three issues exist within the design: lack of intuitive design (hitherto, "switch statement abuse"), parallel inheritance hierarchy, and overhead at the garbage collection level.  These issues are not inherently the result of bad choices as the very original singular event system was merely to demonstrate how multiple players could be achievable at all.  Adding congruent other event systems to distribute tasking and lighten loads was proper, but also unexpectedly lazy.  In practice, the first two issues lead to the third issue within proper and robust operation.

__Switch Statement Abuse__
Each pipeline explicitly managed a conversion heirarchy between messages and the replies to those messages, and any message not declared and implemented in the heirarchy was completely unsupported by that pipeline. On paper, that statement should make sense and sound safe.  In practice, it just caused bloat.  There is no formal connection between `MESSAGE_ENVELOPE` and `REPLY_ENVELOPE` or every `MESSAGE` and `REPLY` except in a hard-coded dictionary sense.  All pipelines were required to look like so:
```
INTO_PIPELINE1 -[
   MESSAGE_ENVELOPE1(TO_A1, MESSAGE_A1(FILTER_A, DATA_A1...)) becomes REPLY_ENVELOPE1(OUT_A1, FILTER_A, REPLY_A1(DATA_A1 ...)))
   MESSAGE_ENVELOPE1(TO_B1, MESSAGE_B1(FILTER_B, DATA_B1...)) becomes REPLY_ENVELOPE1(OUT_B1, FILTER_B, REPLY_B1(DATA_B1 ...)))
   ...
   MESSAGE_ENVELOPE1(TO_*1, MESSAGE_*1(FILTER_*, DATA_*1...)) becomes REPLY_ENVELOPE1(OUT_*1, FILTER_*, REPLY_*1(DATA_*1 ...)))
                ]- OUT_OF_PIPELINE1
```
... as written out exhaustively in the code.  If the pipeline is to process a `MESSAGE_ENVELOPE1(TO_*1, MESSAGE_Z1(FILTER_A, DATA_Z1...))` then it has to be modified to catch it, then intentionally spit out a corresponding response.  An actual benefit to this is that `switch` statements, being fixed compiled pathway code, can be quick to resolve.  However, they lead to:

__Parallel Inheritance Hierarchy__
Every pipeline has an exclusive `MESSAGE_ENVELOPE` to carry one of any exclusive types of `MESSAGE` and then converted those into equally exclusive `RESPONSE_ENVELOPE` and `REPLY`.  The input side necessitates an existence on the other side.  To be brief about the content explosion this causes, for every `PIPELINE1` and everything given a "1" form because of it, there is a `PIPELINE2` with its own ensemble of "2" form entities that are required to be defined.  Even if there exists a `MESSAGE1` that is visually and functionally the exact same as `MESSAGE2`, even if they could be allowed to inherit the shared components from a common ancestor, they must exist as separate entities and can never venture into one another's workflows.  `MESSAGE1` is for `PIPELINE1` and `MESSAGE2` is for `PIPELINE2` and that was final.  What makes this relationship worse is that this is not actually a common instance of parallel heirarchies born of poorly-planned subclassing and misappropriated or lopsided distribution of tasking.  The pipelines embody this shared feature design but are incapable of sharing the realization of this feature design purely for the reason that they would resemble each other.  The exclusivisity was at one point even forgiven with such specious argument as a means of ensuring that the wrong messages are not sent to the wrong places, but such a handwave still does not justify the bloat, and shouldn't have been a problem to ward against anyway.

__Garbage Collection Overhead__
In prior sections, it has already been demonstrated that a lot of exclusive objects are being created and tossed aside just to support this operation.  The pipeline is a busy thing by intention, by requirement.  Every client has access to it and utilizes it at least once every 250 milliseconds to inform every other client what it is doing and what should be happening.  A popular heuristic we throw around is "for every twenty players, 400 messages are processed"; in reality, it's still the same twenty messages but it's that they need to be processed twenty times by twenty clients within a short time frame.  That's the 400 - instances where a client-facing context processes an event system message per second.  The fortunate news is that only twenty messages are ever generated; the unfortunate news, however, is that everything is uncoordinated and happens through a combination of serendipity.  While such irregular object creation is cheap because object creation is cheap over-all, it tragically staggers garbage collection and the actual lookup for message sending endpoint(s) has to work overtime to achieve the same results.

The `FILTER` is the only object that is preserved as it is between the "message" side to the "reply" side, but there are certain occasions where it isn't used anyway (which impacts `MESSAGE`).  `TO` and `OUT` are typically `String` data and that means the JVM should be able to intern them.  To be quick about the explanation for that, Java and Scala `String` objects are only created once; and, as long as they are not eliminated due to lack of persisting reference, that `String` object will remain to be referenced in all subsequent attempts to create the exact same text object.  Across all pipelines and all messages, regularity of the verbage used in these objects and their consistent referencing by user-level object structures should restrain further `String` creation.  In other words, every `TO` and `OUT` only need be created once for all object creation instances as long as some previous version of `TO` and `OUT` remain in use before garbage collection.  It will still take time to resolve but the pooling should save memory.

There's no saving grace for `MESSAGE_ENVELOPE`, `REPLY_ENVELOPE`, `MESSAGE`, and `RELY`.  The `MESSAGE_ENVELOPE` only exists long enough to transport the `MESSAGE` into the event system with appropriate routing information and both are soon abandoned for `REPLY_ENVELOPE` and `REPLY` along with the unnecessary information swaps between the input and output entities.  They should be garbage collectable by this point at least.  `REPLY_ENVELOPE` and `REPLY` have a complicated time because they will necessarily exist until every client-facing context has had a chance to properly resolve their message and that issue compounds with every new subscriber to the given channel at best, every new client at worst, and any potential lag that might occur will causes these messages to back up and exist for far longer.  Occasionally, `REPLY` will persist for other reasons while `REPLY_ENVELOPE` will be retired early.  The garbage collection time for these entities is indeterminate because only the JVM will know when all parties will have started or will have completed referencing the objects.

___Solution___
Fewer entities are needed and this is an easy solution to achieve.

As we have established that `MESSAGE` and `REPLY` and `MESSAGE_ENVELOPE` and `REPLY_ENVELOPE` are mostly synonymous, then we can eliminate one from either pair easily.  By removing `FILTER` from `MESSAGE(FILTER, DATA ...)` and having it be transported by `MESSAGE_ENVELOPE`, then `MESSAGE(DATA...)` can look exactly like `REPLY(DATA...)` which gives better justification for dropping one, as long as we change the envelope entities.  Sure enough, now `MESSAGE_ENVELOPE(TO, FILTER, MESSAGE)` looks much more like `REPLY_ENVELOPE(OUT, FILTER, REPLY)` and, if the payload's requirements are changed, we can fabricate a `REPLY_ENVELOPE(OUT, FILTER, MESSAGE)` reasoning.  `OUT` is the salted routing information that the event system need to know how to collect the subscribers for channel `TO` and, to be honest, there's no reason this information needs to escape the event system, least of all be placed on a pedestal in the reply-side entities.  By doing this, there is no reason to have a `REPLY_ENVELOPE(OUT, FILTER, REPLY)` anymore, just the initial `MESSAGE_ENVELOPE(TO, FILTER, MESSAGE)` used as a reply-side entity!

Some occasions where having distinct in and out functionality is preferrable do exist.  The code should have some way to verify that the event system was used at all and also know which event system was used, to allow different pipelines to have different outcomes for the same message.  Additionally, not all `MESSAGE` objects become a `REPLY` object with a similar parameter list.  This gives us a reason to intentionally perform conversion for `MESSAGE` to `REPLY`, even if `MESSAGE` will stay as itself the majority of the time, and we should enforce it being a many:one situation.  One type of `MESSAGE` should always become a certain type of `REPLY`, and many types of `MESSAGE` may always become that one kind of `REPLY`, but one `MESSAGE` always becomes only that `REPLY`.  Since we're now far closer to our metaphorical postal analogues that the `MESSAGE` and the `MESS_ENVELOPE` don't need to meaningfully change upon delivery, let's mark the `MESSAGE_ENVELOPE` with an entity that explains how it will be routed and that the contents should be interpreted as a "reply" and not a "message".  We apply a "stamp" that transparently recontextualizes the `MESSAGE_ENVELOPE` into a `REPLY_ENVELOPE`, even though the entity never actually changes.  Entities maintain both their own forms and also the forms of their products.  As long as the individual elements all manage their own conversions, all pipelines can be converted into the form:
```
INTO_PIPELINE -[
   ENVELOPE(TO, FILTER, MESSAGE(DATA...)) becomes ENVELOPE(TO, FILTER, MESSAGE(DATA ...))) { STAMP; REPLY(DATA...) }
                ]- OUT_OF_PIPELINE
```
... where the issues of parallel inheritance and switch statement abuse are no longer.  Any `ENVELOPE`s can be used as an `ENVELOPE`, regardless of the side of the event system it is encountered.  `MESSAGE` and `REPLY` will often be the exact same entity, or it may be a replacement entity, but it's now tagged onto the original entity, and transforms itself.  Even the pipeline does not have to be redefined and specialized for every pipeline to be implemented.

__Support Actors__
On the previous event systems, various child actors that performed some routine entity management existed.  Common actions like base capture, hack de-escalation, and door closing were performed in a way that that required advanced task execution and scheduling at the same level at the event system.  In reality, I could never decide when some behavior of an entity required being tagged on top of the event system and when it could be embodied in the entity control agency itself.  Moving deployable deployment from within the deployable control agency into a support actor, and then back into the control agency, is proof of that indecisiveness.  That bit of impostor syndrome aside, these child actors were hard-coded into the identity of specific event systems and into the laborious handling cases of some messages.  Reintegrating behaviors from these support actors into the separate entity control agencies in the game architecture was never a planned change, so an event system implementation and an input-side envelope implementation were required to continue to support these behaviors without sacrificing the increased generic-ness of the main event system.

The actual behavior of the support actors are unchanged.

___Features___
Actual substantive entities will still exist but changes will rely heavily on `trait` scoping.

__`GenericResponseEnvelope`, `EventResponse`__
Output-side perspective given to the event system entities, that envelope and payload.  Some event system still main unique `*ServiceResponse` envelopes, but those are main used in testing.

__`GenericMessageEnvelope`, `EventMessage`__:
Input-side perspective given to the event system entities, that envelope and payload, including the method of converting the entity to an output-side perspective.

__`SelfRespondingMessage`__
A common combined payload perspective that embodies both the input-side perspective and the output-side perspective.  Most messages are this sort of perspective and it flags that the payload does not change when switching between input-side and output-side.

__`MessageEnvelope`__
The pipeline-agnostic implemention of a basic `GenericMessageEnvelope` entity that satisfies both input-side parameter requirements as well as output-side parameter requirements.  Serves as primary delivery mechanismand core example of joint `GenericResponseEnvelope` behavior when interpreted (cast) from a different perspective.  It replaces all of the previous `*ServiceMessage` and `*ServiceResponse` envelope entities in the previous system; however, wrappers for all of the `*ServiceMessage` entities are retained for efficiency, and modified versions of the `*ServiceResponse` envelope entities  continue to be used, mainly in testing.

__`EventSystemStamp`__
An object that clarifies the difference between event systems and is used on envelopes being transported through event system pipelines to indicate a change from input-side to output-side recognition.

__`GenericEventService`__
The rudimentary but essential event system pipeline.  Contains a bus that also caches subscribers (mostly unchanged from original) and the logic to transform input-side "messages" into output-side "replies".  Individual event systems lack the excessively verbose distinguishing features of their originals - `AvatarService`, `LocalService`, etc. - so a clarifying stamp is provided.  Any entity that is processed as an input-side message by the event system must include the stamp with the message so that the pathway used by the output-side response can be polled.

__`GenericEventServiceWithSupport`, `GenericSupportEnvelope`__
A subclass of more a generic event system that considers the existence of support actors as front-loaded units of additional entity management associated with labels.  That front-loading is performed by `EventServiceSupport` entities in the constructor parameters.  Invocation of these support actors is performed by envelopes that search by the known label and then pass a message to that actor.  One flavor of this envelope communicates with the support actor and then gets handled like a normal input-side to output-side envelope; the other only communicates with the support actor.

__`GenericEventServiceWithCacheAndSupport`, `CacheEnvelope`__
A subclass of the prior subclass of a generic event system that does not immediately perform input-side to output-side translation of envelopes and messages and instead lets such messages pool for a few milliseconds before processing them at once.  Interpretation of caching is a core feature expansion in the PSForever fork https://github.com/Pinapse/giant in an attempt to minimize JVM garbage collector strain.  Only includes a variation "with support (actors)" because it never gets executed in any context where there isn't also one or more support actors on that event system.  `CachedEnvelope` is the unique envelope variety that causes whatever is being communicated by the envelope to be sequestered, pooled, and transmitted later.  Pooling duration is 50 milliseconds from the time of the first message after the last cache flush.

At the moment, caching is hard disabled by replacing the cached event system with a normal event system.  The timing needs to be retuned from its source fork.

___Caveats___
1.  There's a massive difference between services affected by these changes and those that are not.  The ones that have been updated are postal-type and merely propagate newsletter-type messages without spending much time having to process them - `AvatarService`, `LocalService`, etc..  The influence of support actors does not count, only message turnaround time.  Services that incur significant processing towards an operation like the management of outfit and outfit services are unaffected, unless one of its messages is exceptionally regular.
2.  Messages that are cached correspond mainly to messages whose corresponding packets are used to flag player "online" status that is shared between other players, called "upstream packets" internally to the project.  They report a player's basic activities in the world depending on their circumstances and are expected to be recieved by the server frequently.  `PlayerStateMessageUpstream`, `VehicleService`, and `ChildObjectStateMessage` are common examples.  `KeepAliveMessage` can be an example of an "upstream" message but is never communicated with other players.  `ProjectileState` is cached because it maintains a 1:1 frequency with a simultaneous dominant "upstream" message.  As mentioned above, caching is disabled.
3.  Although some messages still seem like they are organized into collections based on event system, that association is mere inner class and a suggestion.  `AvatarAction` messages are treated no differently from `LocalAction` messages or the newly isolated class of "used in multiple event system" messages.  The intention was to remove strict requirement between event system and envelope, and envelope and message and reply.  Despite that, it remains useful to separate the client-facing context responses to these messages based on which event system from which they were received and certain messages do expect to only to handled by certain event systems due to unchanging synchronization requirements.  Even if it were beneficial, fully redesigning the spread of functionality for sessions with unrestricted freedom in mind is not within the wheelhouse of this branch PR and would not be significantly helpful.
4.  While `GenericEventServiceWithCacheAndSupport` manages both cached messages and messages to support actors, there is no practical case in the current server operation where both happen to the same message.  In the event that it does need to occur, the current policy is to defer all processing of the envelope until after the cache has been flushed and that includes support actor interaction.  A test demonstrates this specifically.  There is no method to prioritize one over the other.  As mentioned above, caching is disabled.

___Addenda___
1.  The messages utilized by doors has changed.  Previously, the structured relied on half of the messaging coming from `DoorControl` agency and the other half from `DoorCloserActor` support.  Now, all door opening and closing is managed by an envelope or message or response where the event system is fore-most and feature-local messaging is reduced.
2.  An underutilized feature of the new event system is the ability to accept multiple packets at the same time as long as they share the same channels and filters, but only when using the `SendResponse` message.
3.  An overutilized feature of the new event system is the ability to accept multiple message envelopes bundled together at the same time.
4.  Removed an overloaded constructor from `HackMessage` (packet) because it was underused and was conflicting with a different class also, at the time, called "HackMessage".  It's `HackEntityEnvelope` now but the overload has still be axed.
5.  Fixed `DamageModelTests` independent of needing to repair anything due to the change in event systems.  It was just bugging me that their passing was always being held back from basic `NoGUIDException` problems.
6.  `PropertyOverrideManager` extracts data from its file contents in a manner than does not use `StringBuilder` or other iterative `String` assembly.
7.  Emote recipient determination is now performed in a different manner and at a different place.  I attempted to utilize the session's own knowledge of their blockmap sector.
8.  Corrected an obtuse process with the `GridInventory` knapsacking algorithm intermediary node class.  Child nodes were initialized on the parent node with zero size and were resized a function call later using what looked like an `apply` overloaded constructor.  Now the nodes are just created with the apprioriate sizes in mind and passed to the method that assigns them to their parent node.  It worked just fine before but was design nonsense.  Even the IDE would look at the code and be unable to make heads or tails of it.  It'd just mark it as "code that does nothing".
9.  I attempted to update `codecov.yml` for the `services` folder and included a bunch of other things.  Not really ever gonna be a problem since that just affects meaningless percentages.  If the test fails, it'll still fail, and that means something is failing.
A.  Everything broke systematically at different times during the course of developing this new event system and was rrpaired promptly, only to break down and be repaired again and again each change enacted.  The final product runs.  That's it.
B.  Rather than a song riff as usual, we have straight classic literature.  Friar Laurence, William Shakespeare's __Romeo and Juliet__, Act 5 Scene 2.


Fate-JH:
If anything is broken, I will correct it and you will be compensated*.
&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
<sub><sub>* - No compensation will be forthcoming.</sub></sub>

PR #1380, opened 2026-07-16
GUID Number Pool Shuffling

### LINKED ISSUE #1349: AMS cloaking bubble hiding too much
Currently when inside the bubble, you don't see weapons on the backs of friendlies or in their hands. Their names are visible.
Players that enter a hostile AMS cloaking bubble cannot see enemy names or weapons.
This used to be different.

Some reference material and examples of how it should be:

Image

https://youtu.be/9s6jKhV6ah8?t=1234

Image

https://youtu.be/JeTYFBPg6kk?t=6484

Image

https://youtu.be/6LaSYpxQ9Gg?t=11

Image

https://youtu.be/QGpEOABhMKw?t=211

Image

https://youtu.be/XmX-9iCr8tk?t=3149

Fate-JH:
As per @ScrawnyRonnie's comment in #1349, shifting the global unique identifier number pools around thus that players are higher than vehicles corrects visibility issues regarding holstered and equipped weaponry when interacting with the AMS cloak bubble.  The GUID's for lockers, i.e., the player's internal carrying capacity reference point for their locker space, has been shifted to the end of the list where appropriate.

This solution makes me concerned the client imposes curious GUID requirements for randomly generated entities much like it does for static environment entities.  In the best case scenario, it just processed cloaked entities in an order of GUID for some specific check; and, this is really all we need to ever do to solve the only issue like this we will encounter.

### DIFF
--- a/src/main/resources/guid-pools/default.json
+++ b/src/main/resources/guid-pools/default.json
@@ -6,39 +6,39 @@
     "Selector": "specific"
   },
   {
-    "Name": "players",
+    "Name": "vehicles",
     "Start": 3001,
-    "Max": 4500,
-    "Selector": "random"
-  },
-  {
-    "Name": "lockers",
-    "Start": 4501,
     "Max": 5000,
     "Selector": "random"
   },
   {
-    "Name": "tools",
+    "Name": "lockers",
     "Start": 5001,
-    "Max": 9500,
+    "Max": 5500,
+    "Selector": "random"
+  },
+  {
+    "Name": "tools",
+    "Start": 5501,
+    "Max": 10000,
     "Selector": "random"
   },
   {
     "Name": "ammo",
-    "Start": 9501,
-    "Max": 23000,
+    "Start": 10001,
+    "Max": 23500,
     "Selector": "random"
   },
   {
     "Name": "kits",
-    "Start": 23001,
-    "Max": 36500,
+    "Start": 23501,
+    "Max": 37000,
     "Selector": "random"
   },
   {
     "Name": "items",
-    "Start": 36501,
-    "Max": 39500,
+    "Start": 37001,
+    "Max": 39700,
     "Selector": "random"
   },
   {
@@ -60,7 +60,7 @@
     "Selector": "specific"
   },
   {
-    "Name": "vehicles",
+    "Name": "players",
     "Start": 45001,
     "Max": 47000,
     "Selector": "random"
--- a/src/main/resources/guid-pools/z3.json
+++ b/src/main/resources/guid-pools/z3.json
@@ -6,14 +6,8 @@
     "Selector": "specific"
   },
   {
-    "Name": "players",
+    "Name": "vehicles",
     "Start": 4001,
-    "Max": 5500,
-    "Selector": "random"
-  },
-  {
-    "Name": "lockers",
-    "Start": 5501,
     "Max": 6000,
     "Selector": "random"
   },
@@ -45,6 +39,12 @@
     "Name": "rc-projectiles",
     "Start": 64001,
     "Max": 64300,
+    "Selector": "random"
+  },
+  {
+    "Name": "lockers",
+    "Start": 64301,
+    "Max": 64800,
     "Selector": "random"
   }
 ]
--- a/src/main/resources/guid-pools/z4.json
+++ b/src/main/resources/guid-pools/z4.json
@@ -6,14 +6,8 @@
     "Selector": "specific"
   },
   {
-    "Name": "players",
+    "Name": "vehicles",
     "Start": 3101,
-    "Max": 4600,
-    "Selector": "random"
-  },
-  {
-    "Name": "lockers",
-    "Start": 4601,
     "Max": 5100,
     "Selector": "random"
   },
@@ -45,6 +39,12 @@
     "Name": "rc-projectiles",
     "Start": 64001,
     "Max": 64300,
+    "Selector": "random"
+  },
+  {
+    "Name": "lockers",
+    "Start": 64301,
+    "Max": 64800,
     "Selector": "random"
   }
 ]


Dethdeath:
Issue is fixed.

GeekOfWires:
Nice catch on the root behavior — the vehicle case is clearly right. Sharing a deeper look at the concern itself and at how this plays out across the rest of the pool code, plus a few suggestions.

### On the concern: does the client impose a "curious GUID requirement" on random entities like it does on static ones?

It's worth separating two different kinds of requirement, because they aren't the same:

- **Static/environment entities have a hard *identity* requirement.** Their GUIDs have to equal the specific numbers the continent data already assigns them — which is exactly why that pool is `Selector: "specific"` (and why `Zones.init` seeds `projectiles`/`locker-contents` at fixed numbers via `map.addLocalObject`). The client knows those objects by fixed id from the map, so the server must return the *same value*.
- **Randomly-generated entities don't appear to have that.** The client treats dynamic-object GUIDs as **opaque handles** — it registers them in a flat GUID→object table plus an **ascending-ordered index** keyed on the raw 16-bit number. It doesn't care what the value is, only that it's unique.

What the client *is* sensitive to is **order**: it keeps live objects in that ascending index and resolves a child's visibility at attach time (held/holstered weapons are child objects). A spatial cloak like the AMS bubble is plausibly resolved by walking objects in that order, emitter-before-receiver. So your best-case reading is the well-founded one — this isn't the client demanding specific numbers for dynamic objects, it's the client processing in GUID order, and putting vehicles below the things they should cloak is the correct lever. The only place I'd soften *"all we need to ever do"* is that the same ordering logic could apply to more than just vehicles (below).

### Looking at the whole pool setup (not just the diff)

Post-PR ascending layout of `default.json`:

| Pool | Range | Selector | Note |
|---|---|---|---|
| environment | 0–3000 | specific | static map entities |
| **vehicles** | **3001–5000** | random | AMS now sorts right after static geometry ✓ |
| lockers | 5001–5500 | random | |
| tools | 5501–10000 | random | weapons |
| ammo | 10001–23500 | random | |
| kits | 23501–37000 | random | |
| items | 37001–39700 | random | |
| rc-projectiles | 39701–40000 | random | |
| projectiles | 40100–40149 | specific | |
| locker-contents | 40150–40450 | specific | |
| **players** | **45001–47000** | random | now high ✓ |
| terminals | 47001–48000 | random | |
| deployables | 48001–64000 | random | highest |

Five things fall out of the surrounding code that the diff alone doesn't show:

1. **The only ordering the system can guarantee is band-level.** Every dynamic pool uses `RandomSelector`, which hands out a pseudo-random number within the band (it even re-introduces returned numbers, so repeats happen well before the band is exhausted). Because `NumberPoolHub.AddPool` enforces disjoint pools, cross-type order (`vehicles < players`) always holds regardless of the randomness — which is *why* the fix is robust — but nothing finer than band order is ever reliable. So this is safe, just don't ever lean on within-band ordering.

2. **Coverage is actually complete — good news.** Of the 27 zone pool files, only `default.json`, `z3.json`, and `z4.json` pin their own `players`/`vehicles` band; the other 24 inherit both from `default` via `customizePools`. So fixing `default` propagates the reorder to all of them automatically, and `z3`/`z4` are the exact two zones that override, both of which this PR handles. No continent is left on the old ordering. (The reason z3/z4 pin their own low pools at all is that their `environment` ranges are bigger — 0–3732 and 0–3076 — so they had to push the dynamic bands up.)

3. **The move introduces a new low-end coupling worth guarding.** Pre-PR, `vehicles` lived at 45001, so it could never collide with any `environment` pool. Post-PR its floor is **3001**, sitting just above the largest *inheriting* zone's environment (`z9` = 2921 → ~79 ids of slack; `z7` = 2816, `z10` = 2657). No collision today, but any inheriting zone whose static set ever grows past 3000 would now overlap the inherited `vehicles` band and hard-fail at zone init via `AddPool`. (z4 is already tight on its *own* numbers: environment 0–3076, custom vehicles start 3101 — 25 ids.) This didn't exist before the PR.

4. **The fix is verified for vehicles; the deployable class is an open question.** `deployables` (48001–64000) sits *above* players and includes the cloak-capable deployables (cloaking spitfire, sensor/cloak bubbles), so if the GUID-order theory is the mechanism, it'd be worth checking whether deployable-based cloak shares this behavior — the AMS fix works because AMS is a *vehicle* that went to the very bottom, and it's not obvious that reasoning carries to deployables sitting at the top. Not asserting a bug, just flagging it as the next thing to poke at.

5. **z3/z4 lockers now sit in the client's local-GUID window.** The new `lockers` at 64301–64800 is legal server-side (`MaxNumberSource(max = 65536)`) but overlaps the range the client seeds its own local GUID factory into (~64535–65535). Probably fine — the client's lower local windows already overlap server ranges without incident, implying coordination — but it's the one spot the reshuffle pushes server ids into client-owned territory, so worth a conscious check.

### Recommendations

- **Turn the invariants into a load-time assertion so they can't silently regress.** The ordering + the new environment/vehicles coupling currently live implicitly in hand-edited `Start`/`Max`. `addPools` in `Zones.scala` is the natural chokepoint:
  ```scala
  private val ClientReservedTop = 64535 to 65535   // client seeds a local GUID factory here

  private def validatePools(guids: Seq[GuidNumberPool], zoneId: String): Unit = {
    def band(name: String) = guids.find(_.name == name).map(p => p.start to p.max)

    // (1) static 'environment' must not run into the lowest dynamic band (now 'vehicles')
    val envMax = guids.find(_.name == "environment").map(_.max).getOrElse(-1)
    val lowDyn = guids.filter(_.name != "environment").map(_.start).minOption.getOrElse(Int.MaxValue)
    if (envMax >= lowDyn)
      throw new IllegalStateException(s"[$zoneId] environment (…$envMax) collides with lowest dynamic pool (start $lowDyn)")

    // (2) cloak-order: emitters must sort below their receivers
    val emitterMax  = Seq("vehicles").flatMap(band).map(_.end)
    val receiverMin = Seq("players", "tools").flatMap(band).map(_.start)
    for (e <- emitterMax; r <- receiverMin if e >= r)
      throw new IllegalStateException(s"[$zoneId] cloak-order: emitter max $e >= receiver start $r")

    // (3) keep server pools out of the client's local-GUID window
    guids.filter(p => p.max >= ClientReservedTop.start && p.start <= ClientReservedTop.end)
         .foreach(p => throw new IllegalStateException(
           s"[$zoneId] pool '${p.name}' (${p.start}-${p.max}) overlaps client window $ClientReservedTop"))
  }
  ```
  Check (1) is stricter and clearer than the raw `AddPool` overlap error and directly guards the new coupling from point 3; check (3) catches point 5.

- **If you want to close the loop on the deployable question, test it empirically.** Deploy a cloaking spitfire / cloak bubble next to infantry and see whether any holstered/equipped-weapon visibility oddity shows up. If it does, deployable emitters behave like AMS and would want a low band too (and could then join `emitterMax` in check (2)); if not, the vehicle fix stands on its own.

- **Add a pre-merge overlap check to `customizePools`.** The override is name-based (`custom ++ remainder`) and works here only because the `players`→`vehicles` rename lines up with the default remainder. A future zone override whose *range* overlaps a non-renamed default pool would only surface later as a generic `AddPool` collision. Failing fast with the zone id is friendlier:
  ```scala
  private def customizePools(base: Seq[GuidNumberPool], custom: Seq[GuidNumberPool], zoneId: String = "?"): Seq[GuidNumberPool] = {
    val merged = custom ++ base.filterNot(b => custom.exists(_.name == b.name))
    for (Seq(a, b) <- merged.combinations(2).toSeq if a.start <= b.max && b.start <= a.max)
      throw new IllegalStateException(s"[$zoneId] pools '${a.name}' and '${b.name}' overlap after merge")
    merged
  }
  ```

- **Longer term, derive the ranges from an ordered intent list rather than hand-numbering ~27 files.** Something like `Seq(("environment", 3001), ("vehicles", 2000), ("lockers", 500), …)` packed contiguously (skipping the reserved top window) makes list-position *the* ordering contract and removes the whole "someone renumbered and cloaking/init broke again" class of bug. Happy to sketch it if there's appetite. And if any dynamic entity ever truly needs a *specific* client-visible value like the statics do, that's what the `specific`/`StrictInOrder` selectors are for — but that's a different requirement than the ordering one this PR is really about.

Net: the vehicle fix is right and its coverage across zones checks out. The above is about making the ordering + the new low-end coupling explicit, and about whether deployable emitters are worth a second look.

---
*Transparency note: this review was carried out by **Claude Opus 4.8** (Anthropic). The analysis compared this PR against the PSForever `master` codebase (the `guid-pools` configs, the `Zones` loader, `NumberPoolHub`, and the selector classes) and against reviewed original client GUID-handling behavior. It's been reviewed before posting — treat the empirical claims (esp. the deployable-cloak question) as "worth verifying," not settled.*

PR #1381, opened 2026-07-17
Leza Terms

### LINKED ISSUE #1353: Leza Equipment Terminals
Some of the equipment terminals at Leza on Cyssor cannot be repaired with a nano dispenser. The terminals in question are the lobby equipment terminal and the right most equipment terminal in the spawn room (closest to the lockers). Enemies were also not able to hack or interact with these terminals after the base repaired them.

ScrawnyRonnie:
Fix #1353


Dethdeath:
Terms are working as they should now.

PR #1387, opened 2026-07-27
You can grab your balls again

ScrawnyRonnie:
- Should fix llu issues such as not being able to pick it up
- Capturing a tower should no longer trigger a zone benefit lock when a single facility is still owned by the opposing empire

Neither has linked issues; llu was mentioned in chat and the tower capture thing has a few #bug-reports


Dethdeath:
Both issues seem fixed.

PR #1390, opened 2026-08-11
Explicit Shuttle Empire

ScrawnyRonnie:
Whoever's been piloting the HART shuttles shouldn't get lost anymore


Dethdeath:
Shuttle colors all correct on the test server.

PR #1392, opened 2026-08-18
I can leech BEP again

### LINKED ISSUE #1389: Squad Messages
#1379 introduced some issues with squads. From what I've seen, listing a squad and promoting a squad member to leader aren't working. Listing and making changes to the squad details should immediately inform the clients.

ScrawnyRonnie:
I think I did this right & should fix #1389

### DIFF
--- a/src/main/scala/net/psforever/services/teamwork/SquadService.scala
+++ b/src/main/scala/net/psforever/services/teamwork/SquadService.scala
@@ -374,8 +374,9 @@ class SquadService extends Actor {
    */
   def JoinByFaction(faction: String, sender: ActorRef): Unit = {
     val path = s"/$faction/Squad"
-    log.trace(s"$sender has joined $path")
-    subs.SquadEvents.subscribe(sender, path)
+    val routed = SquadStamp.routing(path)
+    log.trace(s"$sender has joined $routed")
+    subs.SquadEvents.subscribe(sender, routed)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Fate-JH:
Does this really work?

I thought I had excluded `SquadService` from core changes to event service structures , e.g., the `EventStamp` mechanism, and generalization because it didn't match the exact mold of of the other services I had been modifying.  The corresponding stamp does exist but that's mainly because I only got that far before realizing the service wasn't (easily) compatible for the approach I had envisioned.  Other event systems just proliferate.  The squad system keeps track of state and makes decisions.  That's why it's still just an plain `Actor` with a `GenericEventBus` hiding underneath it rather than a `GenericEventService` which is where the stamp comes into play.  Additionally, squad response messages should be caught before the switch that changes handling based on what kind of stamp is attached to the envelope comes into play (see [`SessionActor.scala`](https://github.com/ScrawnyRonnie/PSF-LoginServer/blob/d122ec7cb83fcfe110f636440da8a84c5676ea4b/src/main/scala/net/psforever/actors/session/SessionActor.scala#L208)).

The stamp adjusts channel name.  Did I screw up somewhere?

ScrawnyRonnie:
Well, every squad list/update/join/leave/promote event is working again. I just did the usual println everything to see where it broke. It looked like the system was using `/Faction/Squad/out` but the players were joining `/Faction/Squad` and never received the published messages because they joined a different channel name.

Dethdeath:
Can confirm squad member promotions, joins, leaves, listings and updates all work again.

PR #1394, opened 2026-08-31
Turret Upgrade Block

### LINKED ISSUE #1370: Turret upgrade - stuck in animation
When a player is upgrading a wall turret and another player enters the turret during the process, the upgrader gets stuck in an animation.

https://github.com/user-attachments/assets/b943738a-d8d0-4a27-a10f-5cf4b7a5b391

Fate-JH:
I admit to not testing the initial fault conditions but this change was tested for the following, as per the linked issue:

1. Turret will not upgrade while occupied
2. Turret will not allow another player to mount it while it is in the process of being upgraded
3. Turret WILL allow another player to mount it if upgrading has been started but an upgrade path has yet to be selected; once the upgrade path has been decided, the current upgrade process will terminate (because the turret is occupied)


Dethdeath:
Working as described, issue fixed.

PR #1395, opened 2026-08-31
Updating Global Pop

ScrawnyRonnie:
Should send `ZonePopulationUpdateMessage` as players enter and leave zones to keep the global population accurate. It currently only sends once on login. This updates the 'Global' windows on the map with % of players in each zone


Dethdeath:
Working great. This is more accurate than /who now as it even counts players on the spawn select screen.

PR #1397, opened 2026-09-14
Logical Building Logic Local

Fate-JH:
Moved the selection of `BuildingLogic` that was given to bases from the constructor of `BuildingActor` to the global extractor and instantiator of all zones and zone objects in `Zones.scala`.  Appropriate function signatures were adjusted to enforce the requirement of external logic assignment.  Tests were modified so overloaded constructors for `Building` entities could be culled.


Fate-JH:
Zones now take a reference to their canned information object instead of the exploded fields from that said object.

Zone entities report how they are caverns and virtual reality is a different way.  `PlanetSideGameObject.IsInVRZone` is replaced by `Zone.isVR` and `ZoneMap.cavern` is replaced by `Zone.isACavern`.

PR #1398, opened 2026-09-17
Vanu Packets

Fate-JH:
Two packets, both tested, used in one way or another in service of vanu modules.

__`ModuleLimitsMessage`__
It is not known what this packet does.  Its effects can not be observed in any known UI element and its values are always defaulted to the following values: `255`, `15`, and `3`.  255 fills out the full length of the datatype, 15 isn't related to any specific to modules (at least according to the wiki), and 3 is at best the number of modules that can be moving around freely in any zone; but, none of this can be confirmed currently.  The weirdest aspect of this packet is that it fires off once per a zone's initialization packet cluster at login time but it does not define which zone to which it belongs in the same way other packets do using the zone number.  If it does perform some action per zone, the presumed zone is defined by the surrounding packets instead.

Because this packet never strayed from default values during the course of our packet capture period, the real ingame conditions this packet relates to are unknown.

__`VanuModuleUpdateMessage`__
This packet serves the same purpose for Vanu module units as `CaptureFlagUpdateMessage` performs for lattice logic units.  In our packet captures, it only defines a zone number and an empty information list.  The information list includes the type of the module being displayed on the map, the location of the module on the map, and whether the module flashes red or green.  What the flashing color is supposed to represent is unknown, though carrying availability, carried by friend or enemy, or module charge state, are all possibilities.

The data utilized in the test for this packet is entirely fabricated.  Because this packet never strayed from default values during the course of our packet capture period, real packet information related to modules to be displayedcan not be sampled.


Fate-JH:
> A module icon flashing **GREEN** means a friendly has picked it up. A **RED** flashing icon means an enemy has it, and a **BLACK** flashing icon means the module is on the ground.

-- https://www.psforever.net/modules/

Flashing color information, emphasis is mine.  Either my monitor colors are messed up or I've not been able to determine how the black flashing works/looks yet.  Flashing color is supported by a two-bit field, with all but 3 being green from what I can gather, and 0 being our defaulted red.  I'll review my packet color options.

Fate-JH:
Despite manipulating a number of fields in combination, I could not replicate the black flashing animation that the wiki describes.  Itis perhaps still possible but the method eludes me.

What I did discover are unique icons beyond just the modules.  The field describing the icon is 4 bits, filling out 15 integer values from hex `0` to hex `F`. 2-7 are the specific module types; 0, 1, and 7-9 appear the same as non-powered; and, 13 (`D`) - 15 (`F`) do not render as anything.  Icons 10 (`A`) through 12 (`C`), however, are the standard empire logos from TR to VS in module icon size.  They flash red and green like the normal module icons though it can be hard to see the red flash on the TR logo for obvious reasons.  I am not certain what their purpose could be at this time.

